PR #6, opened 2015-04-25
feat(installer): adds a simple github installer

zewa666:
might be used to install packages directly from github repos. Gets the latest available tag and installs it. Runs NPM and JSPM installs afterwards

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@ var program = require('commander');
 var bundler = require('./lib/bundler');
 var pjson = require('./package.json');
 var chalk = require('chalk');
+var installer = require('./lib/installer');
 
 function Aurelia(env) {
   this.env = env;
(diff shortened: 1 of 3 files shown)


zewa666:
@ahmedshuhel I'm very sorry I mixed up something while installing the npm modules. Should be fixed now.

joelcoxokc:
@zewa666 I like the installer. I wish I would have seen this earlier. I made something similar. However, I do not have the npm and jspm install functionality running. I say we stay with yours. I am just using child_process.spawn.

joelcoxokc:
Actually I may need to use this code for different things. @EisenbergEffect Can we go ahead and merge this, and I will apply it to some of the things I am using child_process.spawn for. I will also need to write some tests around it.

zewa666:
Glad it's helpful. That's the similar approach I'm using for the generator-aurelia. First I had used child_process.spawn as well, but have changed it because of two reasons.
First it depends on proper Environment Path setups and proper access rights (Windows Run As Administrator) depending on how it's installed. Second if we ever thrive to work without global dependencies this approach suites better since only the required modules (except npm) are used. Didn't wrote any tests so far as I wasn't sure whether this is helpful or not but could definitely add those if needed.

EisenbergEffect:
@joelcoxokc You should have rights to do the merge. Please go ahead.

Issue #11, opened 2015-04-27
Create CLI Logger


joelcoxokc:
Let's create a logger that output messages to the console with nice prefixes. 

The following prefixes are some examples that may be necessary. 

[aurelia]: 
[aurelia] [Error]:
[aurelia] [Success]: 

Colors: 
aurelia -> magenta
Error    -> red
Success -> green

Issue #9, opened 2015-04-27
Export an instance from the main module file.


joelcoxokc:
The main module file "index.js" is currently exporting the constructor. 

This can become an issue when the user wants to require aurelia-cli in multiple places.

This also requires the logic in the bin/aurelia.js to be changed as well, because we are creating a new instance and then passing it to the configFile. 

#Solution

If we would like access to the constructor, we can place the constructor on the prototype, and export a new instance.

joelcoxokc:
I can set the constructor on the prototype one of two ways. 

By convention we see something like the following

```
Aurelia.prototype.constructor = Aurelia
```

However, Some libraries will do the following 

```
Aurelia.prototype.Aurelia = Aurelia;
```

I am going to stick with convention, unless anyone has any objections.

Issue #8, opened 2015-04-25
Separate Library and CLI


EisenbergEffect:
All functionality of the cli should be in a separate library. The cli should just automate this library. This would allow the library to be used directly inside of gulp or any other node-based tool.

Perhaps `aurelia-tools` is the best repo to move the core functionality into?

ahmedshuhel:
Rather than a single separate library could we consider multiple smaller focused libraries? CLI houses different functionalities with different purposes. Let's take `bundling` and `scaffolding` for example. They are very different in nature. So it's probably make more sense to factor them out into two separate libs. 

The modules in the CLI are not very mature yet. So we should probably wait a little more before refactorig.

EisenbergEffect:
Sounds fantastic. Keep working in the current repo and let's just keep this in mind. When things mature, we will talk about how they should be split apart. We'll leave this issue open as a reminder of that.

joelcoxokc:
I have some solutions to this issue, I will be making a pull request soon on.

joelcoxokc:
One of the issues we have is that our commander logic is in the same file that will be required on the library side. I have pulled all the commander logic from the index.js. and placed it into a file called interface.js I figured this would be a valid name, as it is the command line interface.

joelcoxokc:
We also should be exporting an instance from the index file, rather than the constructor. Otherwise, it will be difficult to re access the instance if not stored properly. 

Should I make this another issue?

PR #12, opened 2015-04-27
feat(cli-logger): fix(environment-issues):

joelcoxokc:
This pull request should have been broken up into several. 
The following pull requests will be made much more simple. 

Below is a list of issues resolved and features added

#8 resolve issue by creating an interface file that provides all executable commands, and separates the Library from the CLI tool.

#9 We are now exporting an instance of Aurelia-Cli from the main module, which resolves potential issues from users who may want to use the Library in other places. 

#11 Create a CLI Logger with simple prefixes

### DIFF
--- a/bin/aurelia.js
+++ b/bin/aurelia.js
@@ -1,7 +1,8 @@
 #!/usr/bin/env node
+
 const Liftoff = require('liftoff');
 const argv = require('minimist')(process.argv.slice(2));
-
+const interface = require('../lib/interface');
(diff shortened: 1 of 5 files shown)


zewa666:
Nice work @joelcoxokc, mind renaming the `interface`. A lot of IDEs scream like that's a reserved word. Besides that though this approach is awesome since e.g. my Yeoman generator can easily access the separated library to perform the same functionality :+1:

joelcoxokc:
Yes I am going to reject this PR

Issue #19, opened 2015-04-27
CLI Always requires a local instance


joelcoxokc:
The CLI exits if no local install is found. This prevents some one from creating a new project to begin with. 

Simple Fix. In the executable file, remove the process.exit() when checking for the config file. 

However, We should only provide specific commands if the config file is not found, like Help or New, or Init.

PR #22, opened 2015-04-28
refactor(split-lib-cli): Make clear separation between Lib and CLI

joelcoxokc:
Create a cli interface at lib/cli
Resolve potential symlink issues revolving around linux and windows machines
Create object on process.AURELIA for linux and windows machines
Create object on process.env.AURELIA for other machines that do not have symlink issues
Create process import method for importing local modules

Refactor resolves issue #8 
Refactor resolves issue #19


joelcoxokc:
May I merge this PR

Issue #17, opened 2015-04-27
Feature/ spawn-promise


joelcoxokc:
Create a feature/lib that provides a nice API for spawning commands to child_process, and promises the output. 

This will allow us to spin up child processes and run commands as if the user was running them.
The resolve will return the output that would originally be printed to the terminal. 

We can parse this output and use it to our advantage, by generating code relevant to their environment.

PR #25, opened 2015-04-28
feat(spawn-promise): Spawn commands and promise the output

joelcoxokc:
Create Service for spawning commands and promising the output
Provide ability to pass mulitple commands that run in order

Feature resolves issue #17


joelcoxokc:
Can I merge this?

PR #34, opened 2015-05-08
feat(generate): generates a viewmodel/view

zewa666:
This adds the generator feature to create a viewmodel and a view, together with injection handling. The user may see a preview of the file and cancel creation.


zewa666:
As discussed with @PWKad lets wait for @joelcoxokc to validate this if it aligns with his approach of ARI

PR #51, opened 2015-05-16
fix(jspm): fixes jspm install process

zewa666:
this fixes the installation procedure of jspm via the Node API.


zewa666:
@PWKad that should now also work for your update command

PWKad:
This one shouldn't affect the other PRs and is small so I went ahead and merged it.

Issue #52, opened 2015-05-17
Refactor CLI to provide Simple API usage


joelcoxokc:
I would like to refactor the CLI. 

I am going to remove LyftOff... as this is now becoming useless. LyftOff also creates issues with the environment, when trying to create an external API.

Issue #45, opened 2015-05-11
Create CMD for project environment Setup


joelcoxokc:
Create a command to set up a project environment that includes custom plugin repos and templates for generating code. 

The Environment Architecture

```bash
cwd/
    plugins/
        # contains plugin repositories
    templates/
        # Contains custom templates for generating your own code
    project/
        # Contains the project repository

    # the configuration file that contains the environmental settings.
    # this config file could go in the projects ConfgiFile 
    .aurelia-config.json 
    
   
```

Issue #38, opened 2015-05-11
Aureliafile.js is not extensible


PWKad:
Currently by relying on a JavaScript file as the configuration for the CLI it prevents the ability to 'extend' the functionality of the CLI or make it able to work when upgrading versions.  If the CLI is not upgraded it will throw errors for methods not being implemented.

I think the Aureliafile.js should be converted into a file that is of JSON format so that it is more extensible.

zewa666:
Hmm makes sense. But besides that how about offering a `sanitze` task to fix the Config-file, whatever it's called?

joelcoxokc:
I say we do both!

joelcoxokc:
I actually like the idea of sticking with a JSON file. I will handle this, and make a PR.

Aaike:
i like the sanitize idea also, json is nice and solid , but a js file allows me to comment out lines and make the configuration more dynamic if needed.

joelcoxokc:
Hmm, Good observation. I am going to take a dive into JSPM and see how they are using the config.js. I think it is possible to keep a js file. The only concern I have is speed. It shouldn't affect us for more than a couple milliseconds... However, Let me take a look anyway!

joelcoxokc:
OK I found a solution to using a JS file. I am going to create a PR with a module that contains the functionality to save, read , and write to this config file.

Issue #26, opened 2015-04-28
Consider ES6


EisenbergEffect:
The rest of Aurelia itself is written with ES6/7. Consider setting up a process that would allow the cli to use ES6. This would likely involve configuring babel to compile and generate modules in CommonJS format. I haven't used Babel in this context yet, but I think it would be worth investigating for consistency sake.

joelcoxokc:
Yes I would really like to run this in ES6 however I believe babel or IO.js have a note on their site about not using ES6 if creating node libraries just yet. Otherwise, I'll get started on this one.

EisenbergEffect:
Sounds good. We can keep this in the backlog until we feel comfortable we know how to make it work. If you want to investigate that further go ahead. If not, no big deal. I don't want to distract you all from the feature work.

PR #54, opened 2015-05-18
Refactor/cli

joelcoxokc:
Remove LiftOff
Add ES6/ES7 support
Create command(create): For creating project environments

I have re-structured our code base once again.

Originally I asked yall to place all logic that supports each commands in the `commands/` directory. However, We need to separate the logic required for the CLI and the logic required for the external API.
So, The api logic will be placed in the api directory. If you would like to provide an api to a command, require the API in your command file, and use it there.

No worries about refactoring existing commands, I have already done this.

Any logic that can be shared amongst the API and the Command should be placed in the lib directory.

There may be a few kinks to work out, as this PR makes some drastic changes. So please run your code, and make sure it works properly.

If you find issues, then Let me know, or submit a PR to the branch `refactor/cli`;

This is going to require us to run `gulp watch` while developing, to compile our code.

We will have an src directory. This directory contains the actual ES6 source code.
The output directory is the dist folder.
**NOTE** The output or source directory can/may change.

**NOTE** I have added the bundler to `Api.prototype.bundler` and the installer to `Api.prototype.installer`

**Example**
`./src/lib` will be compiled to `./dist/lib`

Please let me know if you have any objections to this.

We now have three directories that we will be working in
`./src/commands`
`./src/lib`
`./src/api`

This is another big change, which could have also produced some issues.

Please checkout the Branch, and make sure any code you have written runs properly.

Everything you need for the Environment exists on process.AURELIA.

Please use the term cli when using process.AURELIA.
**Example**
```javascript
var cli = process.AURELIA
```

**process.AURELIA**

Contains the some helpers

Most everything that was provided from lyftOff is now on `cli.env`

Important information about process.AURELIA

```javascript
var cli = process.AURELIA
var env = cli.env;

cli.name           // aurelia-cli
cli.moduleName     // aurelia-cli
cli.aurelia        // An instance of `./Aureliafile.js` at aurelia-cli's root
cli.extensions     // possible extensions required for the Aureliafile
cli.pkg            // aurelia-cli's package.json
cli.mainPath       // Path to file, that returns the instance passed to the Aureliafile's function
                   // which is found at the cli's root ./Aureliafile.js
cli.isLaunched     // boolean whether the cli has been launched or not
cli.isAureliaFile  // boolean whether the cli.aurelia has been invoked

cli.root           // function returns path.join('dist', [arguments])
cli.logger         // a simple logger that prefixes and colors our logs
                   Logger.prototype.log  [aurelia]: ## Your message ##
                   Logger.prototype.ok   [aurelia] [Ok]: ## Your message ##
                   Logger.prototype.err  [aurelia] [Err]: ## Your message ##

cli.emit           // bound to program.emit from commanderjs
cli.on             // Promise bound to program.on from commanderjs

cli.launch         // function called in `bin/aurelia` that launches the cli
cli.initialize     // called by cli.launch, but before any local configurations have been checked
cli.ready          // called by cli.launch, Only when and if your local environment has been initialized

cli.import         // function returns requiring the path from the root to your specified module
                   // cli.import('lib/someModule')
cli.exec           // function requires a name parameter, returns a function that when called will require and invoke `command/{name}`
cli.api            // the instance of aurelia-cli api found at api/index.js
cli.store          // config store for reading and writing to the local Aureliafile

env.isConfig       // boolean if local Aureliafile exists
env.isModule       // boolean if local aurelia-cli exists

env.configPath     // path to local Aureliafile
env.configBase     // Base path to local Aureliafile
env.modulePath     // path to local aurelia-cli

env.configFile     // Instance of required local Aureliafile
env.argv           // arguments passed to the cli
env.args           // argv._ an array of arguments passed to the cli.
env.cwd            // The correct cwd
env.ENV            // process.env

```

Resolves issue #26 
Resolves issue #52 
Resolves issue #38 
Resolves issue #45 
Resolves issue #15 
Resolves issue #10


src/lib/config/index.js
```diff
@@ -0,0 +1,143 @@
...
+var defaults = {}
+defaults.config = {
+  "paths": {
+    "*": "dist/*.js",
+    "github:*": "jspm_packages/github/*.js",
+    "npm:*": "jspm_packages/npm/*.js",
+    "aurelia-skeleton-navigation/*": "lib/*.js"
+  },
+  "env":{},
+  "baseURL": '/Users/Shuhel/Workspace/aurelia/skeleton-navigation/'
```
PWKad: Need to remove the baseURL here

src/lib/promise-mkdirp.js
```diff
@@ -0,0 +1,27 @@
...
+   ,Promise = require('bluebird')
+   ,map     = require('lodash/collection/map')
+   ,cli     = process.AURELIA
+   ,mkdirp  = require('mkdirp')
+   ;
+
+
+var logger = cli.import('lib/logger');
+
+function mkdir(dirs, initialDirs){
```
PWKad: Does mkdir here only work with an array of directories?  Should it be able to handle either or implement a second called `mkdirs` instead?

src/lib/spawn-promise.js
```diff
@@ -54,11 +54,16 @@ function SpawnPromise(options) {
...
+        reject('' + data);
+      } else {
+        resolve('' + data);
+      }
       process.stdout.write('' + data);
     });
 
     child_process.on('close', function(code) {
+      // reject('' + code);
       // process.stdout.write(''+code);
```
PWKad: If neither of these are needed should we go ahead and remove them?

src/lib/store.js
```diff
@@ -0,0 +1,89 @@
...
+var _instance;
+
+var Store = (function(){
+
+  function _Store(configPath) {
+    configPath = configPath || process.env.PWD;
+    this.configName = '.aurelia-config.json'
+    this.configPath = configPath +'/'+this.configName;
+    this.configRoot = path.join.bind(path, process.cwd());
+    this.defaultConfig = exists(this.configPath)
```
PWKad: Probably more of a personal preference but obfuscating the value assignment using a ternary operator over a few lines is a bit difficult to read.  Functionally it makes sense.

bin/aurelia.js
```diff
@@ -1,38 +1,128 @@
...
+    .option('-r --remove <remove_path>', 'Remove file path or file from bundle')
+    .option('-l, --list', 'List paths and files included in bundle')
+    .action(cli.exec('bundle'));
+
+  program.command('generate <type>')
+    .alias('g')
+    .description('Generate new file type based on type specified')
+    .option('-n, --name <name>', "Name of the file / class")
+    .option('-v, --view', "Create a view for generated file type")
+    .option('-i, --inject <name>', "Name of dependency to inject")
```
zewa666: This parameter should be parsed as a list, take a look at the previous version:
https://github.com/aurelia/cli/blob/master/lib/program.js#L91

package.json
```diff
@@ -27,18 +27,42 @@
...
     "jasmine-node": "^2.0.0",
-    "injectr": "^0.5.1"
+    "jshint-stylish": "^1.0.2",
+    "karma": "^0.12.31",
+    "object.assign": "^2.0.1",
+    "require-dir": "^0.3.0",
+    "run-sequence": "^1.1.0",
+    "vinyl-paths": "^1.0.0",
+    "yargs": "^3.9.0"
   }
```
zewa666: `gulp-template` is missing in the devDependencies

dist/api/bundler.js
```diff
@@ -0,0 +1,113 @@
...
+
+var api = require('jspm/api');
+var whacko = require('whacko');
+var glob = require('glob');
+var fs = require('fs');
+var url = require('url');
+var path = require('path');
+
+var pluginName = 'view';
+var loader = api.Builder().loader;
```
zewa666: This global declaration needs to be moved into the functions needing this, as this code expects the destination to have a package.json file, which is not happening on the first run

PWKad:
Overall I don't see anything preventing us from merging this PR as a new base but I'd just like to add that I think we need to squash commits more (especially when removing or cleaning up everyone's code before it's been merged) and try to focus on smaller PRs so everyone can get a good feel for the changes.  Thanks for taking the time to add all of the ES6/7 stuff here.  

Any considerations that you think we need to make when using ES6/7 syntax to prevent causing compiler-introduced errors?

zewa666:
In general please let's make sure we add/remove un/necessary semicolons and maybe agree on a certain variable declaration style. Again my preferred one would be like this:

```
// Module requires
var api    = require('jspm/api')
  , whacko = require('whacko')
    ...
  , glob   = require('glob');

// Separate local file requires
var logger      = cli.import('lib/logger')
  , whatever    = cli.import('lib/logger');
```

src/lib/promise-mkdirp.js
joelcoxokc: Resolved

src/lib/config/index.js
joelcoxokc: Resolved

src/lib/spawn-promise.js
joelcoxokc: Resolved

src/lib/store.js
joelcoxokc: Are you talking about `configPath = configPath || process.env.PWD;`

bin/aurelia.js
joelcoxokc: Resolved

package.json
joelcoxokc: Resolved

dist/api/bundler.js
joelcoxokc: I believe the issue is that I am requiring the file as the API.prototype. This file should never be required if it is not needed. I will refector.

joelcoxokc:
@/all 
all in favor of making all lists of variable declaration that are three variables or more look like this, "I".

```
var api    = require('jspm/api')
    , whacko = require('whacko')
      ...
    , glob   = require('glob')
    ;
```

dist/api/bundler.js
joelcoxokc: Resolved

joelcoxokc: Instead of requiring the file as the API.prototype[somefile]

I am creating a getter.

```
Object.defineProperty(API.prototype, name, {
      get: function() {
        return require(p);
      }
    });
    // instead of 
   API.prototype[name] = require(p);
```

bin/aurelia.js
```diff
@@ -1,38 +1,134 @@
...
+  program.version(pjson.version)
+    .on('--help', function () {
+      console.log(['\n'
+        , '    ' + chalk.bgMagenta(' ') + chalk.bgRed('  ') + chalk.bgMagenta(' ') + '\n'
+        , '  ' + chalk.bgRed.black(' aurelia ') + '\n'
+        , '    ' + chalk.bgMagenta(' ') + chalk.bgRed('  ') + chalk.bgMagenta(' ') + '\n'
+        ]).join('');
+    });
+
+  program.command('new <type>')
```
ahmedshuhel: All these command registration should be in a separate file. We will have to amend this every time we add a new command or change the option of a command. It makes more sense if we separate it from the cli initialization.

dist/api/index.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+
+var fs = require('fs');
+var _f = require('fs-utils');
+var path = require('path');
+var cli = process.AURELIA;
+var API = function API() {};
+
+fs.readdirSync(__dirname).forEach(function (file) {
```
ahmedshuhel: The way we are building the API object. There are two aspect of this:
- Every time a user run a command this script will search the folder and build the API
- Any user/human consuming the API would not get a clear picture of what he is consuming
- API is a contract between `Consumer` and `Producer` This should be as explicit as possible. I should not change dynamically.

PWKad:
I've started rebasing changes off of this so unless there is anything breaking I think we need to merge this in so we can get started to get this ready for the release Friday.

ahmedshuhel:
@PWKad I am really not sure about the Friday release. If you have followed yesterday's chat on gitter and I am sure you did, we made some if I put it mildly `serious choices` about the cli infrastructure previously. We have to clean it up and make a stable core. We have to test it ourselves for every command on every platform. 

We have to make sure the code for cli boot up is as lean and stable as possible. The cil init process will be handled by the global installation and the rest that means every commands will be handled by the local per project installation of the cli except `new` command. Only this way we will be able to protect the users and ourselves from burning their/our face with breaking changes that we may make in the early releases. This version will be a serious release that people is actually going to use.

As far as any software project is concern, infrastructure is not the king. It's the actual feature/functionalists that it offers is the king that people actually use. And, we are sorely lacking features right now. But lately we were/are all busy with changing, breaking and re building infrastructure. Some of it we did for the sake of performance(?) and optimization(?) another thing that should not be done so early in the project life cycle. 

Anyway, let's just be focused on the number of quality features we can add from now on leaving the infrastructure and premature optimization out of the way.

We have done some wonderful work together. Let's push it to the next level.

@zewa666 @joelcoxokc

joelcoxokc:
@/all I am going to merge this PR, I've already got some more coming soon that depend on it!

PWKad:
Alright sounds good

bin/aurelia.js
```diff
@@ -1,38 +1,50 @@
...
-  argv: argv,
-  cwd: argv.cwd,
-  configPath: argv.aureliafile,
-  require: argv.require,
-  completion: argv.completion,
-  verbose: argv.verbose,
-  Lyftoff: AureliaCLI,
-  launchFile: 'lib/program.js'
-});
+  , CLI     = require('../dist/cli')
```
ahmedshuhel: This should be `CLI = require(../src/cli')` I suppose ?

bin/aurelia.js
```diff
@@ -1,38 +1,50 @@
...
+  , Liftoff = require('liftoff')
+;
+
+var // Variables
+  AureliaCLI = new Liftoff({
```
ahmedshuhel: var name should be `aureliaCLI` it's an instance of Liftoff not the constructor.

joelcoxokc: Resolved

bin/aurelia.js
joelcoxokc: No because the `../src` is being compiled to `../dist`

PR #71, opened 2015-05-23
feat(update): Add update command

PWKad:
This adds the ability to update official aurelia dependencies in one command

### DIFF
--- a/src/commands/updater.js
+++ b/src/commands/updater.js
@@ -1,0 +1,14 @@
+import * as logger from '../lib/logger';
+import * as updater from '../lib/updater';
+
+var cli = process.AURELIA;
+
+// Updater
+//
+// Executable Command for updating Aurelia
+export function action(options) {
+  logger.log(options)
+  logger.log('Updating Aurelia...');
+  logger.log('-----------------------------------');
+  updater.update(options);
+}
--- a/src/lib/updater/index.js
+++ b/src/lib/updater/index.js
@@ -1,0 +1,120 @@
+import * as logger from '../logger';
+import * as installer from '../installer';
+import * as spawnPromise from '../spawn-promise';
+var util = require('util')
+  , fs = require('fs')
+  , SpawnPromise = spawnPromise.SpawnPromise;
+
+function update(options) {
+  if (options.nuclear){
+    logger.log('Going NUCLEAR!')
+    logger.log('Clearing jspm packages...');
+    moveJSPMPackages(function(){
+      logger.log('Clearing npm packages...');
+      moveNPMPackages(function(){
+        normalUpdate();
+      });
+    });
+  } else {
+    logger.log('Updating normally...');
+    normalUpdate();
+  }
+}
+
+function unbundle(){
+  logger.log('Running jspm unbundle');
+  return SpawnPromise({
+    command: 'jspm',
+    args: ['unbundle']
+  });
+}
+
+function normalUpdate(){
+  var updateCommands = [];
+  var repoList = readConfigJs();
+  unbundle().then(function(){
+    installer.runNPMInstall(function(){
+      logger.log('Successfully npm installed');
+      logger.log('Updating all aurelia libs');
+      repoList.forEach(repo => {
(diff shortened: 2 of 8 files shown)


PWKad:
@joelcoxokc @zewa666 @ahmedshuhel The only issue remaining that I see is jspm clean is run too early once the first promise is resolved.  Gladly accept any ideas for fixing that otherwise we can open an issue to add that as an enhancement to spawn-promise which should resolve that issue.

src/commands/updater.js
```diff
@@ -0,0 +1,14 @@
+import * as logger from '../lib/logger';
```
zewa666: So according to our last discussion should we do this one via `require`?

src/lib/updater/index.js
```diff
@@ -0,0 +1,120 @@
...
+  var updateCommands = [];
+  var repoList = readConfigJs();
+  unbundle().then(function(){
+    installer.runNPMInstall(function(){
+      logger.log('Successfully npm installed');
+      logger.log('Updating all aurelia libs');
+      repoList.forEach(repo => {
+        updateCommands.push({ command: 'jspm', args: ['install', repo]})
+      });
+      SpawnPromise(updateCommands)
```
zewa666: Guess this is what you meant with the synchronous multi execution. Guess we should, as mentioned in the gitter channel, use something like Bluebirds Promise.join feature

src/lib/updater/index.js
```diff
@@ -0,0 +1,120 @@
...
+  });
+}
+
+function moveJSPMPackages(done){
+  fs.rename('./jspm_packages', './jspm_packages_backup', done);
+}
+
+function moveNPMPackages(done){
+  fs.rename('node_modules', 'node_modules_backup', done);
+}
```
zewa666: If you'd like to rewrite those to promises you can use bluebirds promisifyAll feature. https://github.com/petkaantonov/bluebird/blob/master/API.md#promisepromisifyallobject-target--object-options---object

src/lib/updater/index.js
```diff
@@ -0,0 +1,120 @@
...
+          for (var p in v) {
+            if (!v.hasOwnProperty(p))
+              continue;
+            cfg[c][p] = v[p];
+          }
+        }
+        else
+          cfg[c] = v;
+      }
+    },
```
zewa666: Phew that one's really hard to read. Now this is only my opinion so feel free to ignore if you don't like it but I'd add braces even on single step if's as it makes the code more readable. Next the variable naming convention makes understanding this section quite hard, so longer more descriptive names could help.

src/start.js
```diff
@@ -127,4 +127,26 @@ export function start() {
...
+          , info : 'Blow it up'
+          , required: false
+        },
+      });
+
+      // logger.log('  Examples:');
+      // logger.log();
+      // logger.log('    $ aurelia plugin -a i18n');
+      // logger.log('    $ aurelia p -r i18n -c');
+      // logger.log();
```
zewa666: Are those comments meant as an example output of the --help method above?

src/lib/updater/index.js
PWKad: Yeah most of this is borrowed from [jspm config loader](https://github.com/jspm/jspm-cli/blob/master/lib/config/loader.js#L68-L92) for now until I can find a way to take advantage of that loader through their api.

src/start.js
PWKad: Good question, I had that there uncommented before but I noticed that all of the examples were commented above in the same file.  Actually while reading it those are from the plugin anyway so need to remove.

joelcoxokc: I have actually not been adding braces around simple conditions... I think braces are only necessary if the condition block is large.

for example I wouldn't use braces here

```javascript
if (1 === num)
    return 'yes';
```

joelcoxokc: I like this PR. I would like to merge. 

Or would you rather wait for my PR with the updated spawnPromise?

joelcoxokc:
@PWKad Would you like me to convert this PR to our new API?

PR #73, opened 2015-05-24
fix(cmd-new-init)

### LINKED ISSUE #62: command-init --env installs on every execution.
When `init --env` is executed, it will install a new project. We need to first check if a project already exists. Then prompt the developer on whether he would like to install a project or not.

### LINKED ISSUE #69: command-init not running properly

joelcoxokc:
Refactored cli, placing everything on env
Create prompts for new command

fix resolves #69 
fix resolves #62

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ bower_components
 .idea
 .DS_STORE
 *.swp
+.blog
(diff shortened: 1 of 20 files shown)


joelcoxokc:
Merging PR #75 will automatically merge this.

PR #80, opened 2015-05-26
feat(new-commander): Create new commanderjs

joelcoxokc:
Create new API for creating commands, replaces commanderjs


ahmedshuhel:
Please exclude the `dist` folder for the PR. It is crating a lot of noise. Hard to follow the code. We only build the dist when we will make a public release and will check in then other wise we won't.

bin/aurelia.js
```diff
@@ -8,7 +8,7 @@ process.env.INIT_CWD = process.cwd();
 var DEV_ENV = parseInt(process.env.AURELIA_CLI_DEV_ENV, 10);
 
 const cli = new Liftoff({
-  name: 'aurelia',
+  name: 'aurelia-cli',
```
joelcoxokc: @ahmedshuhel Liftoff wants the name of the package in node_modules `aurelia-cli` is the correct name

bin/aurelia.js
```diff
@@ -56,17 +56,15 @@ function handleCommands(env) {
   }
 
   aurelia.init({
-    env : env
+      env : env
+    , argv: argv
```
joelcoxokc: @ahmedshuhel I need access to argv before we run aurelia.run()

So I am passing it on to the config.

Yes I know it's on config.env.argv, 
This should be a flat and easily access property.

bin/aurelia.js
```diff
@@ -56,17 +56,15 @@ function handleCommands(env) {
...
   });
 
-  if (env.configPath) {
+  if (env.configPath && env.modulePath) {
     require(env.configPath);
```
joelcoxokc: If we `require(env.configPath)`

The Aureliafile contains.

```
require('aurelia-cli');
```

This does not work. if the modulePath does not exist.

bin/aurelia.js
```diff
@@ -56,17 +56,15 @@ function handleCommands(env) {
...
     logger.info('Using Aureliafile: %s', env.configPath);
   } else {
     logger.log('warn', 'Aureliafile not found');
   }
-
-  process.nextTick(function() {
-    aurelia.run(process.argv);
-  });
+  aurelia.run(argv);
```
joelcoxokc: Removed process.nextTick()

We are already in a process.nextTick()

src/index.js
```diff
@@ -1,51 +1,75 @@
...
-
 import BundleCommand from './commands/bundle';
 import InitCommand from './commands/init';
 import NewCommand from './commands/new';
 
 class Aurelia {
+
   constructor() {
     this.commands = {};
+    this.aliases  = {};
```
joelcoxokc: @ahmedshuhel 
Added this.aliases
This allows me to not have to loop through the commands for each argv.

src/index.js
```diff
@@ -1,51 +1,75 @@
...
 
   init(config) {
-    this.config = config;
-    let bundle = new BundleCommand(program, this.config, this.logger);
-    let init = new InitCommand(program, this.config, this.logger);
-    let newCmd = new NewCommand(program, this.config, this.logger);
+    this.program      = new Program(config);
+    this.store        = new Store(config);
+    this.config       = config;
+    this.config.store = this.store;
```
joelcoxokc: @ahmedshuhel 
Added store to config. I don't want to keep it here, I am working on another way of using it.

src/lib/spawn-promise.js
```diff
@@ -0,0 +1,70 @@
+var Promise = require('bluebird');
```
joelcoxokc: We lost this file, some other lobs depend on it.

ahmedshuhel:
@joelcoxokc still see the `dist` files in the PR. Please rebase it on the master branch.

joelcoxokc:
I have rebased. Let me see what I can do.

PWKad:
Are there more breaking changes to the cli api in this PR or is it stabilized yet?  I need to update and get another PR in.

joelcoxokc:
Yes this one introduces a new api for commander... I've like to stop
refactoring this and move along, so hopefully!

bin/aurelia.js
ahmedshuhel: @joelcoxokc We have resolved these in the previous PR. And as far as I know It's working correctly. Suppose when use executes `new` command it does not have the `moduelPath` so the global version is being executed. Here is the code from the `master` branch:

```javascirpt
if (!env.modulePath) {
    logger.log('warn', 'Local aurelia installation not found!');
    aurelia = require('../index');
  } else {
    aurelia = require(env.modulePath);
  }
```

bin/aurelia.js
ahmedshuhel: @joelcoxokc  `process.nextTick()` make sure the `Aurelia` object is instantiated properly before `aurelia.run` executes.

bin/aurelia.js
ahmedshuhel: @joelcoxokc this is been fixed on the master branch.

ahmedshuhel:
@joelcoxokc I see you have some issues here that is related to the the PR 79 that is already merged and working properly(tested the command on windows, linux and mac) but we  are trying to alter the behavior with this PR. Why do we need that? As we have talked the other night, `new commander` should not introduce breaking changes rather it should have all the facilities and api that https://github.com/tj/commander.js has plus some new one. We were talking about making your `new commander` an npm package just like `commander` so that we can just npm install it.

PR #81, opened 2015-05-26
feat(generate): Add generate according to new architecture

zewa666:
This adds the generate command according to the new architecture, as well as some unit tests


ahmedshuhel:
This looks awesome to me. :+1:

PR #94, opened 2015-06-03
fix(vm): fixes multiple inject decorators to single

zewa666:
the template will now use one inject decorator and apply multiple targets as attributes

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,8 +93,7 @@ import { inject } from 'aurelia-framework';
 import { Element } from 'aurelia-framework';
 import { HttpClient } from 'aurelia-framework';
 
-@inject(Element)
-@inject(HttpClient)
+@inject(Element, HttpClient)
 export class YOURNAME {
   hello = 'Welcome to Aurelia!';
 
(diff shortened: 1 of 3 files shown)


PWKad:
:+1: From me, assuming syntactically everything is correct this looks fine.

joelcoxokc:
This has been created... Checkout @EisenbergEffect and my conversation... and the docs I have been posting about the commander API. I actually Like your implementation better than mine... We should probably consider using yours in `aurelia-command`. However. Rob suggested that we try to use the DI. What do you think @zewa666

EisenbergEffect:
FYI That generated code is still not correct. `Element` is not in aurelia-framework. That is a browser global. Also, if you use the `@inject` decorator, you will need to import that from aurelia-framework.

zewa666:
Damn need to make better reviews before commiting stuff. Regarding imports, I could create an exception for Element, but other then that how would we know where to import from? So the idea was to always use the framework and let the user fix it afterwards.

EisenbergEffect:
Hmmm. There might be lots of exceptions.

zewa666:
@EisenbergEffect the `inject` object was already imported in case of injection used, dunno if you saw something else. Have added now the exception for the global Element object.

As said not sure whether we have a chance to automatically know where things come from, especially also if the user injects local files. So I guess we'll have to have our users fix the import paths.

ahmedshuhel:
:+1: from me too.

Issue #31, opened 2015-05-02
Consider creating an "update-aurelia" command


jdanyow:
After a release people need to manually `jspm install` each aurelia dependency, then `jspm clean` to tidy things up.  Should a cli command be added to automate this?

PWKad:
I agree with this going through this issue now with updating a project and it's definitely a hassle.  I'm going to self-assign this one and PR when I get a good solution.

PWKad:
@jdanyow What do you think about this - 

1. `jspm clean`
2. Iterate over the properties in package.json in the dependencies list
3. Call `jspm install {depName}` for each one

PWKad:
As mentioned in the gitter I think we also need to add a `jspm unbundle` step in there

jdanyow:
I think clean is the last step- otherwise sounds perfect

PWKad:
Any thoughts on a `-nuclear` option here as well?  Deletes all local package folders and reinstalls everything.

jdanyow:
that would be great- this option should probably clear the `map` property of the `config.js` file prior to running `jspm install`

joelcoxokc:
@PWKad Keep in mind, you can use spawn-promise, to run jspm commands from the cli

joelcoxokc:
This way the developer doesnt have to run jspm. Let me know if you need an example!

zewa666:
Just keep in mind that this way you'd require the user to have jspm installed globally. Ok in our case it's no biggy since we have a hard dependency on that module. As an alternative JSPM also provides a nice API as npm module. Here is an example https://github.com/aurelia/cli/blob/master/lib/installer.js#L65-L68

EisenbergEffect:
If at all possible, I think we should try to rely on the npm modules. That will make it easier for the editor tooling to use this and not have to worry about external installation of command line tools.

PWKad:
@EisenbergEffect If we rely on NPM only that will prevent most users who are using JSPM from being able to use this no?  I wonder if we should just target the ~80% (made up math) for now and add flags and functionality later for npm only or w/e package managers when we have that requirement.

EisenbergEffect:
I'm not sure I follow. Can't we use jspm's npm module that provides its capabilities as an api? or will that not cover the scenarios we need?

PWKad:
[...]

PR #95, opened 2015-06-05
feat: update command @PWKad #31

ahmedshuhel:


ahmedshuhel:
@aurelia/cli-team I will rebase this again as soon as the core gets settled.

PWKad:
This looks good to me as well.  I think the nuclear option can probably be removed but we can do so after confirming no one has issues with the standard update.

ahmedshuhel:
@PWKad @zewa666 @joelcoxokc Rebased it. Please review.

zewa666:
Ive noticed the Promise.all use instead Promise.join. Is that OK?

ahmedshuhel:
@zewa666 I am not sure.

ahmedshuhel:
https://github.com/petkaantonov/bluebird/blob/master/API.md#all---promise
@PWKad  Please confirm whether this is the behavior we want.

PWKad:
Last I saw @zewa666 had mentioned to use Promise.join and @joelcoxokc was refactoring the `spawn-promise` module to use it.  I believe you guys had two different PRs and you refactored it to not use that and instead use Promise.all and something else.  I'm good with either method, I've not been testing it since you took this issue over.

ahmedshuhel:
https://github.com/petkaantonov/bluebird/blob/master/API.md#promisejoinpromisethenablevalue-promises-function-handler---promise

Promise.join works for a fixed set of ops but performance wise superior. But, for us here as the number of modules to update are not fixed, `all` seems appropriate. I would request you to give it a run and see if it works way it should.

joelcoxokc:
I think we moved to promise.join because we want the first to resolve
before the second.

ahmedshuhel:
@joelcoxokc https://github.com/ahmedshuhel/cli/blob/update-command/src/lib/updater/index.js#L45

This is installing `arbitrary` number of jspm package in parallel. It's like running `jspm install` from multiple shell at the same time. One install does not depend on the  other. So `order` is not important here as much as I understand.

PWKad:
@ahmedshuhel @joelcoxokc @zewa666 If this is working I'd like to merge it in before the next release as it takes a long time to explain how to update each repo to everyone :)

ahmedshuhel:
@PWKad I believe it's working and we can get this In. Could you please give it a try and let us know the result?

PR #98, opened 2015-06-08
refactor(commands): Apply new aurelia-command API syntax to existing commands

joelcoxokc:
Apologies for the multiple commits. Apparently I hadn't deleted the old branch, so it caused some issues.


EisenbergEffect:
It seems that the cli has evolved significantly since the time of this PR.

PR #106, opened 2015-06-17
When inject is set to true, creates bundle in the appropriate path and inject bundle with correct `moduleName`

ahmedshuhel:


joelcoxokc:
Good To GO

PWKad:
:+1:

PR #133, opened 2015-07-10
feat(bundler): allow specifying baseURL manually

### LINKED ISSUE #132: Change baseURL
We should be able to specify the `baseUrl` that the bundler uses overriding the one it gets from the loader.  
```
aurelia.command('bundle', {
    js: {...},
    template: {...}
}, {
    force: true,
    baseUrl: 'some/other/base/for/bundling'
});
```

olanod:
I haven't tried it yet but it looks like its the only change needed to be able to specify the baseURL for the bundler.
fixes #132

### DIFF
--- a/src/lib/bundler.js
+++ b/src/lib/bundler.js
@@ -14,7 +14,7 @@ function bundleJS(moduleExpression, outfile, options) {
 export default function bundle(config, bundleOpts) {
 
   var loader = api.Builder().loader;
-  var baseURL = loader.baseURL;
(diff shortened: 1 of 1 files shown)


ahmedshuhel:
Wouldn't it be nice if you try it first?

olanod:
Yes, it was just a quick proof of concept ... and it doesn't work :disappointed:. Now I have a working example but I'm still missing some details so I'll update this PR once it's done.

olanod:
I'll make a new PR when I get it right.

PR #140, opened 2015-07-29
Specify moduleName without version in aureliafile.

ahmedshuhel:
As of now, we are directly using `systemjs-builder` for bundling js modules. For `systemjs-builder` a module is not just it's name rather it's `name + version`. So, a valid module name would look like: `github:aurelia/templating-binding@0.12.0`. When we create the `aurliafile` configuration for bundling js modules, we specify both (name + version). This creates a maintenance issue. Whenever we update the project dependencies, we potentially invalidate the `aureliafile` as update tend to change the version number.  To solve this issue, we need to specify the configuration without the version number and internally develop a lookup mechanism for resolving correct moduleName with version from `config.js` for `systemjs-builder`.

One detail to remember is that if we find multiple versions where one is not specified in the aureliafile, we should throw an exception. The exception message should be clear about that is happening and recommend the possible solution of running `jspm clean` to remove old versions.

Work in progress.
@EisenbergEffect

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,10 +85,29 @@ aurelia bundle
   - `paths` are important to consider when specifying `module names` 
 
 ### Template bundle option
-- Option
-  - *inject:* injects a `<link aurelia-view-bundle rel="import" href="bundle_name.html">` at the end of the body tag of  `index.html` to include the bundle file.
 
-- Note:
+#### Options
+  - *inject:* injects a `<link aurelia-view-bundle rel="import" href="bundle_name.html">` at the end of the body tag of  `index.html` to include the bundle file.
+  - *inject* can be an object too. 
+
+```javascript
+  template: {
+    "dist/app-bundle": {
+      pattern: 'dist/*.html',
+      options: {
+        inject: {
+          indexFile: 'index.html',
+          destFile: 'dist/index.html'
+        },
+      }
+    }
+  }
+```
+  - *indexFile* : Path of the `index.html` relative to baseURL. If not specified default is `baseURL/index.html`.
+  - *destFile* :  Path of the new html file with the injected link. When not specified defaults to `indexFile`.
+  
+#### Note:
+
   - Globs template relative to `baseURL`
   - Glob files. 
   - Multiple glob pattern can be specified as `['dist/about/*.html', 'other/**/*.html']`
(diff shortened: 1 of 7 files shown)


EisenbergEffect:
Thanks for creating the details issue description! That sounds good. Keep me posted and let me know when you want me to review.

ahmedshuhel:
@EisenbergEffect  Initial implementation is done. Please start reviewing.. After a few more refactoring sessions, this feature should be solid.

ahmedshuhel:
The command only searches the `config.js` 's map for locating full moduleNames with version. Is there any other thing we can include here as source?

ahmedshuhel:
@EisenbergEffect Please continue reviewing. 

I have completed:

- Feature 2 Specify moduleName without version
- Feature 5 : support creating new html file with the injected link without overwriting original  index.html
- Option to specify the `index.html` file path/name relative to baseURL.

EisenbergEffect:
Great! Feel free to merge and release as soon as you get the readme updated. Please let me know if there are any breaking changes. After this is all released, let me know what our next set of tasks are. I can help prioritize. I've got a list of things we will need to add as well...

ahmedshuhel:
Need to implement these features before release:
- `Un bundle` command : Clear injected js bundle from config.js and remove `aurelia-view-bundle` links from `index` html file.

- Providing a way to clean/clear invalid injected bundles both from `config.js` and html file. We may implement it as part of `unbundle` command as:

```bash
aurelia unbundle --clear-invalids
```

And, every new `aurelia bundle` should also `clean` invalid injections first.

Merging this one.

ahmedshuhel:
I should also mention that there should be no breaking changes so far. I kept that in mind while developing the new features.

e.g. Previously template bundle config was like:

```javascript
template: {
    "dist/app-bundle": {
      pattern: 'dist/*.html',
      options: {
        inject: true
      }
    }
  }
```

Now it is:
```javascript
  template: {
    "dist/app-bundle": {
      pattern: 'dist/*.html',
      options: {
        inject: {
          indexFile: 'index.html',
          destFile: 'dist/index.html'
        },
      }
    }
  }
```

But the previous version will also work as the `indexFile`, `destFile` all defaults to `baseURL/index.html`

PR #143, opened 2015-08-07
feat(plugin): add plugin install command

zewa666:
this adds the basic command to install plugins upon asking the user which one to install.

### DIFF
--- a/src/lib/installer/index.js
+++ b/src/lib/installer/index.js
@@ -1,8 +1,8 @@
-var ghdownload = require('download-github-repo')
-  , GitHubApi  = require("github")
-  , Promise = require('bluebird')
-  , request = require('request')
-  , jspm = require('jspm');
+import download from 'download-github-repo';
+import GitHubApi  from 'github';
+import Promise from 'bluebird';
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
@ahmedshuhel Please review. Thanks!

src/lib/installer/index.js
```diff
@@ -1,6 +1,8 @@
 var ghdownload = require('download-github-repo')
```
ahmedshuhel: Could we change these `require` to es6 `import` statements and the module.exports to `export`? I know we have other lib modules that have still these `require`s. We may take an approach of converting an old module as we need a change there.

ahmedshuhel:
As always @zewa666 's code is clean and looks good. I will give the command a try.

src/lib/installer/index.js
zewa666: Sure no Problem, not at home but will change that tomorrow

zewa666:
@ahmedshuhel I'm sorry it took me so long but now the changes have been supplied and I think it's ready for review

PR #144, opened 2015-08-07
Robust View Bundling - Following SystemJS Injection Pattern

ahmedshuhel:
Until now, our view bundle injection is just adding a `<link aurelia-view-bundle href="bundle-file-path" /> ` in `index` file. This works but we can do better. We want something like SystemJS bundle injection. When SystemJS tries to load a module, it searches  `bundle: []` section in `config.js`  and if the module is found there, SystemJS loads the entire bundle. This way multiple bundle may be injected in `config.js` and loader will load as needed. A large app with multiple  `sections` really benefits from this as we can create bundle per section. This helps `initial loading` faster because as we are not loading everything upfront.

This PR is aiming to achieve the same pattern for  `Aurelia Views`

### DIFF
--- a/src/lib/bundler/template-bundler.js
+++ b/src/lib/bundler/template-bundler.js
@@ -7,11 +7,11 @@ import globby from 'globby';
 import utils from 'systemjs-builder/lib/utils';
 
 export function bundleTemplate(pattern, fileName, options, bundleOpts) {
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
Hold off on this. See my notes to you in the gitter.

PR #147, opened 2015-08-12
feat(docs): add readme for plugin command

zewa666:
an additional readme text for the plugin command


zewa666:
@ahmedshuhel sry I forgot to add some info about the plugin to the Readme. Here it is.

PR #154, opened 2015-08-17
Refector bundling bits out and create a separate lib

ahmedshuhel:
@EisenbergEffect  
The bundling bits completely factored out. Created a lib called `aurelia-bundler` and published to npm. CLI is now using that lib instead. 

Ready to merge.


EisenbergEffect:
@ahmedshuhel Do I need to create a repo for you? Should I just call it "bundler" then you can push your code there and update this PR to reference that. Does that sound good?

EisenbergEffect:
@ahmedshuhel Should this be closed?

ahmedshuhel:
Yes, as we are not moving ahead with cli, we can close this. And the other one too.

nomack84:
@EisenbergEffect @ahmedshuhel `Yes, as we are not moving ahead with cli` What's this means?

EisenbergEffect:
@nomack84 We are moving our bundling code into a separate library so that it can be used directly from Grunt/Gulp/etc. Then we are going to move to a standard set of gulp tasks that do everything our cli did. That will make it easier for people to re-use our bundler and other tools in a variety of contexts...picking the tool they want rather than being forced to use our CLI. We'll also have an online repository or starter projects and other services to help people get started.

Issue #155, opened 2015-08-19
ghdownload is not defined


Depool:
'aurelia new project' fails having both aurelia-cli 0.3.1 and 0.3.0 installed on Windows.
Getting 'ghdownload is not defined' in /aurelia-cli/dist/lib/installer/index.js.
Using aurelia-cli 0.2.1 went ok

zewa666:
Hi there, yep sry for the stupid mistake ... will get that fixed soon, PR is already created

PR #156, opened 2015-08-19
fix(download): fixed wrong variable usage

zewa666:
fixes the download issue with ghdownload

fixes issue https://github.com/aurelia/cli/issues/155


zewa666:
@ahmedshuhel sorry dunno how that stupid error happened ... this should fix it. You can try it out with `aurelia new`

Issue #161, opened 2015-08-25
Systemjs-builder 0.13.4 breaks template bundling


JeroenVinke:
System-js builder 0.13.4 had some changes to `getCanonicalName()`(https://github.com/systemjs/builder/commit/290b84bfb90a8786900e8652671a0aab36f7e5fc)

These changes changed the behaviour and caused paths in the templating bundle (`app-bundle.html`) to contain the dist folder (or other `*` mapping in `config.js`)

Example: `<template id="dist/components/projectDetailView/projectDetailView.html!view">`

If I install systemjs-builder 0.13.2 the problem is gone. Not sure why 0.13.4 gets installed though.

Possibly also causing this issue: https://github.com/aurelia/skeleton-navigation/issues/146

JeroenVinke:
Workaround:

go into the node_modules\aurelia-cli folder and make this change to `package.json`: https://github.com/JeroenVinke/cli/commit/829db74e7da5418d6a5a4d52b819a4bb4b4c8db7

then run npm install. To confirm you can check if `node_modules\aurelia-cli\node_modules\systemjs-builder\package.json` has the version `0.13.2`

ahmedshuhel:
Thanks for reporting and the PR.

EisenbergEffect:
@ahmedshuhel Has this been addressed? Can we close this?

ahmedshuhel:
@EisenbergEffect I think it's resolved already. We are using systemjs-builder v0.13.5 now. Hopefully @JeroenVinke is using aurelia-bundler now with latest version of Aurelia. 

@JeroenVinke Feel free to open an issue [here](http://github.com/aurelia/bundler/issues), if you still have the problem.

JeroenVinke:
aurelia-bundler works great :smile:

PR #162, opened 2015-08-25
Fix(dependency): Lock down systemjs-builder version

JeroenVinke:
Fixes https://github.com/aurelia/cli/issues/161

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "mkdirp": "^0.5.1",
     "npm": "^2.10.1",
     "request": "^2.60.0",
-    "systemjs-builder": "^0.13.0",
+    "systemjs-builder": "0.13.2",
     "whacko": "^0.18.1",
     "winston": "^1.0.0"
   },


cehoffman:
Changing the installed version to 0.13.2 did not fix the problem for me.

cehoffman:
So I tracked this down. The problem continues to occur despite this pull because jspm specifies at least 0.13.4 version of systemjs-builder. Hence when you follow @JeroenVinke's directions form issue #161 to change the package.json dep in the installed aurelia-cli the jspm dependency was configured to use the aurelia-cli systemjs-builder.

In summary, locking down systemjs-builder from aurelia-cli alone does nothing to fix the problem. The best that can be done is to specify systemjs-builder of 0.13.2 and a jspm version that has systemjs-builder dependency that is satisfied by 0.13.2.

EisenbergEffect:
We are moving away from the CLI and to a bundler library that can be used from a variety of context. If there is an issue can you please report it on the new bundler repo? Thank you! cc @ahmedshuhel

PR #182, opened 2016-06-22
fix(generators): fix incorrect message in value converter generator

atsu85:
Fix incorrect messsage:
```cmd
>au generate value-converter

What would you like to call the generator?
```


EisenbergEffect:
I think I made a few "copy/paste" errors ;)

EisenbergEffect:
Releasing in just a minute.

Issue #192, opened 2016-06-24
Trying to get aurelia-i18n running with typescript


masterpoi:
I've added this to aurelia.json

``` json
       {
            "name": "i18next",
            "path": "../node_modules/i18next/dist/umd",
            "main": "i18next"
          },
          {
            "name": "i18next-browser-languagedetector",
            "path": "../node_modules/i18next-browser-languagedetector/dist/umd",
            "main": "i18nextBrowserLanguageDetector"
          },
          {
            "name": "i18next-localstorage-cache",
            "path": "../node_modules/i18next-localstorage-cache/dist/umd",
            "main": "i18nextLocalStorageCache"
          },
          {
            "name": "i18next-xhr-backend",
            "path": "../node_modules/i18next-xhr-backend/dist/umd",
            "main": "i18nextXHRBackend"
          },
          {
            "name": "aurelia-i18n",
            "path": "../node_modules/aurelia-i18n/dist/amd",
            "main": "aurelia-i18n"
          }
```

however the package name specified in for instance i18next-xhr-backend is actually i18nextXHRBackend:

``` javascript
(function (global, factory) {
  typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
  typeof define === 'function' && define.amd ? define('i18nextBrowserLanguageDetector', factory) :
  (global.i18nextBrowserLanguageDetector = factory());
}(this, function () { 'use strict';
```

So these imports

``` typescript
import * as XHR from 'i18next-xhr-backend';
import * as Cache from 'i18next-localstorage-cache';
import * as LanguageDetector from 'i18next-browser-languagedetector';
```

import `undefined` since the package names don't match.

If i change the imports to 

``` typescript
import * as XHR from 'i18nextXHRBackend';
```

the tracer won't find src/i18nextXHRBackend.js

If i change the declarations like:

``` json

                    {
            "name": "i18nextXHRBackend",
            "path": "../node_modules/i18next-xhr-backend/dist/umd",
            "main": "i18nextXHRBackend"
          },

```

the package analyser uses the wrong directory to try to find package.json "../node_modules/i18nextXHRBackend/package.json".

There doesn't seem to be a way to change this behavior of using the name of the package to find the package.json file.

Is this an aurelia-thing (i18n, cli?)  or an i18next thing?
One way i can see this issue being resolved is to be able to override the path the analyser looks for package.json, but not sure if that is the most elegant way.

EisenbergEffect:
[...]

PR #193, opened 2016-06-24
fix: Make sure to use path configuration to resolve package.json 

masterpoi:
Make sure to use path configuration to resolve package.json when set

https://github.com/aurelia/cli/issues/192


masterpoi:
This is not right at all. Will make another attempt on monday.

PR #204, opened 2016-06-29
enhancement: Connect History API Fallback

alsoicode:
- Added `historyApiFallback()` function to middleware for `serve` task.

Adding the [connect-history-api-fallback](https://github.com/bripkens/connect-history-api-fallback) allows paths added via PushState to be accessed directly when browsing by typing the path into the querystring, e.g. `/contact` and also when BrowserSync reloads the application due to changes.

This makes it a lot more convenient than having to navigate back to the page you were on when your app reloads.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
   "dependencies": {
     "aurelia-dependency-injection": "^1.0.0-beta.1.2.0",
     "aurelia-polyfills": "^1.0.0-beta.1.1.2",
+    "connect-history-api-fallback": "^1.2.0",
     "npm": "^3.8.6"
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
   "dependencies": {
     "aurelia-dependency-injection": "^1.0.0-beta.1.2.0",
     "aurelia-polyfills": "^1.0.0-beta.1.1.2",
+    "connect-history-api-fallback": "^1.2.0",
```
EisenbergEffect: Can you revert this change to the package.json. It's not the cli that needs this dependency, but rather the developer's project. We set that up a different way and I can add it if you can remove this.

alsoicode: Sure is no problem at all.

EisenbergEffect:
Thanks!

PR #229, opened 2016-07-13
fix(config) : make sure that moduleId is set in config

masterpoi:
If a library in an extra bundle (not vendor-bundle) is referenced in the app code, the bundle config/metadata injected in vendor-bundle is missing the moduleIds (being emitted as null). (Which causes the module files not being loaded correctly at run time). This change fixes that.


EisenbergEffect:
Thank you!

PR #251, opened 2016-07-29
experimental install command

kristianmandrup:
Please have a look at this experimental install command feature.
I haven't yet been able to test since the project doesn't seem to provide any guidelines for how to test the cli on a project.

I tried to `npm link` it so far to no avail...

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,42 @@ There are many ways to style components in Aurelia. The CLI sets up your project
 Bear in mind that you can always configure things any way you want by modifying the tasks in the `aurelia_project/tasks` folder.
 For styling purposes, you can modify the `process-css.js` file.
 
+### Install libraries
+
+*Experimental WIP*
+
+`au install <library>`
+
+Example:
+
+`au install bootstrap`
+
+Requires an entry for the library in `lib/commands/install/registry.json`
+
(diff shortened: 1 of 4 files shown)


kristianmandrup:
I'd like to propose a way to install components (view models with views) as separate artefacts.
I first thought about a way to leverage node for this via `node_modules` but I don't think that would work. View models are resolved on the server runtime, so we need to ensure view models and related resources are bundled with the main app.

Reusable/external Aurelia components could be installed under `aurelia_components` in the project root to not be mixed up with node modules.

I think `aurelia.json` is the key, specifically `bundles/app-bundle.js/source`. 

```js
    "bundles": [
      {
        "name": "app-bundle.js",
        "source": [
          "[**/*.js]",
          "**/*.{css,html}",
          "../aurelia_components/**.{css,html,js}"
        ]
      },
```

With regard to dependencies, each such component could come with its own `aurelia.json` file which is deep merged with the application `aurelia.json` file. Same goes for the `package.json` file (exact details how are still unclear, need experimentation).

kristianmandrup:
Please ignore this. I'm creating a separate `aurelia-installer` project. Learned a lot since then ;)

PR #266, opened 2016-08-09
chore(linting): fix eslint warnings in resources

BBosman:
After creating a new project with the Aurelia Cli, I was looking through the code it generated to see what it does. One thing I noticed that `eslint`, with the rules provided by the Cli itself, was giving some errors and warnings on the generated code.

This PR fixes those errors and warnings.

The codebase contains more `eslint` errors and warnings, but I've intentionally scoped this PR to just the resources for now as that's the only code output to client projects as far as I can tell.

### DIFF
--- a/lib/resources/generators/attribute.js
+++ b/lib/resources/generators/attribute.js
@@ -26,7 +26,7 @@ export default class AttributeGenerator {
   }
 
   generateSource(className) {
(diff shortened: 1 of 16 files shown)


EisenbergEffect:
Awesome. Thank you!

PR #268, opened 2016-08-09
chore(dependencies): update npm

BBosman:


EisenbergEffect:
There's a merge conflict but I think we don't really need to do this since the semantic versioning will cause the update upon install.

PR #271, opened 2016-08-09
fix(build): create ouput directory when missing

### LINKED ISSUE #241: au build fails if output folder does not exist

* **Library Version:**
0.16.1

* **Operating System:**
Windows 10

* **Node Version:**
4.4.7

* **NPM Version:**
3.10.5

* **Browser:**
n/a

* **Language:**
TypeScript 2.0.0

**Current behavior:**

I changed in aurelia.json:
platform.output = 'build'
build.targets[0].output = 'build' 

`au build` fails with log-output:

`
Writing vendor-bundle.js...                                                                                                                 
{ uid: 8,                                                                                                                                   
  name: 'writeBundles',                                                                                                                     
  branch: false,                                                                                                                            
  error:                                                                                                                                    
   { [Error: ENOENT: no such file or directory, open 'c:\Users\[...]\build\app-bundle.js']        
     errno: -4058,                                                                                                                          
     code: 'ENOENT',                                                                                                                        
     syscall: 'open',                                                                                                                       
     path: 'c:\\Users\\[...]\\build\\app-bundle.js' },                                        
  duration: [ 17, 887043932 ],                                                                                                              
  time: 1468919595509 }                                                                                                                     
[...]                                        
`

There is no error, if I create the folder 'build' before 'au build'.

Should work without 'build' (or other configured target) being present //
the missing folder should be created by aurelia-cli (/gulp-task).

**What is the motivation / use case for changing the behavior?** 
CI-Server checks out code -> no 'build'-Folder present.

BBosman:
Fixes #241.

It uses a small wrapper around `fs.writeFile` to recursively create the directory it wants to write to when it doesn't exist.

The `filendir` module (and its dependency `mkdirp`) are small, so if you don't want to take an external dependency on them, we could inline them.

### DIFF
--- a/lib/file-system.js
+++ b/lib/file-system.js
@@ -1,5 +1,7 @@
 "use strict";
 const fs = require('fs');
+const nodePath = require('path');
+const mkdirp = require('./mkdirp').mkdirp;
 
 exports.exists = function(path) {
   return new Promise((resolve, reject) => fs.exists(path, resolve));
@@ -13,6 +15,15 @@ exports.mkdir = function(path) {
     });
   });
 };
+
+exports.mkdirp = function(path) {
+  return new Promise((resolve, reject) => {
+    mkdirp(path, error => {
+      if(error) reject(error);
+      else resolve();
+    });
+  });
+}
 
 exports.readdir = function(path) {
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
I think we'll inline these as part of our file-system module and then use it that way. We can provide attribution to the original author.

BBosman:
I've updated the PR and inlined part of `mkdirp`.

PR #282, opened 2016-08-15
feat(bundler): allow the user to add custom configuration options

Jenselme:
This allows the user to add an object under `build.loader.config` to
pass additional parameters to the loader. For instance if you need to
increase the timeout for requirejs, you can add this object under
`build.loader.config`:

```
{
  "waitSeconds": 60
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,23 @@ Sometimes you can't get a library to work with the module loading system. That's
 }
 ```
 
+#### Configuring the loader
+
+You can configure the loader by adding a `config` key to `build.loader` with the options you want to add. For instance, if you want to increase the timeout for requirejs, you would do this:
+
+```
+"build": {
+    "loader": {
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
This sounds reasonable. I need to think it through a bit more but I think we can get it in.

AStoker:
+1 for this feature

diegohb:
@Jenselme / @JeroenVinke - any idea if this should work for new SystemJS loader as well? Or does [this comment](https://github.com/aurelia/cli/issues/684#issuecomment-318811173) mean it may but not tested... ?

JeroenVinke:
I think you can use that, although I that config will be passed to amodrotrace as well, which uses requirejs internally. So if you use a systemjs specific config in there, it's likely not going to work. If this is the case then we need to think about how to solve it.

Issue #270, opened 2016-08-09
Mismatched anonymous define when using resources


JeroenVinke:
**I'm submitting a bug report**

* **Library Version:**
0.17.0

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
6.3.0

* **NPM Version:**
3.10.5

* **Browser:**
all

* **Language:**
all 

**Current behavior:**
When using the `resources` property of a bundle config in aurelia.json the following error is thrown: `Mismatched anonymous define`. [full error](http://i.imgur.com/wtjUaAI.png)
```
          {
            "name": "my-plugin-with-resources",
            "path": "../node_modules/my-plugin-with-resources/",
            "main": "index",
            "resources": [
              "**/*.js",
              "**/*.html"
            ]
          }
```

When I remove the resources property the error is gone, but the resources are not in the bundle

**Expected/desired behavior:**

1. au new
2. (use defaults)
3. npm install git+https://git@github.com/JeroenVinke/my-plugin-with-resources.git --save
4. add the following to the vendor-bundle config in aurelia.json:

  ```
  {
    "name": "my-plugin-with-resources",
    "path": "../node_modules/my-plugin-with-resources/",
    "main": "index",
    "resources": [
      "**/*.js",
      "**/*.html"
    ]
  }
  ```
5. update app.html
  ```
  <template>
    <require from="my-plugin-with-resources/counter"></require>

    <counter></counter>
    <h1>${message}</h1>
  </template>
  ```
6. au run --watch

* **What is the expected behavior?**

* **What is the motivation / use case for changing the behavior?**

pkozul:
+1

I also came across this issue when trying to get aurelia-kendoui-bridge working with the Aurelia CLI.

pkozul:
I think I've figured out what's going on.

I used the following project as the starting point:

	[](https://github.com/JeroenVinke/kendo-aureliacli)

The problem I was encountering is that the kendo components (grid, charts, etc.) were not being loaded. I believe they were not even being included in the bundle.

Below is what the bundle section looked like (in aurelia.json):

```
  {
	"name": "aurelia-kendoui-bridge",
	"path": "../node_modules/aurelia-kendoui-bridge/dist/amd",
	"main": "index"
  }
```

To force the Area CLI to include the grid and col components (for example), I added them using the "resources" node:

```
  {
	"name": "aurelia-kendoui-bridge",
[...]

PR #283, opened 2016-08-18
fix(source-inclusion): BundledSource must be registered in the bundler

JeroenVinke:
closes https://github.com/aurelia/cli/issues/270

BundledSource's were created for resources in the bundle config, but the BundledSource's were not registered in the bundler ([this](https://github.com/aurelia/cli/blob/master/lib/build/bundler.js#L96) did not happen). So when the tracer begins to trace all the files (including the resources) it would find [here](https://github.com/aurelia/cli/blob/master/lib/build/bundled-source.js#L82) that the file is not in the bundle and it would create another BundledSource.

Then after logging the files array [here](https://github.com/aurelia/cli/blob/master/lib/build/bundle.js#L115) I noticed that there were two files with the same path: `C:\Development\scaffolded\cli-issue-270\node_modules\my-plugin-with-resources\counter.js`. One of those had a `define` and the other one didn't, and both were persisted to the bundle.

### DIFF
--- a/lib/build/bundler.js
+++ b/lib/build/bundler.js
@@ -102,6 +102,8 @@ exports.Bundler = class {
     } else {
       subsume(this.bundles, found);
     }
+
+    return found;
   }
 
   updateFile(file, inclusion) {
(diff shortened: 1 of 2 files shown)


pkozul:
Tested this fix using the aurelia-kendoui-bridge.  Confirming that this fix does resolve the bundling issue.

EisenbergEffect:
I'll need to do careful review of this but we hope to get it in.

doktordirk:
fixed an issue for me

Issue #278, opened 2016-08-12
allow customizing sources folder location


atsu85:
**I'm submitting a bug report**
after renaming "src" folder to"xxx/yyy" `au run --watch` produces errors
- **Library Version:**
  0.17.0

**Please tell us about your environment:**
- **Operating System:**
  Windows 10
- **Node Version:**
  v6.3.0
- **NPM Version:**
  3.10.3
- **Browser:**
  all
- **Language:**
  TypeScript

**Current behavior:**

**Expected/desired behavior:**
should be able to relocate (not just rename) src folder: "src"-> "xxx/yyy"
- **What is the expected behavior?**
  after renaming "src" folder to "xxx" and changing aurelia.json, tsconfig.json, aurelia-karma.js and app.spec.ts applications and tests work as expected (`au run --watch` and `au run test`), **but after renaming the folder from "xxx" to "xxx/yyy" and updating previously mentioned conf files to reflect the cnage `au run --watch` produces errors**
- **What is the motivation / use case for changing the behavior?**
  I'd like to change the project folder layout

*\* Console output **
au run --watch
Starting 'readProjectConfiguration'...
Finished 'readProjectConfiguration'
Starting 'processMarkup'...
Starting 'processCSS'...
Starting 'configureEnvironment'...
Finished 'processCSS'
Finished 'processMarkup'
Finished 'configureEnvironment'
Starting 'buildTypeScript'...
Finished 'buildTypeScript'
Starting 'writeBundles'...
Tracing app...
Tracing environment...
Tracing main...
Tracing resources/index...
Tracing app...
Tracing text...
error c:\tmp\proj\playfield\aurelia-cli-test2\xxx\scripts\text.js
Tracing aurelia-bootstrapper...
error c:\tmp\proj\playfield\aurelia-cli-test2\xxx\node_modules\aurelia-bootstrapper\dist\amd\aurelia-bootstrapper.js
Tracing aurelia-binding...
error c:\tmp\proj\playfield\aurelia-cli-test2\xxx\node_modules\aurelia-binding\dist\amd\aurelia-binding.js
Tracing aurelia-dependency-injection...
error c:\tmp\proj\playfield\aurelia-cli-test2\xxx\node_modules\aurelia-dependency-injection\dist\amd\aurelia-dependency-injection.js
Tracing aurelia-framework...
error c:\tmp\proj\playfield\aurelia-cli-test2\xxx\node_modules\aurelia-framework\dist\amd\aurelia-framework.js
[...]

PR #295, opened 2016-08-27
fix(package-analyzing): allow deeper levels for source root

alpox:
Before, the package-analyzer looked for the node_modules folder in only exactly one path step back
to the source files.
I changed this behaviour to use a node-like lookup which looks for the node_modules folder in
the following manner:
./node_modules, ../node_modules, ../../node_modules, etc.

And this allows to put the root of the source (project.paths.root) somewhere deeper in a file-structure.
See Issue #278

### DIFF
--- a/lib/build/package-analyzer.js
+++ b/lib/build/package-analyzer.js
@@ -33,29 +33,31 @@ exports.PackageAnalyzer = class {
 }
 
 function loadPackageMetadata(project, description) {
-  setLocation(project, description);
-
-  return fs.readFile(description.metadataLocation).then(data => {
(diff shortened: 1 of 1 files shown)


alpox:
Please do not integrate yet, i found that it cannot find the bundle anymore with deeper source folder. It produces a weird lookup path:
http://localhost:9000/src/Web/src/Web/scripts/app-bundle.js 
It seems that the baseUrl passed to require.js and the bundle path got set together somehow and produce a wrong path.

EisenbergEffect:
I'm closing this for now. Feel free to submit a new PR in the future.

Issue #238, opened 2016-07-16
Provide the ability to exclude files from a bundle


jadrake75:
**I'm submitting a feature request**

* **Library Version:**
0.16.1

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
6.3.0

* **NPM Version:**
3.10.6

* **Browser:**
N/A

**Expected/desired behavior:**
Provide the ability to exclude source or resource patterns from bundling.  From poking through the source I do not see this ability at first glance, but documenting the various options esp. around the bundle config options would be a *huge* help.

* **What is the motivation / use case for changing the behavior?**
You may have source based elements that make sense to process with your other src elements, but they may be further processed or managed by other build artifacts.  The example I have is a splash screen less file which I want to build and process using the theme less files from src into a .css, but want to deploy it to a target resource folder in place of the scripts folder in a bundled resource.  Effectively anything in src/ that matches the .css/.html/.js extension will get picked up in the bundling.  

The other case is you want to partition a portion of the src into different bundles.  Without an exclude rule you have to implicitly declare the inclusions which gets cumbersome.

EisenbergEffect:
You can negate the blog patterns. is that sufficient?
All globbing is done with https://www.npmjs.com/package/minimatch

jadrake75:
Perhaps my globulation skills are amiss.....  in my case I am trying an example to exclude the theme css since it will be included in a bootstrap-theme bundle.  In order to make this work I needed to do the following...

1. Modify the cssProcessor to exclude the theme files since it pipes it to the bundler

```
    "cssProcessor": {
        "id": "less",
        "displayName": "Less",
        "fileExtension": ".less",
        "source": [
            "src/**/*.less",
            "!src/theme/*.less"
          ]
   }
```

I did not have to change the app-bundle.js module, but this may be because of the ! above I am effectively not generating css into the bundle pipeline for that directory, so I am assuming that I will have to also use it there?

```
        "bundles": [
            {
                "name": "app-bundle.js",
                "source": [
                    "[**/*.js]",
                    "[**/*.{css,html}]",
                    "![**/theme/*.css]"
                ]
            } ]

```

I was hoping I could just exclude them in the bundles config and avoid the processCSS source exclusion, because I still want them to be processed, just not into the app-bundle

EisenbergEffect:
Do you want the theme css to go into a different bundle other than app? or no bundle at all?

jadrake75:
[...]

Issue #203, opened 2016-06-29
Creating multiple bundle


dpinart:
Hi
I have an aurèlia application organized in modules, the folder structure looks like
src/
     module1/
         services/
         views/
     module2/
         services/
         views/
     app.html
     app.ts
     main.ts
     environment.ts
I'm trying to create 4 sepaate bundles, vendor-bundle, app-bundle, module1-bundle and module2-bundle. I configured bundle section in aurèlia.json as follows:

```
"bundles": [
      {
        "name": "app-bundle.js",
        "source": [
          "[*.js]",
          "*.{css,html}"
        ]
      },
      {
        "name": "module1-bundle.js",
        "source": [
          "[module1/**/*.js]",
          "module1/**/*.{css,html}"
        ]
      },
      {
        "name": "module2-bundle.js",
        "source": [
          "[module2/**/*.js]",
          "module2/**/*.{css,html}"
        ]
      },

      {
        "name": "vendor-bundle.js",
        "prepend": [
          "node_modules/bluebird/js/browser/bluebird.core.js",
          "scripts/require.js"
        ],
        "dependencies": [
          "aurelia-binding",
          "aurelia-bootstrapper",
          "aurelia-dependency-injection",
          "aurelia-event-aggregator",
          "aurelia-fetch-client",
          "aurelia-framework",
          "aurelia-history",
          "aurelia-history-browser",
          "aurelia-loader",
          "aurelia-loader-default",
          "aurelia-logging",
          "aurelia-logging-console",
          "aurelia-metadata",
          "aurelia-pal",
          "aurelia-pal-browser",
          "aurelia-path",
          "aurelia-polyfills",
          "aurelia-polymer",
          "aurelia-route-recognizer",
          "aurelia-router",
          "aurelia-task-queue",
          "aurelia-templating",
          "aurelia-html-import-template-loader",
          "aurelia-templating-binding",
          {
            "name": "text",
            "path": "../scripts/text"
          },
          {
            "name": "aurelia-templating-resources",
            "path": "../node_modules/aurelia-templating-resources/dist/amd",
            "main": "aurelia-templating-resources"
          },
          {
            "name": "aurelia-templating-router",
            "path": "../node_modules/aurelia-templating-router/dist/amd",
            "main": "aurelia-templating-router"
          },
          {
            "name": "aurelia-testing",
            "path": "../node_modules/aurelia-testing/dist/amd",
            "main": "aurelia-testing",
            "env": "dev"
          }
        ]
      }
[...]

PR #305, opened 2016-09-02
feat(bundle,source-inclusion): allow `exclude` option in bundle source

AStoker:
Fixes: https://github.com/aurelia/cli/issues/238
Address some of the issues mentioned here: https://github.com/aurelia/cli/issues/203  
Allow the `source` option in the bundle config to optionally be an
object with two properties, `include`, and `exclude`. This makes it
easier for patterns to be defined by the user, and allows for a more
consistent approach to creating bundles (similar to the aurelia-bundler)

### DIFF
--- a/lib/build/source-inclusion.js
+++ b/lib/build/source-inclusion.js
@@ -18,6 +18,7 @@ exports.SourceInclusion = class {
 
     this.pattern = pattern;
     this.matcher = this.bundle.createMatcher(pattern);
+	  this.excludes = this.bundle.excludes;
     this.items = [];
   }
 
@@ -27,8 +28,15 @@ exports.SourceInclusion = class {
     this.items.push(item);
   }
 
+  isExcluded(item){
+	  let found = this.excludes.findIndex(exclusion => {			
+			return exclusion.match(item.path);
+		});
+	  return found > -1;
+  }
+  
   trySubsume(item) {
-    if (this.matcher.match(item.path)) {
+    if (this.matcher.match(item.path) && !this.isExcluded(item)) {		
       item.includedBy = this;
       item.includedIn = this.bundle;
       this.items.push(item);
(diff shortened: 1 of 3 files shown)


lib/build/source-inclusion.js
```diff
@@ -27,8 +28,16 @@ exports.SourceInclusion = class {
     this.items.push(item);
   }
 
+  isExcluded(item){
+	  let found = this.excludes.findIndex(x => {			
+			let exclusion = new this.bundle.Minimatch(x);
```
EisenbergEffect: I'm mainly wondering if this could be more efficient by creating the Minimatch objects once, upfront, rather than on every call to isExcluded. @AStoker Any thoughts on that?

AStoker: I was thinking about that, but wasn't set on exactly where to place it. I gotta drive home from work early to avoid the flash floods here, so I'll find a more efficient place for it and have it in in an hour or so.

EisenbergEffect: Sounds good.

AStoker: Part of it might be my lack of knowledge of Minimatch, but I'm not entirely sure how the best way to go about creating the matcher up front. Where the `include` matcher can have a single pattern and loop through an array of files, the `exclude` matcher needs to have a file loop through an array of `exclusions`.  
Perhaps it's the Friday in me. I think I might be able to create the exclusion list on the bundle itself (instead of the source-inclusion), then each exclusion is just created once, and when the source-inclusion goes through trying to match files, it'll match against the bundle's exclusion list. How does that approach sound @EisenbergEffect?

EisenbergEffect: Yes, that's basically the approach I was thinking of. Give it a shot and if it works, amend this PR and I'll get it merged and released this weekend.

AStoker:
@EisenbergEffect, moved the exclusion list creation to here: https://github.com/AStoker/cli/blob/source-exclusion/lib/build/bundle.js#L22
Let me know what you think. 
Also, I added some details to how the bundles work in the `README.md`. There weren't any details before, so I added some basic details to both how bundles work normally, and with the new syntax.

EisenbergEffect:
Release coming in a few minutes.

PR #307, opened 2016-09-06
fix(resources): use paths.root in aurelia-karma.js

### LINKED ISSUE #246: aurelia-karma should use paths.root from aurelia.json

* **Library Version:**
0.16.2

* **Operating System:**
Linux (Fedora)

* **Node Version:**
6.3.1

* **NPM Version:**
3.10.3

* **Browser:**
all

* **Language:**
ESNext

**Current behavior:**
Currently, for `aurelia-karma` to work correctly, the source files must be in a `src`. This is hard coded [here](https://github.com/aurelia/cli/blob/master/lib/resources/test/aurelia-karma.js#L52). So if your sources are in a different directory (like `app`) you will get errors like this from karma:

```
24 07 2016 22:03:51.578:WARN [web-server]: 404: /base/app/app
Firefox 47.0.0 (Fedora 0.0.0) ERROR
  Error: Script error for "app/app", needed by: test/unit/app.spec.js
  http://requirejs.org/docs/errors.html#scripterror
  at dist/vendor-bundle.js:3661
```

- Configure in `aurelia.json` the paths for your app (eg sources in `app`)
- Run `au test`
- No errors, aurelia-karma should figure from `paths.root` in `aurelia.json` how to load the files correctly.

*Note:* If using `path.root` in `aurelia-karma` is too complicated, I think that a proper section in the documentation should be enough.

Jenselme:
To do that, we rely on client.args from karma's configuration to
exchange the values between karma and Aurelia. We can then get the
values in aurelia-karma.js with karma.args.

Close #246

### DIFF
--- a/lib/resources/content/karma.conf.js
+++ b/lib/resources/content/karma.conf.js
@@ -29,6 +29,12 @@ module.exports = function(config) {
     logLevel: config.LOG_INFO,
     autoWatch: true,
     browsers: ['Chrome'],
-    singleRun: false
+    singleRun: false,
+    // client.args must be a array of string.
+    // Leave 'aurelia-root', project.paths.root in this order so we can find
+    // the root of the aurelia project.
+    client: {
+      args: ['aurelia-root', project.paths.root]
+    }
   });
 };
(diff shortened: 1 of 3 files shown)


EisenbergEffect:
Thanks!

Issue #205, opened 2016-06-29
Version numbering


rapitkan:
It should be possible to add an arbitrary version number to the end of bundles (app-bundle.js and vendor-bundle.js) for preventing cache problems after version updates. At the moment there is a hard coded file name (vendor-bundle.js) in index.html.

kuechlerm:
Is there a strategy for cache busting?

rapitkan:
Timestamp could be a great start. Another options is to generate version number by content of the bundle. It would be different each time. Maybe the timestamp with milliseconds is an easier option...? Generated vendor-bundle-xxxxxx.js file could be injected into index.html so that user wouldn't have to worry about it.

TylerJPresley:
I've been playing with a final build step that is working, but is not the best solution. I've renamed a few things from stock, and this won't account for multiple bundles.

**aurelia_project/tasks/revision.js**
```
import gulp from 'gulp';
import replace from 'gulp-replace';
import rev from 'gulp-rev';
import intercept from 'gulp-intercept';
import del from 'del';
import moment from 'moment';
import project from '../aurelia.json';

let timestamp = moment().format('YYYYMMDDHHmmss');
let appBundle = null;

export function cacheBuster() {
  return gulp.src('index.html')
    .pipe(replace(/\?rev=\d{14}/g, `?rev=${timestamp}`))
    .pipe(gulp.dest('./'));
}

export function appVersion() {
  return gulp.src('version.json')
    .pipe(replace(/\d{14}/g, timestamp))
    .pipe(gulp.dest('./'));
}

export function revAppBundle() {
  del(['app/app-*.js']).then(paths => {
    console.log('Deleted files and folders:\n', paths.join('\n'));
  });
  return gulp.src('app/app.js')
    .pipe(rev())
    .pipe(intercept((file) => {
      appBundle = file.path.split('/').splice(-1)[0];
      return file;
    }))
    .pipe(gulp.dest('app/'));
}

export function injectAppBundle() {
  return gulp.src('app/aurelia.js')
    .pipe(replace(/\.\.\/app\/app-.*'/g, `../app/${appBundle.split('.')[0]}'`))
    .pipe(gulp.dest('./app'));
}

export default gulp.series(
  cacheBuster,
  appVersion,
  revAppBundle,
  injectAppBundle
);
```

**aurelia_project/tasks/build.js**
```
...
import revision from './revision';

export default gulp.series(
  readProjectConfiguration,
  gulp.parallel(
    transpile,
[...]

PR #326, opened 2016-09-19
feat(bundler,bundle): allow revision numbers

AStoker:
resolves: https://github.com/aurelia/cli/issues/205
Allow the user to optionally configure build revisions

### DIFF
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -14,6 +14,7 @@ exports.Bundle = class {
     this.dependencies = [];
     this.prepend = (config.prepend || []).filter(x => bundler.itemIncludedInBuild(x));
     this.moduleId = config.name.replace(path.extname(config.name), '');
+    this.hash = '';
     this.includes = (config.source || []);
     this.excludes = [];
     if(this.includes instanceof Array){
@@ -95,7 +96,6 @@ exports.Bundle = class {
       return Promise.resolve();
     }
 
-    console.log(`Writing ${this.config.name}...`);
 
     let work = Promise.resolve();
     let loaderOptions = this.bundler.loaderOptions;
@@ -108,7 +108,7 @@ exports.Bundle = class {
 
     if (loaderOptions.configTarget === this.config.name) {
       work = work.then(() => {
-        files.push({ contents: this.writeLoaderConfig(platform, loaderOptions) });
+        files.push({ contents: "_aureliaConfigureModuleLoader();"});
       });
     }
 
@@ -117,6 +117,8 @@ exports.Bundle = class {
 
       const Concat = require('./concat-with-sourcemaps');
       let concat = new Concat(true, this.config.name, os.EOL);
+      const generateHashedPath = require('./utils').generateHashedPath;
+      const generateHash = require('./utils').generateHash;
       let needsSourceMap = false;
 
       for (let i = 0; i < files.length; ++i) {
@@ -132,11 +134,31 @@ exports.Bundle = class {
 
       let mapContents;
       let contents = concat.content;
-      let mapFileName = this.config.name + '.map';
+      let bundleFileName = this.config.name;
+
+      if (buildOptions.rev) {
+        //Generate a unique hash based off of the bundle contents
+        this.hash = generateHash(concat.content);
+        bundleFileName = generateHashedPath(this.config.name, this.hash);       
+      }
+
+      if (loaderOptions.configTarget === this.config.name) {
+        //Add to the config bundle the loader config and change the "index.html" to reference the appropriate config bundle
+        concat.add(undefined, this.writeLoaderConfig(platform, loaderOptions));
+        contents = concat.content;
+        let outputDir = platform.baseUrl || platform.output; //If we have a baseUrl, then the files are served from there, else it's the output
+        if (platform.index) {
+          this.setIndexFileConfigTarget(platform, path.join(outputDir, bundleFileName));
+        }
+      }
+
+      let mapFileName = bundleFileName + '.map';
       let mapSourceRoot = path.relative(
         path.join(process.cwd(), platform.output),
         path.join(process.cwd(), this.bundler.project.paths.root)
       );
+
+      console.log(`Writing ${bundleFileName}...`);
 
       if (buildOptions.minify) {
         let minificationOptions = { fromString: true };
@@ -160,7 +182,7 @@ exports.Bundle = class {
         contents += os.EOL + '//# sourceMappingURL=' + mapFileName;
       }
 
-      return fs.writeFile(path.join(platform.output, this.config.name), contents).then(() => {
+      return fs.writeFile(path.join(platform.output, bundleFileName), contents).then(() => {
         this.requiresBuild = false;
 
         if (mapContents) {
@@ -184,17 +206,21 @@ exports.Bundle = class {
   }
 
   writeLoaderConfig(platform, loaderOptions) {
+    let loaderConfig = '';
     switch(loaderOptions.type) {
       case 'require':
-        return this.createRequireJSConfig(platform);
+        loaderConfig = this.createRequireJSConfig(platform);
+        break;
       default:
         throw new Error(`Loader configuration style ${loaderOptions.type} is not supported.`);
     }
+    return 'function _aureliaConfigureModuleLoader(){' + loaderConfig + '}'
   }
 
   createRequireJSConfig(platform) {
     let bundler = this.bundler;
     let loaderOptions = bundler.loaderOptions;
+    let buildOptions = this.buildOptions;
     let loaderConfig = bundler.loaderConfig;
     let bundles = bundler.bundles;
     let name = this.config.name;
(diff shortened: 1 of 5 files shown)


lib/build/bundle.js
```diff
@@ -108,7 +108,7 @@ exports.Bundle = class {
 
     if (loaderOptions.configTarget === this.config.name) {
       work = work.then(() => {
-        files.push({ contents: this.writeLoaderConfig(platform, loaderOptions) });
+        files.push({ contents: "_loadAureliaRequireJSConfig();"});
```
EisenbergEffect: Let's change the name of this function so that it doesn't mention RequireJS. In the future we'll support multiple module loaders. So maybe something like `_configureModuleLoader()`?

lib/build/bundle.js
```diff
@@ -220,15 +238,21 @@ exports.Bundle = class {
...
-      config.paths[currentBundle.moduleId] = location + '/' + currentBundle.moduleId;
+      config.paths[currentBundle.moduleId] = location + '/' + currentBundle.moduleId + (buildOptions.rev && currentBundle.hash ? '-' + currentBundle.hash : '');
     }
 
     if (includeBundles) {
       config.bundles = bundleMetadata;
     }
 
-    return 'requirejs.config(' + JSON.stringify(config) + ');'
+    return 'function _loadAureliaRequireJSConfig(){requirejs.config(' + JSON.stringify(config) + ');}'
```
EisenbergEffect: Don't forget to update the name here based on the change above.

EisenbergEffect: Also, the actual code `requirejs.config`... should be extracted to another method. As mentioned above, we'll ultimately have several different loaders we need to support. Each is going to generate their configuration a bit differently.

lib/build/bundle.js
```diff
@@ -220,15 +238,21 @@ exports.Bundle = class {
...
+  }
+  
+  setIndexFileConfigTarget(platform, location){
```
EisenbergEffect: Can add some code comments that explain a bit of what is going on there. I'm pretty sure I've got the idea but I might have trouble looking back on it in another month. So, perhaps some comments on this method and a few other places throughout the code above to explain the process.

lib/build/bundle.js
AStoker: I wanted it to mention something about Aurelia at least. Since this is being inserted in the vendor bundle file, it is possible (unlikely and they shouldn't do it anyway) that somebody define in the global namespace a similar function called `_configureModuleLoader`. So by naming it something like `_aureliaConfigureModuleLoader`, or something like that, we're better ensuring that people aren't going to override our function.

EisenbergEffect: That sounds good to me.

AStoker:
@EisenbergEffect, made the changes. Let me know what you think.

lib/build/bundle.js
AStoker: Done.

stsje:
Should be `crypto` and `path` dependencies be added to `package.json`?
Any reason why the same packages aurelia-bundler uses for rev aren't being used (rev-hash, rev-path)?
https://github.com/aurelia/bundler/blob/master/package.json
```
  "dependencies": {
    "bluebird": "^2.9.25",
    "clean-css": "^3.4.9",
    "glob": "^5.0.9",
    "globby": "^2.0.0",
    "html-minifier": "^1.1.1",
    "lodash": "^3.10.1",
    "rev-hash": "^1.0.0",
    "rev-path": "^1.0.0",
    "systemjs-builder": "^0.15.16",
    "whacko": "^0.19.1"
  }
```

AStoker:
`crypto` and `path` are built into Node, so I don't think it's required @stsje. As for `rev-hash/path`, it was easier for us to limit the external dependencies and pull similar functions into the `utils.js` file, that essentially do the exact same thing.

EisenbergEffect:
@AStoker I can't remember...did you add the license attribution to the acknowledgments doc for those libraries? We need to do that.

AStoker:
@EisenbergEffect, crap, forgot. The functions aren't exactly the same as what we grabbed from those modules. How should that be documented in the acknowledgements?

EisenbergEffect:
Got ahead and include it in the ack document the same as the other libraries listed there. You can see that doc for an example. Even though they aren't the same, lets attribute them fully since it was their code that fed into our version.

EisenbergEffect:
@AStoker Please double check the licenses. If they are anything other than MIT, please let me know so I can review. I want to make sure we are properly attributing them.

AStoker:
@EisenbergEffect, set and ready for review

AStoker:
Hold of temporarily on merging, found a case I'd like to consider when dealing with renaming the references to the bundles. I'll ping you once it's done @EisenbergEffect

EisenbergEffect:
Ok, thanks.

AStoker:
@EisenbergEffect, should be good for one last look now. Works on our production setup (but that's only one production setup 😉), and the skeleton projects. The one thing that I would like someone else to look over particularly is the regex on line 263 of the `bundle.js`. The goal is to isolate src tag that has the vendor bundle, and be lenient on how it's being called (like, allowing users to mix their slashes because the browser can handle it, lenient).  
Let me know how it flies.

lib/build/bundle.js
```diff
@@ -220,15 +244,25 @@ exports.Bundle = class {
...
-    return 'requirejs.config(' + JSON.stringify(config) + ');'
+    return 'requirejs.config(' + JSON.stringify(config) + ')'
+  }
+  
+  setIndexFileConfigTarget(platform, location){
+    //Replace the reference to the vendor bundle in the "index.html" file to use the correct build revision file (or remove the build revision hash);
+    let indexLocation = platform.index;
+    let data = fs.readFileSync(indexLocation);
+    let outputDir = platform.baseUrl || platform.output; //If we have a baseUrl, then the files are served from there, else it's the output
+    let configMatcher = new RegExp(`\\b(?:src=("|')(.*))(${outputDir}(\/|\\\\)${this.moduleId}(.*?).js)(?:("|'))`);
```
niieani: I think you'll need to escape `outputDir` and `moduleId` here, in case they contain slashes (subdirs, scoped modules) or other characters that need to be escaped. I recommend using https://github.com/sindresorhus/escape-string-regexp for this job.

It would be good to add some tests for this method specifically, as I'm not sure what's it supposed to do exactly without digging too deep into the package. Maybe you could refactor the Regex + match-replacement part of this method into a separate, pure function outside of the class. That way you could just create 2-3 simple tests: feeding it some sample text and comparing with desired results? 
Should be relatively, easy to do and will make this review much easier :)

AStoker: Tests are unfortunately lacking in the CLI right now. But what you recommended isn't a bad idea. Lemme see what I can do about cleaning this up a bit.

stsje: It would be nice with a warning if it was unable to update the index src path. Maybe the matcher doesn't find a vendor bundle path to replace.

EisenbergEffect: @AStoker I think this is a good idea.

AStoker: Good idea, I'll put this in there too

AStoker:
@niieani & @EisenbergEffect, updated based off the feedback. Mind giving it one more look over?

lib/build/bundler.js
```diff
@@ -166,7 +166,12 @@ exports.Bundler = class {
     }
 
     return doTransform()
-      .then(() => Promise.all(this.bundles.map(x => x.write(this.project.build.targets[0]))));
+      .then(() => {
+        let configTargetBundleIndex = this.bundles.findIndex(x => x.config.name == this.loaderOptions.configTarget);
```
EisenbergEffect: Just need a bit of clarification...is this just ensuring that the bundle that contains the config is processed in a certain order? I assume that is important for the hashing process. If so, can you just a comment here to explain that?

AStoker: That is correct

AStoker:
5th time is a charm :) @EisenbergEffect, ready if you are.

EisenbergEffect:
@AStoker Awesome! This has been fantastic. One final question...did you test this with a project not using the hash feature, to ensure that the default (and historic) configuration still works correctly?

AStoker:
@EisenbergEffect, it is now. We really should get tests on this library, so that way these regression issues get caught 😄  Thanks for saving me by double checking that!  
Historic projects will lack the `index` property and `rev` property, and since `rev` is falsy, it won't be looking for any of the new properties. If somebody using a historic setup, they would need to add the `index` property to the build targets, and set the `rev` to be true. Default project has the `index` property already set, so they'd just need to set `rev` to be true.

EisenbergEffect:
@AStoker Excellent. I would like to get tests going on this. A CLI like this is an interesting scenario. Any thoughts on how we might handle the testing?

AStoker:
I've never written tests for a CLI before @EisenbergEffect, just regular JS unit tests for our modules and a few Aurelia plugins. I would have to do some research to give you a good answer. But my initial thoughts for some basic tests would be to have a few difference "scenarios" represented by various `aurelia.json` files. From that, we might be able to at least do some basic test cases to make sure that functions will execute correctly given various scenarios, and are returning the expected data.

EisenbergEffect:
That makes sense. Ultimately you end up building a little mini-framework for testing, based on scenario definitions. The trick is the assertions I suppose. Lets just keep brainstorming about this. In the mean time, I'll get this merged in. Thanks for working on it!

AStoker:
Sounds about right. Do we have an issue/feature request to track test coverage?

EisenbergEffect:
I don't think we do yet.

AStoker:
I'll make one, and we can brainstorm there for ideas, and maybe (hopefully) somebody will have some time to hammer it out.  
I got some other features I need to take care of first for the office and our release before I think I can convince the boss to let me work on testing the CLI 😉

EisenbergEffect:
Understood :) This was a great contribution and I always love that its connected to real app needs.

AStoker:
Glad I could help! Aurelia really has been fantastic to work with, so I'm glad that my contributions can help others. There might be a few features that get spawned from this one (that need their own PR), such as getting tests to work with build revisions (should be solved if we get the ability to test the source files instead of bundles, which I'm also working on right now), and cleanup tasks might need to be expanded to make it easier for cleanup of revision numbers. All those are separate PR's though.   
Thanks again for the help in making this happen.

rapitkan:
This is great! Thank you for your effort.

Issue #328, opened 2016-09-21
Ability to run unit tests against unbundled source files instead of bundled files


chuckwatson87:
**I'm submitting a feature request**

* **Library Version:**
0.19.0

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
4.2.4

* **NPM Version:**
3.9.5

* **Browser:**
all 

* **Language:**
all

**Current behavior:**
Currently when running `au test` or `au test --watch`, the tests are being run against the bundled code.

**Expected/desired behavior:**

* **What is the expected behavior?**
to be able run tests against transpiled source code  instead of bundled code

* **What is the motivation / use case for changing the behavior?**
Currently, running the tests against the bundles is causing us to have to manually rebuild every time we make a change to the source files that we are testing.  Also makes debugging a bit harder, since we are only seeing the transpiled, bundled code in the browser when running tests.  Being able to test against the source code without bundling would improve this workflow.

AStoker:
+1 for this feature

Jenselme:
👍 I'd love to see that too.

chuckwatson87:
For anyone who is interested, worked on this a little bit more yesterday and today, and was able to get tests running against unbundled source code in the skeleton project using the following config:

```javascript
"use strict";
const path = require('path');
const project = require('./aurelia_project/aurelia.json');

let testSrc = [
  { pattern: 'node_modules/**/*.js', included: false },
  { pattern: 'node_modules/**/*.html', included: false },
  { pattern: 'node_modules/**/*.css', included: false },
  { pattern: 'node_modules/**/*.map', included: false },
  { pattern: 'node_modules/**/*.woff', included: false },
  { pattern: 'node_modules/**/*.png', included: false },
  { pattern: project.unitTestRunner.source, included: false },
  { pattern: project.transpiler.source, included: false },

  'test/aurelia-karma.js'
];

//let output = project.platform.output;
let entryPoint = "scripts/require.js";

//let appSrc = project.build.bundles.map(x => path.join(output, x.name));
// let entryIndex = appSrc.indexOf(path.join(output, project.build.loader.configTarget));
// let entryBundle = appSrc.splice(entryIndex, 1)[0];
let files = [entryPoint].concat(testSrc);

module.exports = function(config) {
  config.set({
    basePath: '',
    frameworks: [project.testFramework.id],
    files: files,
    exclude: ['node_modules/**/*spec.js', 'node_modules/**/*Spec.js'],
[...]

PR #353, opened 2016-10-04
feat(bundler, bundle): abstract out loader config generator

AStoker:
Resolves:
https://github.com/aurelia/cli/issues/328
Abstracted out the loader config generator from the `bundle.js`.
Exported on the `build` object the function `generateLoaderConfig`
which can be used to create a loader config without writing your
project to files.
BREAKING INTERNAL CHANGE: `bundler.write` now only writes each bundle,
while `bundler.build` will build each bundle and then write them.

### DIFF
--- a/lib/build/bundler.js
+++ b/lib/build/bundler.js
@@ -151,7 +151,7 @@ exports.Bundler = class {
     });
   }
 
-  write() {
+  build() {
     let index = -1;
     let items = this.bundles;
 
@@ -170,8 +170,11 @@ exports.Bundler = class {
         //Order the bundles so that the bundle containing the config is processed last.
         let configTargetBundleIndex = this.bundles.findIndex(x => x.config.name == this.loaderOptions.configTarget);
         this.bundles.splice(this.bundles.length, 0, this.bundles.splice(configTargetBundleIndex, 1)[0]);
-        return Promise.all(this.bundles.map(x => x.write(this.project.build.targets[0])));
       });
+  }
+  
+  write() {
+    return Promise.all(this.bundles.map(x => x.write(this.project.build.targets[0])));
   }
 }
 
--- a/lib/build/loader.js
+++ b/lib/build/loader.js
@@ -1,0 +1,101 @@
+"use strict"
+
+const path = require('path');
+const LoaderPlugin = require('./loader-plugin').LoaderPlugin;
+
+exports.createLoaderCode = function createLoaderCode(platform, bundler) {
+    let loaderCode;
+    let loaderOptions = bundler.loaderOptions;
+
+    switch (loaderOptions.type) {
+        case 'require':
+            loaderCode = 'requirejs.config(' + JSON.stringify(exports.createRequireJSConfig(platform, bundler)) + ')';
+            break;
+        case 'system':
+            loaderCode = exports.createSystemJSConfig(platform);
+            break;
+        default:
+            //TODO: Enhancement: Look at a designated folder for any custom configurations
+            throw new Error(`Loader configuration style ${loaderOptions.type} is not supported.`);
+    }
+
+    return loaderCode;
+}
+
+exports.createLoaderConfig = function createLoaderConfig(platform, bundler) {
+    let loaderConfig;
+    let loaderOptions = bundler.loaderOptions;
+
+    switch (loaderOptions.type) {
+        case 'require':
+            loaderConfig = exports.createRequireJSConfig(platform, bundler);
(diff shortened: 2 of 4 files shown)


lib/build/bundler.js
```diff
@@ -170,9 +170,17 @@ exports.Bundler = class {
         //Order the bundles so that the bundle containing the config is processed last.
         let configTargetBundleIndex = this.bundles.findIndex(x => x.config.name == this.loaderOptions.configTarget);
         this.bundles.splice(this.bundles.length, 0, this.bundles.splice(configTargetBundleIndex, 1)[0]);
-        return Promise.all(this.bundles.map(x => x.write(this.project.build.targets[0])));
+        if (postBuild && typeof postBuild == 'function') {
+          //If we want to take a different action other than writing to a file, this callback is returned
+          return postBuild();
+        }
+        return this.write();
```
EisenbergEffect: This is the one thing I'm not sure about. Would it be better to just remove the call to `write` all together and instead return a promise instead? Then you could simply chain on that to `write` or `generateConfig` depending on who is calling it. Even though it's a internal api (at the moment) I think that would make it cleaner. @AStoker What do you think? Any objections to making that change?

AStoker: @EisenbergEffect, so that function is already inside of the `doTransform` promise. I could just remove those lines like you mentioned, and then have the `index.js` modified like so:
``` javascript
exports.dest = function() {
  return bundler.build()
    .then(() => bundler.write());
}
```
I like that. I was hesitant to make a change like that initially in case people were using those functions (which I don't believe they should be). But since this is still alpha, we can get away with making that change I guess.

AStoker:
Side note, this is just the first step of pulling the config generator away from the bundler. Eventually, it would be nice for us to look to a folder location for any user made custom loader config generators. Just clarifying that feature isn't involved in this PR

lib/build/bundler.js
AStoker: Made the change @EisenbergEffect, let me know what you think.

EisenbergEffect:
@AStoker Looks solid. Since we don't have a test suite just yet...I would like to confirm that you tested this on a real world project as well as the CLI's default new project. Can you confirm?

AStoker:
@EisenbergEffect, I checked last night, and it worked fine, but I want to do a bit more testing before I give the double check star.  
Also, I talked with a co-worker about how we were planning on implementing this feature for tests, and have a small change. Initially I was revealing the `generateConfig` function which would return the complete config for whatever loader you're running (it would have the `requirejs.config(configObject)` output). This is helpful when we simply want a string to write into a file that when called will configure the loader. Where it's not helpful is when I want that configuration object itself to modify and use in tests.  
As a result, I'm going to rename `generateConfig` to `generateLoaderConfig` (it isn't used anywhere yet, so changing the name is fine). So we are defining "Loader Config" as the complete configuration needed for your loader. Additionally, I'll export the `createRequireJSConfig` function that now instead of returning the string for the entire loader config, it just returns the configuration object needed for RequireJS.  
I'll have the update with that coming in shortly.

EisenbergEffect:
That sounds good. I like that much better.

AStoker:
@EisenbergEffect, made the changes, let me know how you like them. Tested it in production code and on a new cli project.  
On a side note, we got to get cracking on those tests 😛 I'd feel a lot safer committing code if I knew that there were tests covering my back side.

lib/build/loader.js
```diff
@@ -0,0 +1,82 @@
+"use strict"
+
+const path = require('path');
+const LoaderPlugin = require('./loader-plugin').LoaderPlugin;
+
+exports.generateLoaderConfig = function generateLoaderConfig(platform, bundler) {
```
EisenbergEffect: The `generate` vs. `create` naming scheme seems too ambiguous to me. What if we call them `createLoaderCode` and `createLoaderConfig`?

AStoker: @EisenbergEffect, I was struggling with the naming. I like that more than what I had. But is it clear enough that people who call it know what they're going to get? I guess that could just be documented somewhere.

AStoker: To clarify, we'd have something like `createLoaderCode` to generate the "code" that configures the loader (i.e. the whole code including `requirejs.config(theConfigObject)`), and then have `createLoaderConfig` be what returns just the configuration object for the loader. Sound good?

AStoker:
@EisenbergEffect, ready for another look

Issue #368, opened 2016-10-08
Plugin with many components does not bundle correctly


JeroenVinke:
**I'm submitting a bug report**

* **Library Version:**
0.20.2

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
6.3.1

* **NPM Version:**
3.10.7

* **Browser:**
all

* **Language:**
all

**Current behavior:**
When using the `resources` property for a dependency that has many files, not all files are traced and added to the bundle. 

I have made an aurelia-cli project that demonstrates this: https://github.com/JeroenVinke/aurelia-cli-dependency-inclusion. There is a folder `large-plugin` that contains 20 box components (view/view-model pairs). This `large-plugin` is referenced as dependency from `aurelia.json`:

```
{
    "name": "large-plugin",
    "path": "../large-plugin/dist/amd",
    "main": "index",
    "resources": [
        "**/**.js",
        "**/**.html"
    ]
},
```

Running this app results in:

![image](https://cloud.githubusercontent.com/assets/2189477/19211664/1ff16da6-8d42-11e6-9308-a41cb6dedb8f.png)

Note that not all 20 boxes were traced.

In the browser the following error is shown:
![image](https://cloud.githubusercontent.com/assets/2189477/19211682/7a4f3f58-8d42-11e6-8c52-850a47bb6b77.png)

If I change the amount of boxes to two it works fine. This can be done by changing the resources property in aurelia.json to:
```
          {
            "name": "large-plugin",
            "path": "../large-plugin/dist/amd",
            "main": "index",
            "resources": [
              "**/box1.js",
              "**/box1.html",
              "**/box2.js",
              "**/box2.html"
            ]
          },
```

And by changing main.js to: `.plugin('large-plugin', (builder) => builder.boxes(2));`

Both files are traced:
![image](https://cloud.githubusercontent.com/assets/2189477/19211711/187e72f2-8d43-11e6-8c1e-74ee5c7deba8.png)

And the app works fine:
![image](https://cloud.githubusercontent.com/assets/2189477/19211710/0fce2332-8d43-11e6-97e4-57e95f37dc6c.png)

**Expected/desired behavior:**

Aurelia-cli should wait until all dependency files are processed

* **What is the motivation / use case for changing the behavior?**

So that the user can use the `resources` property for large plugins such as aurelia-kendoui-bridge and aurelia-materialize-bridge

JeroenVinke:
[...]

PR #369, opened 2016-10-08
fix(dependency-inclusion): wait until all resources are traced before bundling

JeroenVinke:
fixes https://github.com/aurelia/cli/issues/368

After these changes, all 20 box view/view-models are added to the bundle:
image

I believe there is another bug somewhere that prevents materialize from working with the cli, but i'll need to look into that. Will open an issue if I find something

### DIFF
--- a/lib/build/source-inclusion.js
+++ b/lib/build/source-inclusion.js
@@ -47,26 +47,32 @@ exports.SourceInclusion = class {
   }
 
   addAllMatchingResources(loaderConfig) {
-    const vfs = require('vinyl-fs');
+    return new Promise(resolve => {
+      const vfs = require('vinyl-fs');
 
-    let bundler = this.bundle.bundler;
-    let root = path.resolve(bundler.project.paths.root, loaderConfig.path);
-    let pattern = path.resolve(bundler.project.paths.root, this.pattern);
+      let bundler = this.bundle.bundler;
+      let root = path.resolve(bundler.project.paths.root, loaderConfig.path);
+      let pattern = path.resolve(bundler.project.paths.root, this.pattern);
 
-    var subsume = (file, cb) => {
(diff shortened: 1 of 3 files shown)


EisenbergEffect:
@JeroenVinke Looks pretty good. @AStoker Would you mind putting a second pair of eyes on this as well and letting us know if you see any issues?

lib/build/source-inclusion.js
```diff
@@ -47,26 +47,28 @@ exports.SourceInclusion = class {
...
+        let bundledSource = bundler.addFile(file, this);
+        bundledSource.moduleId = moduleId;
 
-      cb(null, file);
-    };
+        cb(null, file);
+      };
 
-    vfs.src(pattern).pipe(mapStream(subsume));
+      vfs.src(pattern).pipe(mapStream(subsume)).on('end', resolve);
```
AStoker: Should there be any error catching here in case the tracing fails? Just want to make sure that it not only _works_, but that when it _fails_ the user knows what happened (i.e. should there also be a `reject` catch in there?)

JeroenVinke: Good call, I have added an error handler:

img

Here `test` is the message of the error

AStoker: Just to double check (since we don't have tests yet on the CLI). Does this work on a new CLI project, as well as on a production setup?

Thanood: If it helps, I made the changes to a production setup by manually editing `node_modules/aurelia-cli`. That should be the same as updating the CLI, right?
In this case, it works. 😃

AStoker:
@EisenbergEffect, looks good to me.

EisenbergEffect:
Thanks to everyone on this PR. You have all been a tremendous help on the CLI!

Issue #322, opened 2016-09-16
Aurelia CLI template for ASP.NET Core doesn't serve files with au run


jcarrascal:
**I'm submitting a bug report**
- **Library Version:**
  au --version
  0.18.0

**Please tell us about your environment:**
- **Operating System:**
  Windows 10
- **Node Version:**
  node -v
  v6.3.1
- **NPM Version:**
  npm -v
  3.10.5
- **Browser:**
  all
- **Language:**
  all

**Current behavior:**
After creating a project with au new --here on an ASP.NET Core project, and runing with au run the server returns the string "Cannot GET /" to the server

**Expected/desired behavior:**
The server returns index.html. The browser displays "Hello, world!"
On ASP.NET Core projects the statics files are served from the wwwroot folder. I found that the template for ASP.NET Core projects create a gulp task called run that configures the browserSync server with is baseDir as ["."]. Changing it to baseDir: ['./wwwroot'] fixes the problem.

**Sugested**
On the ASP.NET Core templates change the baseDir property of browserSync to point to ['./wwwroot'] by default. Maybe also show a message with the name of the file to change (aurelia_project\tasks\run.ts) in case that the wwwroot folder has been renamed.

rsuter:
@jcarrascal how did you create a .NET Core based project with "au new"? I don't have this option...

kodeek:
I still needed to change baseDir in aurelia_project\tasks\run.ts in version 0.23.0.

bgardner87:
@kodeek I'm having the same issue and updated the baseDir in the run task. In my case, the page loads but gets a 404 when trying to load the vendor-bundle.js from the /scripts directory.

This looks like a similar relative path issue, are you seeing the same thing?

kodeek:
Hi @bgardner87, I think your issue is related to this issue - https://github.com/aurelia/cli/issues/173. I hope that will help you.

bgardner87:
@kodeek Thanks for the link, looks like you ended up in the same place I did by changing the build=>targets=>output property. From then on it seems to be working perfectly.

reft:
How did you manage to get the browser sync to work in .net core? I can only get it to work in port:9000, (not 5000).

http://stackoverflow.com/questions/42040786/aurelia-net-core-browser-sync-not-working-on-net-core

JeroenVinke:
Lets continue in https://github.com/aurelia/cli/issues/531 which is the same issue

PR #374, opened 2016-10-13
fix: change browserSync baseDir in run tasks to wwwroot for ASP.NET projects

### LINKED ISSUE #304: au ru with asp.net core/typescript project cannot be served by browsersync correctly
- **Library Version:**
  0.18.0

- **Operating System:**
  Windows 10
- **Node Version:**
  6.3.1
- **NPM Version:**
  3.10.6
- **Browser:**
  all 
- **Language:**
  TypeScript

**Current behavior:**
When running au run having configured the project for asp.net core/typescript, I get an error in multiple browsers (**Cannot Get**).  I think this is because the web server cannot serve the index.html file because it cannot find it.

I think the reason for this is that the base directory for the asp.net core project has been incorrectly configured.

the file **aurelia_project/run.ts** should have the line 26 changed from:

```
    baseDir: ['.'],
```

to

```
    baseDir: ['./wwwroot'],
```

because by default the configuration places the index.html for asp.net aurelia projects into the wwwroot folder.  When I manually made this change to my local project config, the Hello World page is able to be served.

I expect the Hello World page to be served by the web server provided by au
  This will enable the default project created by au for asp.net core to work correct with the command **au run**

arabsight:
baseDir is set to '.' for all projects, but ASP.NET Core projects serve files from wwwroot

fixes #304 and #322

### DIFF
--- a/lib/resources/tasks/run.js
+++ b/lib/resources/tasks/run.js
@@ -5,6 +5,8 @@ import project from '../aurelia.json';
 import build from './build';
 import {CLIOptions} from 'aurelia-cli';
 
+const baseDir = project.platform.id === 'aspnetcore' ? './wwwroot' : '.';
+
 function log(message) {
   console.log(message); //eslint-disable-line no-console
 }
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
Closing this since we can't handle it this way. We don't want to encode "aspnetcore" information into every project that is created. That's going to make Ruby, Node, Java, etc. developers unhappy. This needs to be updated in the asp.net code generation part of the cli.

PR #383, opened 2016-10-19
Fix #382 by catching errors with gulp-plumber

### LINKED ISSUE #382: au run --watch stops when less code doesn't compile
I'm submitting a bug report

**Library Version:**
0.21.0

- **Operating System:**
  Windows 10
- **Node Version:**
  6.6.0
  
  
- **NPM Version:**
  3.10.3
  
  
- **Language:**
  Less

**Current behavior:**
When creating a new project using `au new` and choosing less as the css processor, the task `au run --watch` stops when there is a less file that doesn't compile (invalid syntax for example).

- create a new project with `au new`, select less as the css processor
- run `au run --watch`
- create a file temp.less within src/resources, containing only "{"
- the watcher stops

- **What is the expected behavior?**
  `au run --watch` should not stop when a less file currently doesn't compile. It should just write an error to the console.
  This breaks the otherwise extremely nice workflow of having the editor set to autosave and having the watch process run in the background, especially because you only notice it when the browser stops updating.

jugglingthebits:
Stop breaking the stream when a less file doesn't compile by introducing plumber here as well. Also add better error logging through notify. Fixes #382

### DIFF
--- a/lib/resources/tasks/process-less.js
+++ b/lib/resources/tasks/process-less.js
@@ -2,12 +2,15 @@ import gulp from 'gulp';
 import changedInPlace from 'gulp-changed-in-place';
 import sourcemaps from 'gulp-sourcemaps';
 import less from 'gulp-less';
+import plumber from 'gulp-plumber';
+import notify from 'gulp-notify';
 import project from '../aurelia.json';
 import {build} from 'aurelia-cli';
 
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
Looks good. Thanks!

Issue #371, opened 2016-10-10
Include requirejs and requirejs/text via NPM.


shaunluttin:
Right now these are included in the `scripts` directory instead of letting NPM install them. 

In the aurelia.json, please move from this:

```
"prepend": [                                                   
  "scripts/require.js"                                         
],                                                             
"dependencies": [                                              
{                                                            
  "name": "text",                                            
  "path": "../scripts/text"                                  
},                                                           
```

to this: 

```
"prepend": [                                                   
  "node_modules/requirejs/require.js"                          
],                                                             
"dependencies": [                                              
  "text",                                                      
```

And add this to the package.json dependencies

```
"requirejs": "^2.3.2",
"text": "github:requirejs/text"
```

EisenbergEffect:
Sounds reasonable :) Is it not possible to get require.js from npm?

shaunluttin:
Yes. It is possible to get requirejs from NPM. It is also possible to get `requirejs/text` from there too. The problem is that the `aurelia-cli` scaffolding does not do that, so we have to manually make the above changes.

Gheoan:
Note that the `text` plugin is not in the npm registry, but can be obtained from github via npm.

AStoker:
Many of the people I've talked to have already replaced the included require and text loaders with the ones from npm. It seems simple enough to do and probably worth doing. Just have to modify the default aurelia.json file to load up those files, and include them in the default package.json

shaunluttin:
@AStoker That's what I do too. For one, it's a pain. For two, it's an inconsistent message to people learning Aureila; i.e. why are some libraries pulled from NPM but others installed into `scripts`?

EisenbergEffect:
I think we should fix this. It shouldn't be hard.

AStoker:
@EisenbergEffect, how does this look: https://github.com/aurelia/cli/pull/391

PR #391, opened 2016-10-24
refactor(project-template,dependencies): move scripts to node modules

AStoker:
Resolves: https://github.com/aurelia/cli/issues/371
Move RequireJS and the Text plugin into the node_modules so that
generated code is separated from library code.

### DIFF
--- a/lib/commands/new/project-template.js
+++ b/lib/commands/new/project-template.js
@@ -77,10 +77,6 @@ exports.ProjectTemplate = class {
       this.projectFolder,
       this.src,
       ProjectItem.directory('wwwroot').add(
-        this.scripts.add(
-          ProjectItem.resource('require.js', 'scripts/require.js'),
-          ProjectItem.resource('text.js', 'scripts/text.js')
-        ),
         ProjectItem.resource('index.html', 'content/index.html').askUserIfExists(),
         ProjectItem.resource('favicon.ico', 'img/favicon.ico').skipIfExists()
       ),
@@ -96,10 +92,6 @@ exports.ProjectTemplate = class {
     this.addToContent(
       this.projectFolder,
       this.src,
-      this.scripts.add(
-        ProjectItem.resource('require.js', 'scripts/require.js'),
-        ProjectItem.resource('text.js', 'scripts/text.js')
-      ),
       ProjectItem.jsonObject('package.json', this.package),
       ProjectItem.resource('.editorconfig', 'content/editorconfig'),
       ProjectItem.resource('.gitignore', 'content/gitignore'),
@@ -142,7 +134,9 @@ exports.ProjectTemplate = class {
       'aurelia-bootstrapper',
       'aurelia-fetch-client',
       'aurelia-animator-css',
-      'bluebird'
+      'bluebird',
+      'requirejs',
+      'text'
     ).addToDevDependencies(
       'aurelia-cli',
       'aurelia-testing',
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
Assigned to me so I can review later.

lib/dependencies.js
```diff
@@ -44,6 +44,8 @@ let versionMap = {
   "karma-babel-preprocessor": "^6.0.1",
   "karma-typescript-preprocessor": "^0.2.1",
   "minimatch": "^3.0.2",
+  "requirejs": "^2.3.2",
+  "requirejs-text": "^2.0.12",
```
MeirionHughes: this is wrong. version 2.0.12 doesn't work with aurelia, I've tried. Its also not maintained, nor official. `requirejs` repo has the plug-in inside it and its the only place its up to date.

MeirionHughes: https://github.com/MeirionHughes/aurelia-cli-electron-app/blob/master/package.json#L49

AStoker: Thanks, didn't know that! Let me make that fix real quick.

MeirionHughes: apparently they dont see a need to have an npm package for it. ¯\_(ツ)_/¯

AStoker: Let me know how the latest fix looks :)

MeirionHughes: Yes looks identical to how I've set mine up, which works well.

AStoker: The golden standard, of course ;)

lib/commands/new/project-template.js
```diff
@@ -328,7 +322,7 @@ exports.ProjectTemplate = class {
             "aurelia-templating-binding",
             {
```
Gheoan: Is the object needed? `"text"` worked fine for me.

AStoker: Thanks for the double check on that one. Wouldn't have hurt to have it that way, but there's no reason _to_ have it that way either :)

AStoker:
@EisenbergEffect, can we get this merged in?

lib/dependencies.js
```diff
@@ -44,6 +44,8 @@ let versionMap = {
   "karma-babel-preprocessor": "^6.0.1",
   "karma-typescript-preprocessor": "^0.2.1",
   "minimatch": "^3.0.2",
+  "requirejs": "^2.3.2",
+  "text": "github:requirejs/text",
```
Gheoan: > If you don't specify a commit-ish then master will be used.
> ―npm install docs

There is the [`latest`](https://github.com/requirejs/text/releases/tag/latest) tag which should be more stable then `master`.
That would be `"text": "github:requirejs/text#latest",`.

Sorry for not catching this in my first review.

AStoker: Better late than never @Gheoan! Thanks :)

AStoker:
Take 2 :) Unless @Gheoan finds something else, we should be good @EisenbergEffect.

EisenbergEffect:
I'll get it in for the release. Going to be working on that today.

AStoker:
Thanks. I had a few other PR's in the CLI that I'd like merged too, including the one that fixes the windows/posix separators

EisenbergEffect:
I'm planning to get those all in.

AStoker:
You're the man, thanks.

Issue #332, opened 2016-09-22
Never use windows-style path separators in aurelia.json


jdanyow:
On Windows, `au new` will create an aurelia.json with windows-style path separators (eg `"src\\**\\*.js"`). Windows aurelia.json files do not work on other platforms.

Using the `/` character for a separator works on all platforms.

mattbrooks2010:
Worth noting that `aurelia-cli` v0.20.0 also uses `\` on Windows when generating revision numbers / file hashes for generated bundles.

Example diff:

![image](https://cloud.githubusercontent.com/assets/2045274/18802878/95e14136-81e3-11e6-8315-56887fc9857b.png)

EisenbergEffect:
We need to put a general purpose solution in place to handle this
consistently. There is at least one other place that has this bug as well.

AStoker:
When dealing with revisions and changing the `index.html` file, it uses Node's Path module to create a path. This means that if it's compiled on windows, it'll use the backslash, and on Unix, it'll use the forward slash. I'll have a PR addressing the changing of the path in the index file soon. But we should probably change the default configuration that's generated by `au new` as well.

AStoker:
@EisenbergEffect, seems to enforce the non windows backslash we'll need to replace any place that is using `path.join` since that's platform specific and if compiled on windows will generate the backslash. Should we roll our own path join tool (what I'm would be the way to go), or is there some other approach you'd like to take?

EisenbergEffect:
That was what I was thinking.

AStoker:
If somebody wants to beat me to it, you could add this function to some general util and use it everywhere `path.join` is used.
``` javascript
function joinPath(){
	var joinedPaths = path.join.apply(null, arguments);
	if(process.platform == 'win32'){
		joinedPaths = joinedPaths.replace(/\\/g, '/');
	}
	return joinedPaths;
}
```
I gotta finish up getting the config generator to load up external config generators before I can work on this.

mattbrooks2010:
Another instance of this issue appears in `vendor-bundle.js`

```js
function _aureliaConfigureModuleLoader() {
    requirejs.config({
        baseUrl: "src/",
        paths: {
            text: "../wwwroot/scripts/text",
            "aurelia-animator-css": "../node_modules\\aurelia-animator-css\\dist\\amd\\aurelia-animator-css",
[...]

PR #394, opened 2016-10-24
fix(file-paths): don't use windows style path separators

AStoker:
Resolves: https://github.com/aurelia/cli/issues/332
Use universal (posix) path separators when creating user side code.


EisenbergEffect:
I think we have some Regex in other places that might be able to be replaced by this in the future as well. I didn't know about the `.posix` property. Nice work!

AStoker:
I noticed that there were additional locations, but I wasn't sure if we needed to do it everywhere, since the problem was only happening on the output code. Not sure how much `.posix` should be used internally if it's not needed :)

PR #397, opened 2016-10-26
Removed Typings dependenciy in favor of @types

con-ssc:


lib/dependencies.js
```diff
@@ -46,7 +46,10 @@ let versionMap = {
   "minimatch": "^3.0.2",
   "through2": "^2.0.1",
   "tslint": "^3.11.0",
-  "typings": "^1.3.0",
+  "@types/protractor": "^1.5.20",
+  "@types/jasmine": "^2.5.35",
+  "@types/selenium-webdriver": "^2.53.33",
```
Ullfis: Could you test this?

In my projects I have to lock this to version `"@types/selenium-webdriver": "2.44.26",`

EisenbergEffect:
@niieani Can you review this?

lib/dependencies.js
```diff
@@ -46,7 +46,10 @@ let versionMap = {
   "minimatch": "^3.0.2",
   "through2": "^2.0.1",
   "tslint": "^3.11.0",
-  "typings": "^1.3.0",
+  "@types/protractor": "^1.5.0",
+  "@types/jasmine": "^2.2.0",
+  "@types/selenium-webdriver": "^2.44.0",
```
niieani: Protractor 4 comes with its own typings and they contain jasmine and selenium. No need to even add these three nowadays. The rest is okay.

con-ssc: I can't see that protractor is installed with angular-cli yet. If you could point me at the file where the dependency to protractor is stated.

EisenbergEffect:
@niieani Do we need to do modifications here or is this good to go?

niieani:
@EisenbergEffect the question is whether we want to install typings for Protractor / Selenium / Jasmine by default or not. This is what CLI was doing with the old-style Typings. If the answer is no, then this should be fine to merge.

EisenbergEffect:
For Jasmine we want to install them if they selected the testing option in a TS setup. Otherwise, we don't.

niieani:
In that case OP will want to modify the PR so that adding a dependency on `@types/jasmine` is conditional. Otherwise remove `@types/protractor` and `@types/selenium-webdriver` dependencies and this is good.

milkshakeuk:
@con-ssc is this something you are still working on?

con-ssc:
@milkshakeuk Not really.

lib/commands/new/transpilers/typescript.js
```diff
@@ -24,11 +20,6 @@ module.exports = function(project) {
     'gulp-typescript',
     'gulp-tslint',
     'tslint',
-    'typescript',
```
milkshakeuk: shouldn't `@types/jasmine` and `@types/node` be added to the list of dev dependencies here? for them to be included in the `project.json` ready for when `installDependencies` is called in `project-template.js`?

niieani: devDependencies, yes

EisenbergEffect:
Merged #457

PR #410, opened 2016-11-14
fix(sourcemaps): stop adding non-existent sourcemaps to sourcemap

### LINKED ISSUE #409: HTML Source Maps not being created correctly

* **Library Version:**
0.22.0

* **Operating System:**
Windows 10

* **Node Version:**
6.9.1

* **NPM Version:**
3.10.8

* **Browser:**
all 

* **Language:**
all 

**Current behavior:**

Source maps for HTML files are not being added to `app-bundle.js.map` correctly. There are two issues. 

- HTML file paths in the `sources` array are being given as full absolute drive paths, e.g. `C:/foo/src/app.html`
- The `sourcesContent` array has `null` being placed in it for HTML files.

This issue affects both TypeScript and JavaScript applications.

AshleyGrant:
Fixes #409 

Only add sourcemaps to the sourcemap list if a sourcemap exists for the file.

Interestingly, this also makes the source mapped `src` folder show up the "correct" place.

image

Verified that sourcemaps now work in Edge:
image

### DIFF
--- a/lib/build/concat-with-sourcemaps/index.js
+++ b/lib/build/concat-with-sourcemaps/index.js
@@ -45,7 +45,7 @@ Concat.prototype.add = function(filePath, content, sourceMap) {
   }
   this.contentParts.push(content);
 
-  if (this.sourceMapping) {
+  if (sourceMap && this.sourceMapping) {
     var contentString = content.toString();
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
Thanks!

Issue #407, opened 2016-11-10
[Feature Request] Export static files


fabioluz:
I've been developing an aurelia-plugin for a legacy library. This library loads some static files asynchronously, which means that these files must be "exported" to the dist folder.  

After a long conversation with @AStoker, we came to the conclusion that aurelia-cli needs a feature to export files, like the one that SystemJS Skeleton has. It's good to emphasize that this feature is different from "resources" key in aurelia.json, that includes files in the bundle.  

This feature would be very helpful for handling libraries that weren't build for modular systems, and it will also solve some others problems like https://github.com/aurelia/cli/issues/248 

We were thinking in something like this: 
``` 
{ 
   “name”: “ckeditor”, 
   “path”: “../node_modules/ckeditor”, 
   “main”:”main”, 
    “resources”: [ 
    ], 
   “exports”: { 
      “editor.js”:”exportLocation/editor.js”, 
      “config.js”:”exportLocation/config.js” 
   } 
} 
``` 

However, we have to find anohter name because the key "exports" is being used for other things: 

``` 
{  
   "name": "bootstrap",  
    "path": "../node_modules/bootstrap/dist",  
    "main": "js/bootstrap.min",  
    "deps": ["jquery"],  
    "exports": "$",  
    "resources": [ "css/bootstrap.css" ]  
} 
``` 
 I can make the PR if I get the right direction.

What the rest of the team think about this?

AStoker:
Other name options (not sure of any of them, just throwing ideas):  
- Copy
- Export (no 's', but could easily cause confusion)  
- Move (could be confusion because you're not really _moving_ files)

Thanood:
How about `externals` because these files are "external" to the bundle..
Hmm, not perfect but I'll leave it here. 😃

AStoker:
I kind of like that. Could it be confused with thinking external to the project though? Like CDN's and things external to the build folder? Or is that overthinking it?

Thanood:
Maybe you're right.. Not sure if that's overthinking. I looked at my proposal for too long that I'm overthinking it myself right now. 😂 

- `static`
- `lazyLoad` (anything "lazy" actually)
- still like the `external`-ish direction somehow

fabioluz:
I think that `copy` and `copyToDist` are the best so far.

AStoker:
@fabioluz, for the sake of moving forward, I'd `copy`. If you're able to code it up in a way that makes renaming easy down the road, that'd be helpful.

Issue #248, opened 2016-07-29
Cannot add glyphicons to bootstrap resources


AshleyGrant:
**I'm submitting a bug report**
- **Library Version:**
  0.17.0

**Please tell us about your environment:**
- **Operating System:**
  Windows 10
- **Node Version:**
  6.3.0
  
  
- **NPM Version:**
  3.10.5
  
  
- **Browser:**
  all | Chrome XX | Firefox XX | Edge XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView
- **Language:**
  all

**Current behavior:**
Adding bootstrap's glyphicons file(s) causes the CLI to fail while bundling.

This configuration for bootstrap fails:

``` javascript
{
  "name": "bootstrap",
  "path": "../node_modules/bootstrap/dist",
  "main": "js/bootstrap.min",
  "deps": [
    "jquery"
  ],
  "exports": "$",
  "resources": [
    "css/bootstrap.css",
    "fonts/glyphicons-halflings-regular.woff2"
  ]
}
```

Here is the error messaged displayed:

``` sh
Tracing bootstrap/css/bootstrap...
Tracing bootstrap/fonts/glyphicons-halflings-regular...
{ uid: 9,
  name: 'writeBundles',
  branch: false,
  error:
   { Error: ENOENT: no such file or directory, open 'C:\Users\Ashley\temp\demo\node_modules\bootstrap\dist\fonts\glyphicons-halflings-regular.js'
       at Error (native)
       at Object.fs.openSync (fs.js:640:18)
       at Object.fs.readFileSync (fs.js:508:33)
       at Object.exports.readFileSync (C:\Users\Ashley\temp\demo\node_modules\aurelia-cli\lib\file-system.js:38:13)
       at amodroTrace.fileRead (C:\Users\Ashley\temp\demo\node_modules\aurelia-cli\lib\build\bundled-source.js:83:31)
       at Object.context.fileRead (C:\Users\Ashley\temp\demo\node_modules\aurelia-cli\lib\build\amodro-trace\lib\loader\Loader.js:176:18)
       at Object.context.load (C:\Users\Ashley\temp\demo\node_modules\aurelia-cli\lib\build\amodro-trace\lib\loader\Loader.js:357:30)
[...]

Issue #393, opened 2016-10-24
[Docs] Document CLI Recipes


AStoker:
## **I'm submitting a documentation request**

* **What is the motivation / use case for changing the behavior?** 

Many of the issues that arise for the CLI project are actually questions about implementation disguised as bugs. To help with this, a proposed "Recipes" section is being created.  
"Recipes" are quick examples of how to accomplish a specific task, with code examples included. For example, a recipe might contain how to configure the CLI to use Electron, or how to get the tests to run off of source code instead of the bundled code.  

Use this issue to discuss what kinds of recipes should be added, and which ones will be most helpful.  

## Recipes to add  

- [ ] Test Driven Development (how to configure)
- [ ] Generate external loader config file
- [ ] more..

Thanood:
A few first recipes:

### Enable BrowserSync to proxy a backend server

In `aurelia_project/tasks/run.js` change the `server` task to:
```javascript
let serve = gulp.series(
  build,
  done => {
    browserSync({
      // online: false,
      // notify: true,
      open: false,
      port: 9000,
      logLevel: 'silent',
      proxy: {
        target: 'localhost:24562'
      }
    }, function(err, bs) {
      let urls = bs.options.get('urls').toJS();
      console.log(`Application Available At: ${urls.local}`);
      console.log(`BrowserSync Available At: ${urls.ui}`);
      done();
    });
  }
);
```

Replace `localhost:24562` to the url you'd like to proxy.


###  Enable requirejs loading from 'scripts' directory
*IMO, this should work but doesn't, see #387*

Add this to `build.loader` in `aurelia.json`:
```json
"config": {
  "baseUrl": "scripts/"
},
``´

adriatic:
@AStoker allow me to register my big interest in participating in this discussion. @Thanood and I represent the same group (Aurelia-Tools / Aurelia-UI-Toolkits) and from all of our projects, it is **[Monterey](https://aurelia-ui-toolkits.gitbooks.io/monterey-overview/content/)** that is best suited to disseminate the information we hope to collect here.

Reference to Monterey (https://github.com/monterey-framework/monterey) allows me to mention a lot more ambitious tasks we plan to discuss with you (on recommendation from @EisenbergEffect who is well aware of Monterey's future plans and role in Aurelia community). I will stop by just this introduction as I do not want to disrupt the process already taking place in this article.

cc: @JeroenVinke

CasiOo:
- Changing output folder. The skeletons use `/dist`.
[...]

PR #415, opened 2016-11-17
feat(copy-files): copy specified files to the output directory

fabioluz:
Resolves: https://github.com/aurelia/cli/issues/407 and https://github.com/aurelia/cli/issues/248

### DIFF
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -97,7 +97,8 @@ exports.Bundle = class {
     if (!this.requiresBuild) {
       return Promise.resolve();
     }
-    
+
+
     let work = Promise.resolve();
     let loaderOptions = this.bundler.loaderOptions;
     let buildOptions = this.buildOptions;
@@ -190,26 +191,6 @@ exports.Bundle = class {
           return fs.writeFile(path.posix.join(platform.output, mapFileName), mapContents);
         }
       });
-    }).then(() => {
-      //copy files to the output directory
-      if (!this.config.dependencies) {
-        return Promise.resolve();
-      }
-
-      const root = this.bundler.project.paths.root;
-      let filesToCopyList = this.config.dependencies.filter(x => x.copy);
-
-      for (let filesToCopy of filesToCopyList) {
-        let modulePath = path.posix.join(root, filesToCopy.path);
-        for (let file in filesToCopy.copy) {
-          let source = path.posix.join(modulePath, file);
-          let target = path.posix.join(platform.output, filesToCopy.copy[file]);
-          console.log(`Copying ${source} to ${target}`);
-          fs.copyFile(source, target);
-        }
-      }
-
-      return Promise.resolve();      
     });
   }
 
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -5,7 +5,6 @@ const Convert = require('./convert-source-map');
 const fs = require('../file-system');
 const SourceInclusion = require('./source-inclusion').SourceInclusion;
 const DependencyInclusion = require('./dependency-inclusion').DependencyInclusion;
-const copyProperty = 'copy';
 
 exports.Bundle = class {
   constructor(bundler, config) {
@@ -98,8 +97,7 @@ exports.Bundle = class {
     if (!this.requiresBuild) {
       return Promise.resolve();
     }
-
-
+    
     let work = Promise.resolve();
     let loaderOptions = this.bundler.loaderOptions;
     let buildOptions = this.buildOptions;
@@ -192,30 +190,26 @@ exports.Bundle = class {
           return fs.writeFile(path.posix.join(platform.output, mapFileName), mapContents);
         }
       });
-    });
-  }
+    }).then(() => {
+      //copy files to the output directory
+      if (!this.config.dependencies) {
+        return Promise.resolve();
+      }
 
-  copyFiles() {
-    if (!this.config.dependencies) {
-      return Promise.resolve();
-    }
-
-    const output = this.bundler.project.platform.output;
-    const root = this.bundler.project.paths.root;
-    let filesToCopyList = this.config.dependencies
-      .filter(x => x instanceof Object && x[copyProperty]);
-
-    return Promise.resolve().then(() => {
+      const root = this.bundler.project.paths.root;
+      let filesToCopyList = this.config.dependencies.filter(x => x.copy);
 
       for (let filesToCopy of filesToCopyList) {
         let modulePath = path.posix.join(root, filesToCopy.path);
-        for (let file in filesToCopy[copyProperty]) {
+        for (let file in filesToCopy.copy) {
           let source = path.posix.join(modulePath, file);
-          let target = path.posix.join(output, filesToCopy[copyProperty][file]);
+          let target = path.posix.join(platform.output, filesToCopy.copy[file]);
           console.log(`Copying ${source} to ${target}`);
           fs.copyFile(source, target);
         }
       }
+
+      return Promise.resolve();      
     });
   }
 
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -5,6 +5,7 @@ const Convert = require('./convert-source-map');
 const fs = require('../file-system');
 const SourceInclusion = require('./source-inclusion').SourceInclusion;
 const DependencyInclusion = require('./dependency-inclusion').DependencyInclusion;
+const copyProperty = 'copy';
 
 exports.Bundle = class {
   constructor(bundler, config) {
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -201,16 +201,17 @@ exports.Bundle = class {
     }
 
     const output = this.bundler.project.platform.output;
+    const root = this.bundler.project.paths.root;
     let filesToCopyList = this.config.dependencies
       .filter(x => x instanceof Object && x[copyProperty]);
 
     return Promise.resolve().then(() => {
 
       for (let filesToCopy of filesToCopyList) {
-        let modulePath = path.join('aurelia_project', filesToCopy.path);
+        let modulePath = path.posix.join(root, filesToCopy.path);
         for (let file in filesToCopy[copyProperty]) {
-          let source = path.join(modulePath, file);
-          let target = path.join(output, filesToCopy[copyProperty][file]);
+          let source = path.posix.join(modulePath, file);
+          let target = path.posix.join(output, filesToCopy[copyProperty][file]);
           console.log(`Copying ${source} to ${target}`);
           fs.copyFile(source, target);
         }
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -232,7 +232,7 @@ exports.Bundle = class {
 
     return 'function _aureliaConfigureModuleLoader(){' + config + '}'
   }
-
+  
   setIndexFileConfigTarget(platform, location){
     //Replace the reference to the vendor bundle in the "index.html" file to use the correct build revision file (or remove the build revision hash);
     const escapeForRegex = require('./utils').escapeForRegex;
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -194,6 +194,29 @@ exports.Bundle = class {
     });
   }
 
+  copyFiles() {
+    if (!this.config.dependencies) {
+      return Promise.resolve();
+    }
+
+    const output = this.bundler.project.platform.output;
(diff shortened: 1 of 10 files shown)


AStoker:
I plan on reviewing this in the next day or so.

AStoker:
Sorry I haven't looked at this yet, I'm on a tighter schedule at work right now. In the mean time though, could you also submit a PR for Documentation of this feature, and be sure to link the two together (so that the documentation won't get merged until this is)?

fabioluz:
Sure!

alexisargyris:
Any news regarding this issue?

AStoker:
Going to review now :)

alexisargyris:
Thanks!

BTW, when requiring a css file, the text plugin seems to get used. But maybe I'm doing something wrong.

AStoker:
@alexisargyris, does that text plugin thing relate to this PR? Or is that a separate issue?

alexisargyris:
to this PR

In aurelia.json I add something like this:

    {
      "name": "some-lib",
      "path": "../node_modules/some-lib/css",
      "copy": {
         "some-lib.min.css": "styles/some-lib.min.css"
        }
    }

and from the template:

    <require from="../../../scripts/styles/some-lib.min.css"></require>

AStoker:
@alexisargyris, I don't believe that's a correct configuration. Your config doesn't include a `main` path for the javascript.  
But I'm not even sure what you're saying broken... Why is the text plugin a problem?

AStoker:
@fabioluz, one concern I have (not a deal breaker) is that the "to be copied" files are copied every time a file changes when using the `--watch` flag. This seems like it could cause poor performance if there are many or large files that need to be copied. Would it be possible to just copy them over when they're needed?  
In the case of the bootstrap example, that's a `vendor-bundle` file, so I'd imagine that at a minimum it would only be copied over when anything in the vendor bundle changes (ideally just if that file has been modified, but baby steps). However, what's happening is that whenever the writing takes place for any bundle (app-bundle in my case), it seems to trigger the copy.  

Ideas?
Also, I like what you've done here 😄

lib/build/bundle.js
```diff
@@ -5,6 +5,7 @@ const Convert = require('./convert-source-map');
 const fs = require('../file-system');
 const SourceInclusion = require('./source-inclusion').SourceInclusion;
 const DependencyInclusion = require('./dependency-inclusion').DependencyInclusion;
+const copyProperty = 'copy';
```
AStoker: Does this need to be a constant? Is there something wrong with just using something like `x.copy` instead of `x[copyProperty]`?

alexisargyris:
There are cases, e.g. font-awesome, where there is no javascript, only css.

Regarding the text plugin, requirejs fails to find the copied file, although it has been copied in the proper place and fails with a message like this

    plugin.load(map.name, localRequire, load, config)

While trying to see what was going on I noticed that the text plugin is being used and I assumed that this is a problem

lib/build/index.js
```diff
@@ -24,8 +24,8 @@ exports.createLoaderCode = function(p){
   return buildLoaderConfig(project)
       .then(() => {
         let platform = project.build.targets[0];
-        return createLoaderCode(platform, bundler);
```
AStoker: Just for the sake of viewing "changes", could you remove the changes for the white space here? Nit picky, I know 😛

lib/build/index.js
```diff
@@ -47,7 +47,12 @@ exports.bundle = function() {
 
 exports.dest = function() {
   return bundler.build()
-    .then(() => bundler.write());
+    .then(() => bundler.write())
+    .then(() => bundler.copyFiles())
```
AStoker: Maybe make the copyFiles function part of the write command for each bundle, that way you only copy the files when you're doing a write for that particular bundle...

AStoker:
@alexisargyris, I'm not sure that's necessarily related here, since I don't believe that's a feature currently supported by the CLI. That's something that we want to get, but I don't believe works now anyway.

lib/build/bundle.js
fabioluz: There's nothing wrong in my opinion. The only reason for that is because you said "code it up in a way that makes renaming easy down the road" (we weren't sure about `copy` or `copyToDist`).  So, I thought this would make renaming easier. I vote for using `x.copy` directly

AStoker: Ha ha, gotcha! Yah, I thought that might be what it was. After seeing what you've written, I think using `x.copy` will be just fine :) Thanks.

lib/build/index.js
fabioluz: My bad!! I didn't see it there!

lib/build/bundle.js
```diff
@@ -194,6 +195,30 @@ exports.Bundle = class {
...
 
+  copyFiles() {
+    if (!this.config.dependencies) {
+      return Promise.resolve();
+    }
+
+    const output = this.bundler.project.platform.output;
+    const root = this.bundler.project.paths.root;
+    let filesToCopyList = this.config.dependencies
+      .filter(x => x instanceof Object && x[copyProperty]);
```
fabioluz: @AStoker this is another thing that is bothering me is this. To be honest, I don't think this is common composition. Do you know if there is a problem using instanceof Object ? typeof x == 'object' seems to be more common.

Any ideas?

alexisargyris:
@AStoker I see, thanks for pointing it out

lib/build/bundle.js
AStoker: Well, I guess a better question is why can't you just say `.filter(x => x.copy)`? That will return truthy if `copy` exists, and falsy otherwise, which should take care of your filtering.

AStoker:
@alexisargyris, your problem is still valid though. I'm not sure if there's a feature request being tracked for that yet. If you could look to see if there is one, and if there isn't, could you create one?

lib/build/bundle.js
fabioluz: Indeed!

alexisargyris:
will do, thanks

alexisargyris:
I think that the 'copy files' recipe suggested by @martonsagi [here](https://github.com/aurelia/cli/issues/393) covers the case where an external lib needs to copy files to the output dir but has no js file of its own.

fabioluz:
@alexisargyris #248 is one of the reasons of why we are doing this. This pull request should solve that problem since that is related to a javascript library

fabioluz:
@AStoker I made some changes in the code, I strongly believe that the problem you mentioned won't happen anymore.  I'll try to update the docs tomorrow.

AStoker:
Awesome, thanks @fabioluz! I should get to this by Monday (weekend is full of house hunting). If somebody else is able to confirm that it only writes out when files in the associated bundle change, and that it works on a large scale/production setup, I should be able to approve the changes sooner :)

fabioluz:
PR of the documentation related to this feature: https://github.com/aurelia/framework/pull/666

AStoker:
@EisenbergEffect, what do you think of this feature?

EisenbergEffect:
Is there some reason this couldn't be done with gulp?

fabioluz:
IHO, the reason for not doing this with gulp is that the user would have to create the function himself, or the resource files would have to be updated.

AStoker:
This would be an additional tool built into the CLI that makes some cases much easier. Yes, the user could create a gulp task to do all this. And technically, the user can create a gulp task to do a whole lot. Libraries like Bootstrap, and many others, unfortunately pull in files in weird ways where they're assumed to be at a specific location. By tying in this "copy" functionality _into_ where you're configuring your bundle, you're allowing the users to maintain a much cleaner setup, and not have to worry about maintaining a gulp file for 5 different libraries.  

This might also (eventually) help solve one of the problems of libraries that are css/html only. A workaround would be to not require a "main" file, and then just copy your needed files into your output directory. We would need a slight tweak to make sure that "main" is optional, but with something like that in there, you'd be enabling users to more easily use many more libraries without lots of gulp knowledge.

AStoker:
@EisenbergEffect, how do you feel about the explanation I gave?

EisenbergEffect:
I'm not sure. I really wanted to not rebuild the world into the CLI. Is there some reason that a gulp task couldn't get access to the same information? I haven't looked at how the PR works. But, a good way might be to just have the CLI generate a gulp task for new projects that takes the data from a new section in the aurelia.json file and does something with it. I'm hesitant to build it into the core cli. Does that make sense?

fabioluz:
This is an example of how we could do this with gulp:


**aurelia.json**
```
 "bundles": [ ... ], 
 "copyFiles": {
    "../node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.woff2": "../bootstrap/fonts"
    "../node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.woff": "../bootstrap/fonts",
    "../node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.ttf": "../bootstrap/fonts"
  }
```

**tasks/run.js**
```
function copyFiles(done) {
  let filesToCopy = project.build.copyFiles;
  if (!filesToCopy) {
    done();
    return;
  }
 
  for (let file in filesToCopy) {
    let source = path.posix.join(project.paths.root, file);
    let target = path.posix.join(project.platform.output, filesToCopy[file]);
    log(`Copying ${source} to ${target}`);
    gulp.src(source).pipe(gulp.dest(target));
  }
 
  done();
}

let serve = gulp.series(
  build,
  copyFiles,  
  done => {
     ...
  }
);
```
I think the `serve` object is a good place to attach the `copyFiles` function because the other functions run whenever the user changes something in the code (considering that --watch was provided).

@EisenbergEffect is that what you had in mind?

Now the question is, is it worth to be pushed into the CLI? I think it is because this is something that almost everybody will use someday. Also, this is good subject to be in the docs.

EisenbergEffect:
Yes, this is what I had in mind. I think it's worth pushing this into the CLI as a task that is generated into the tasks folder and connected into the serve task as you mention. I would make copy-files its own task file.

fabioluz:
OK! I'll edit the PR with these changes.

fabioluz:
I've updated the code. We are using gulp strategy now and the `copyFiles()` has its own file.

The set up like this: 
```
"copyFiles": {
  "FILE_YOU_WANT_TO_COPY_BASED_ON_ROOT": "DESTINATION_FOLDER_BASED_ON_OUTPUT"
}
```
By default `root` is the *src* folder and the `output` is the *scripts* folder.

An example for bootstrap fonts would be this:
```
 "bundles": [ ... ], 
 "copyFiles": {
    "../node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.woff2": "../bootstrap/fonts"
    "../node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.woff": "../bootstrap/fonts",
    "../node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.ttf": "../bootstrap/fonts"
  }
```

@AStoker could you review the code and give your opinion about this?

fabioluz:
Any update on this?

thomas-darling:
@AStoker @EisenbergEffect I have a couple of concerns about this.

> By default `root`is the _src_ folder and the `output` is the _scripts_ folder.

I don't understand why the `output` is be relative to the  _scripts_ folder.
I'd very strongly prefer to have it be relative to the project output folder instead, so I don't have to write _../_ when copying files that are not scripts - I think copying non-script files will be a _much_ more common scenario for this, such as in the example above, which actually copies font files.

I'm also wondering why this task runs as part of the `serve` task - that's for development only, so this won't work on a build server. From my point of view, copying additional files should just be part of the build, and those files should be watched for changes, just like any other source file.

Thoughts?

EisenbergEffect:
I agree with all points. Would you be willing to submit a PR to make those
changes?

fabioluz:
@thomas-darling I thought that the `output` folder declared in the **aurelia.json** would be most reasonable folder to be the _output_ of copied files. But I agree with you when you say that this feature will be mostly used to copy non-script files. So, I think that changing the `output` folder to the `root` folder is a nice thing to do.

Regarding running the `copy` task as part of the `serve` task, that was my mistake, sorry! I totally forgot that the files won't be copied when building for production. However, we cannot run the copy task as part of the `build` task because this could cause performance issues.

The watchers run the `build` task whenever something changes in the source-code. Imagining a high-scale scenario, where several files should be copied, these files would be copied in every change, consequently delaying the browser-sync to refresh the page. 

To solve this, we have to edit the `build` task, ensuring that the files will only be copied if necessary. 

Thoughts? @thomas-darling @EisenbergEffect @AStoker

PR #417, opened 2016-11-19
fix(transform): enable wrap shim for shimmed definitions

DrSammyD:
This allows aurelia.json specify in the loader config that it wants to wrap a shimmed library in a define. Allows this section of code to be useful https://github.com/aurelia/cli/blob/master/lib/build/amodro-trace/write/defines.js#L41-L55


EisenbergEffect:
Thanks!

DrSammyD:
Sorry, yeah it looks like it's a typo.

Issue #405, opened 2016-11-09
Build emits outputs to wrong directory on ASP.NET Core


JamesFenton:
**I'm submitting a bug report**

* **Library Version:**
0.22.0

**Please tell us about your environment:**
* **Operating System:**
Windows 8.1

* **Node Version:**
6.5.0

* **NPM Version:**
3.10.9

* **Browser:**
all

* **Language:**
all

**Current behavior:**
Running `au build` produces the bundles in the ./scripts folder instead of ./wwwroot/scripts in ASP.NET Core projects

**Expected/desired behavior:**

- au new --here
- Choose 2 (ASP.NET Core)
- Choose 1 or 2 (Babel or TypeScript)
- Choose 1 (No CSS pre-processor)
- Choose 1 (Yes for unit testing)
- Choose 1 (Create project)
- Choose 1 (Install dependencies)
- run `au build`
- a 'scripts' folder is created with the bundles in the current folder, not in the wwwroot folder

* **What is the expected behavior?**
The build output should be in ./wwwroot/scripts
This can be fixed after creating the project by changing the `build.targets.output` in aurelia.json from 'scripts' to 'wwwroot/scripts'

* **What is the motivation / use case for changing the behavior?**
index.html cannot load the bundles (404) and the app does not work

AStoker:
Reproducible on my machine too.

ghidello:
In order to use the **rev** option I also had to update the `build.targets.index` property to `wwwroot/index.html`

JamesFenton:
@ghidello I have the same issue. I believe that is related to #357

ghidello:
Yes, looks like it's the same thing, thanks for pointing out. I started reading the CLI sources and putting together a solution but now I see it has been closed because they're already working on it. Well, I needed to start somewhere so it was worth it anyway.

EisenbergEffect:
Feel free to submit a PR :)

malk2013:
hi,
i have started new aurelia project but still this issue is hanging around look at my aurelia.json
"build": {
    "targets": [
      {
        "id": "aspnetcore",
        "displayName": "ASP.NET Core",
        "output": "wwwroot/scripts",
        "index": "wwwroot/index",
        "baseUrl": "scripts"
      }
    ],
when i run ---->    au run     its not showing my default output . please help

alvaradojl:
same problem here, the fix is still necessary

PWKad:
The fix here for asp.net using --here flag -

[...]

PR #418, opened 2016-11-20
fix(project-template): use wwwroot as build target for aspnet core

ghidello:
The "new" command configuration has been changed in order to prepend the folder "wwwroot" to both
the output and index properties used during the bundling of an aspnet core project

Closes issue #405


AStoker:
Can you confirm that these changes don't affect any of the other configuration setups? We don't have tests on the CLI yet, so right now we have to manually test all the features to make sure we don't have regression issues.

EisenbergEffect:
This looks pretty good. I'll need to spend some more time testing it before we merge.

PR #419, opened 2016-11-23
fix #303 Time required to build and/or run a project is seemingly excessive

### LINKED ISSUE #303: Time required to build and/or run a project is seemingly excessive
**I'm submitting a performance request**
- **Library Version:**
  0.17.0

- **Operating System:**
  OSX 10.11
- **Node Version:**
  6.4.0
  
  
- **NPM Version:**
  3.10.6
  
  
- **Browser:**
  Chrome 52.0.2743.116
- **Language:**
  TypeScript

**Current behavior:**
When executing au run, or au build it takes considerably longer to build my project than it did with gulp. The longest part of the build process is when it traces my src directory. Is there a configuration param that will speed things up a bit?

- **What is the expected behavior?**
  For au run to start up faster than it currently does. Preferably as fast as gulp did.
  Speeding up workflow. The bigger my project gets, the longer it takes. As of right now I have 15 small files in my src directory and it takes 30+ seconds before I can run my app.

huochunpeng:
This fix patches amodro to use cached deps and contents to avoid duplicated work. It's able to cut of ~70% build time on my project.

Works in watch mode too. (it seems cli doesn't use amodro to process additional file change)
Works in `au build --env prod` too at least on my project.

This should close #303.

### DIFF
--- a/lib/build/amodro-trace/index.js
+++ b/lib/build/amodro-trace/index.js
@@ -89,6 +89,7 @@ module.exports = function trace(options, loaderConfig) {
 
     function onOk() {
       var context = loader.getContext();
+      var traceCache = context._traceCache || {};
 
       // Pull out the list of IDs for this layer as the return value.
       var paths = context._layer.buildFilePaths,
@@ -144,10 +145,9 @@ module.exports = function trace(options, loaderConfig) {
         }
 
         if (options.includeContents && filePath) {
-          const cached = options.traced && options.traced.find(item => item.id === id);
-
-          if (cached && cached.contents) {
-            item.contents = cached.contents;
+          var cacheEntry = traceCache[id];
+          if (cacheEntry && cacheEntry.contents) {
+            item.contents = cacheEntry.contents;
           } else {
             var contents = context._readTransformedContents[filePath];
             if (!contents && context.fileExists(id, filePath)) {
--- a/lib/build/amodro-trace/index.js
+++ b/lib/build/amodro-trace/index.js
@@ -146,7 +146,7 @@ module.exports = function trace(options, loaderConfig) {
         if (options.includeContents && filePath) {
           const cached = options.traced && options.traced.find(item => item.id === id);
 
-          if (cached) {
+          if (cached && cached.contents) {
             item.contents = cached.contents;
           } else {
             var contents = context._readTransformedContents[filePath];
--- a/lib/build/amodro-trace/index.js
+++ b/lib/build/amodro-trace/index.js
@@ -144,22 +144,28 @@ module.exports = function trace(options, loaderConfig) {
         }
 
         if (options.includeContents && filePath) {
-          var contents = context._readTransformedContents[filePath];
-          if (!contents && context.fileExists(id, filePath)) {
-            contents = context.fileRead(id, filePath) || '';
-            if (options.writeTransform) {
-              contents = options
-                         .writeTransform(context, id, filePath, contents);
-            }
-          }
-          if (contents && options.writeTransform) {
-            contents = options.writeTransform(context,
-                                              id,
-                                              filePath,
-                                              contents);
-          }
+          const cached = options.traced && options.traced.find(item => item.id === id);
 
-          item.contents = contents;
+          if (cached) {
+            item.contents = cached.contents;
+          } else {
+            var contents = context._readTransformedContents[filePath];
+            if (!contents && context.fileExists(id, filePath)) {
+              contents = context.fileRead(id, filePath) || '';
+              if (options.writeTransform) {
+                contents = options
+                           .writeTransform(context, id, filePath, contents);
+              }
+            }
+            if (contents && options.writeTransform) {
+              contents = options.writeTransform(context,
+                                                id,
+                                                filePath,
+                                                contents);
+            }
+
+            item.contents = contents;
+          }
         }
 
         if (filePath) {
--- a/lib/build/amodro-trace/lib/loader/Loader.js
+++ b/lib/build/amodro-trace/lib/loader/Loader.js
@@ -139,6 +139,9 @@ function __exec(contents, r, d, c) {
     // Stores the _options.readTransform contents for use in top level results.
     context._readTransformedContents = {};
 
+    // Stores previous trace results for use in this trace.
+    context._traceCache = {};
+
     // For tracing, do everything sync so that build runs are more reproducible.
     // Otherwise, async IO could lead to differing results.
     context.nextTick = function (fn) {
@@ -293,7 +296,7 @@ function __exec(contents, r, d, c) {
     context.load = function (moduleName, url) {
       /*jslint evil: true */
       var contents, pluginBuilderMatch, builderName,
-        shim, shimExports,
+        shim, shimExports, traceCacheEntry,
         currentLoadingUrl = context.currentLoadingUrl;
 
       // If this module is no longer in the registry, because it was undefined,
@@ -352,66 +355,77 @@ function __exec(contents, r, d, c) {
             layer.existingRequireUrl = url;
           }
         } else {
-          // Load the file contents, process for conditionals, then
-          // evaluate it.
-          contents = context.fileRead(moduleName, url);
-
-          if (context.config._options.readTransform) {
-            contents = context.config._options
-                       .readTransform(moduleName, url, contents);
-          }
-
-          if (context.config._options.includeContents) {
-            context._readTransformedContents[url] = contents;
-          }
-
-          // Find out if the file contains a require() definition. Need to
-          // know this so we can inject plugins right after it, but before
-          // they are needed, and to make sure this file is first, so that
-          // define calls work.
-          try {
-            if (!layer.existingRequireUrl &&
-                parse.definesRequire(url, contents)) {
-              layer.existingRequireUrl = url;
-              context._cachedDefinesRequireUrls[url] = true;
+          traceCacheEntry = getOwn(context._traceCache, moduleName);
+          if (traceCacheEntry) {
+            contents = 'define("' + moduleName + '",';
+            if (traceCacheEntry.deps) {
+              contents += '["' + traceCacheEntry.deps.join('","') + '"]';
+            } else {
+              contents += '[]';
             }
-          } catch (e1) {
-            throw new Error('Parse error using esprima ' +
-                    'for file: ' + url + '\n' + e1);
-          }
+            contents += ');';
+          } else {
+            // Load the file contents, process for conditionals, then
+            // evaluate it.
+            contents = context.fileRead(moduleName, url);
 
-          if (hasProp(context.plugins, moduleName)) {
-            // This is a loader plugin, check to see if it has a build
-            // extension, otherwise the plugin will act as the plugin
-            // builder too.
-            pluginBuilderMatch = pluginBuilderRegExp.exec(contents);
-            if (pluginBuilderMatch) {
-              // Load the plugin builder for the plugin contents.
-              builderName = context.makeModuleMap(pluginBuilderMatch[3],
-                                context.makeModuleMap(moduleName),
-                                null,
-                                true).id;
-              contents = context.fileRead(builderName,
-                                          context.nameToUrl(builderName));
+            if (context.config._options.readTransform) {
+              contents = context.config._options
+                         .readTransform(moduleName, url, contents);
             }
-          }
 
-          // Parse out the require and define calls.
-          // Do this even for plugins in case they have their own
-          // dependencies that may be separate to how the pluginBuilder
-          // works.
-          try {
-            if (falseProp(context.needFullExec, moduleName)) {
-              contents = parse(moduleName, url, contents, {
-                insertNeedsDefine: true,
-                has: context.config.has,
-                findNestedDependencies:
-                              context.config._options.findNestedDependencies
-              });
+            if (context.config._options.includeContents) {
+              context._readTransformedContents[url] = contents;
             }
-          } catch (e2) {
-            throw new Error('Parse error using esprima ' +
-                    'for file: ' + url + '\n' + e2);
+
+            // Find out if the file contains a require() definition. Need to
+            // know this so we can inject plugins right after it, but before
+            // they are needed, and to make sure this file is first, so that
+            // define calls work.
+            try {
+              if (!layer.existingRequireUrl &&
+                  parse.definesRequire(url, contents)) {
+                layer.existingRequireUrl = url;
+                context._cachedDefinesRequireUrls[url] = true;
+              }
+            } catch (e1) {
+              throw new Error('Parse error using esprima ' +
+                      'for file: ' + url + '\n' + e1);
+            }
+
+            if (hasProp(context.plugins, moduleName)) {
+              // This is a loader plugin, check to see if it has a build
+              // extension, otherwise the plugin will act as the plugin
+              // builder too.
+              pluginBuilderMatch = pluginBuilderRegExp.exec(contents);
+              if (pluginBuilderMatch) {
+                // Load the plugin builder for the plugin contents.
+                builderName = context.makeModuleMap(pluginBuilderMatch[3],
+                                  context.makeModuleMap(moduleName),
+                                  null,
+                                  true).id;
+                contents = context.fileRead(builderName,
+                                            context.nameToUrl(builderName));
+              }
+            }
+
+            // Parse out the require and define calls.
+            // Do this even for plugins in case they have their own
+            // dependencies that may be separate to how the pluginBuilder
+            // works.
+            try {
+              if (falseProp(context.needFullExec, moduleName)) {
+                contents = parse(moduleName, url, contents, {
+                  insertNeedsDefine: true,
+                  has: context.config.has,
+                  findNestedDependencies:
+                                context.config._options.findNestedDependencies
+                });
+              }
+            } catch (e2) {
+              throw new Error('Parse error using esprima ' +
+                      'for file: ' + url + '\n' + e2);
+            }
           }
 
           context._cachedFileContents[url] = contents;
@@ -655,8 +669,14 @@ function __exec(contents, r, d, c) {
       inlineText: true
     });
 
+    var context = this.getContext();
+    context._setToolOptions(options || {});
 
-    this.getContext()._setToolOptions(options || {});
+    if (options.traced) {
+      options.traced.forEach(function(traceItem) {
+        context._traceCache[traceItem.id] = traceItem;
+      });
+    }
   }
 
   Loader.prototype = {
--- a/lib/build/bundled-source.js
+++ b/lib/build/bundled-source.js
@@ -5,6 +5,19 @@ const cjsTransform = require('./amodro-trace/read/cjs');
 const allWriteTransforms = require('./amodro-trace/write/all');
 const fs = require('../file-system');
 
+let _cache = {};
+
+function cacheTraced(traced) {
+  traced.forEach(item => {
+    const {id, deps, contents} = item;
+    _cache[item.id] = {id, deps, contents};
+  });
+}
+
+function cachedTraced() {
+  return Object.values(_cache);
+}
+
 exports.BundledSource = class {
   constructor(bundler, file) {
     this.bundler = bundler;
@@ -70,6 +83,7 @@ exports.BundledSource = class {
       {
         rootDir: rootDir,
         id: moduleId,
+        traced: cachedTraced(),
         fileExists: function(defaultExists, id, filePath) {
           return !!bundler.getItemByPath(calculateFileName(filePath));
         },
@@ -101,6 +115,7 @@ exports.BundledSource = class {
       loaderConfig
     ).then(traceResult => {
       let traced = traceResult.traced;
+      cacheTraced(traced);
 
       for (let i = 0, ii = traced.length; i < ii; ++i) {
         let result = traceResult.traced[i];
--- a/lib/build/bundled-source.js
+++ b/lib/build/bundled-source.js
@@ -9,8 +9,7 @@ let _cache = {};
 
 function cacheTraced(traced) {
   traced.forEach(item => {
-    const {id, deps, contents} = item;
-    _cache[item.id] = {id, deps, contents};
+    _cache[item.id] = {id: item.id, deps: item.deps, contents: item.contents};
   });
 }
 
--- a/lib/build/bundled-source.js
+++ b/lib/build/bundled-source.js
@@ -81,7 +81,9 @@ exports.BundledSource = class {
         id: moduleId,
         traced: bundler.getTraced(),
         fileExists: function(defaultExists, id, filePath) {
-          return !!bundler.getItemByPath(calculateFileName(filePath));
+          // always force amodro to use our fileRead func.
+          // this bypasses amodro bug https://github.com/amodrojs/amodro-trace/issues/6
+          return true;
         },
         fileRead: function(defaultRead, id, filePath) {
           let location = calculateFileName(filePath);
--- a/lib/build/bundled-source.js
+++ b/lib/build/bundled-source.js
@@ -14,7 +14,11 @@ function cacheTraced(traced) {
 }
 
 function cachedTraced() {
-  return Object.values(_cache);
+  let values = [];
+  for(let key in _cache) {
+    values.push(_cache[key]);
+  }
+  return values;
 }
 
 exports.BundledSource = class {
--- a/lib/build/bundled-source.js
+++ b/lib/build/bundled-source.js
@@ -5,22 +5,6 @@ const cjsTransform = require('./amodro-trace/read/cjs');
 const allWriteTransforms = require('./amodro-trace/write/all');
 const fs = require('../file-system');
 
-let _cache = {};
-
-function cacheTraced(traced) {
-  traced.forEach(item => {
-    _cache[item.id] = {id: item.id, deps: item.deps, contents: item.contents};
-  });
-}
-
-function cachedTraced() {
-  let values = [];
-  for(let key in _cache) {
-    values.push(_cache[key]);
-  }
-  return values;
-}
-
 exports.BundledSource = class {
   constructor(bundler, file) {
     this.bundler = bundler;
@@ -29,6 +13,7 @@ exports.BundledSource = class {
     this.includedBy = null;
     this.moduleId = null;
     this._contents = null;
+    this._deps = null;
     this.requiresTransform = true;
   }
 
@@ -38,6 +23,14 @@ exports.BundledSource = class {
 
   get path() {
     return this.file.path;
+  }
+
+  set deps(deps) {
+    this._deps = deps || [];
+  }
+
+  get deps() {
+    return this._deps;
   }
 
   set contents(value) {
@@ -86,7 +79,7 @@ exports.BundledSource = class {
       {
         rootDir: rootDir,
         id: moduleId,
-        traced: cachedTraced(),
+        traced: bundler.getTraced(),
         fileExists: function(defaultExists, id, filePath) {
           return !!bundler.getItemByPath(calculateFileName(filePath));
         },
@@ -118,13 +111,13 @@ exports.BundledSource = class {
       loaderConfig
     ).then(traceResult => {
       let traced = traceResult.traced;
-      cacheTraced(traced);
 
       for (let i = 0, ii = traced.length; i < ii; ++i) {
         let result = traceResult.traced[i];
         let item = bundler.getItemByPath(result.path);
 
         if (item) {
+          item.deps = result.deps;
           if (item.requiresTransform) {
             item.contents = result.contents;
             item.requiresTransform = false;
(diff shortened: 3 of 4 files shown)


AStoker:
I plan on looking at this. I tried it on our production setup though, and it failed silently. I need to dig in and find out why. In the mean time, have you tested this on a large scale site?

huochunpeng:
@Astoker bundler already caches processed file contents (_contents in BundledSource) in itemLookup.  This patch introduced another global cache for file contents again, so we actually have two copies of all files contents in memory.
This definitely have room to be improved. I am thinking about saving deps on BundledSource instance, and use bundler to rebuild the cache feed for amodro.

huochunpeng:
I only have one project in aurelia, I have not tested on other app.

huochunpeng:
I refactored the code to use bundler's internal cache. The behaviour may changed, try again pls.

huochunpeng:
Synchronised with the latest change of amodro-trace issue5-trace-cache branch.

EisenbergEffect:
@AStoker Are you able to look at the updated version?

AStoker:
Sorry, was gone for Thanksgiving. I'll give this another try momentarily (need to wait for VM Fusion to finish updating before I start up my Windows VM for work).

AStoker:
Tried again, and it still fails. It seems to be failing on the same file (I think). The last output was tracing a file that had this contents:
``` javascript
import 'jquery';
import ko from 'knockout';
import Q from 'q';
import _Form from 'form/viewer/form';
import 'form/viewer/autoSave/autoSave';

requirejs.s.contexts._.config.paths['template'] = _Form.Path;

var viewmodelModuleID = _Form.Path + 'viewmodel.js';
/*Global Loaders*/
window.$ = $; 
window.ko = ko;
window.Q = Q;

require([viewmodelModuleID], function (vm) {
    _Form.Initialize(vm);
});
```
Still seems to fail silently with no errors and no files written out. @huochunpeng, anything in here stick out to you that would cause it to fail?

AStoker:
It seems to be failing inside the amodro trace function, as I put some console logs around and it doesn't hit the `then` callback when trying to trace that module. Also, the amodro trace `catch` callback didn't fire with any error. So I'll have to dig in further.

huochunpeng:
I would guess it's because your source code try to modify requirejs contexts. This probably confused the newly updated amodro code.

Your source code assumes requirejs is in use instead of jspm/webpack.
I see you try to use ko component itself (which calls require underneath) to dynamicly load the template.
Is it possible to replace the dynamic ko VM require with two normal es6 import lines (one for the vm, one for a template)? Then refactor the _From to load both ko VM and template, this will remove the need of require call from ko side.

In es6 syntax, the dynamic loading is through `System.import`, not sure is there any support of that in requirejs bundler.

AStoker:
Update: it seems to be related to importing a specific file. Could be a circular dependency that's not being caught?... 
I have two files (one is the one in question above) that import `form/viewer/form`. When both of those imports are removed, tracing continues and finishes up... Works fine in previous setup, so there's definitely something strange.

AStoker:
This particular file is one of our "special" files unfortunately. Our system does only use RequireJS, so we're okay with this particular hack (if you can ever actually be "okay" with a hack I'm not sure). And we can't use all ES6 for this particular setup because it's part of a shared resource that gets run off an iframe that is not being transpiled (if we didn't have to support IE, we would be so set right now :P )

AStoker:
Part of my concern is going to be that even if there's some bad code in there, if it had worked previously, it would _hopefully_ continue to work. If not, then we need to identify specifically where the issue lies, and make sure that it's documented so that when people come in and complain that our update broke everything, we have a solution for them :)

AStoker:
Line that _appears_ to be the issue:
``` javascript
import _FormStyles from 'text!form/viewer/form.css';
```
We're using the RequireJS Text loader, so this should be working.

huochunpeng:
Maybe to load bluebird in cli and turn on all warnings to see if there is anything it complains about the missing resolve/reject.

AStoker:
You got a line number for that (save me some time searching)?

AStoker:
Can you try in your base project to import a css file as text? Confirm that it is just my machine.

huochunpeng:
I mean to use bluebird promise to replace default promise implementation at the beginning of cli.

huochunpeng:
I will try the css import when I was in office.

AStoker:
Thanks. In the mean time I'll try and load up bluebird and see if I can't get more detailed info.

AStoker:
@huochunpeng, tried pulling in promises and setting them globally. But didn't get any more detailed information.
Did this to pull it in (included the bluebird file minified):
``` javascript
const Promise = require('./bluebird');

Promise.config({
	warnings: true,
	longStackTraces: true
});

global.Promise = Promise;

```

huochunpeng:
Yes, I got problem too.
Better in my setup, it raises an error. There is a missing path for module "text!./test.css" from amodro tracing result.
Could you put one debug line here in `./node_modules/aurelia-cli/lib/build/bundled-source.js` after line 116:

      for (let i = 0, ii = traced.length; i < ii; ++i) {
        let result = traceResult.traced[i];
        if (!result.path) console.log(JSON.stringify(result));   // debug line
        let item = bundler.getItemByPath(result.path);

In my project, it prints `{"id":"text!test.css","dependents":["main"]}`, missing `path` field.
Please confirm if we have exact same issue.

AStoker:
I put that line there, but I still don't get any kind of error printed out. But what you're experiencing sounds like what I'm having problems with.

huochunpeng:
My issue might be different. I got same issue with official cli 0.22.0. My app doesn't use bundler to bundle src/**/*.css file, I modified my app aurelia_project/tasks/process-css.js to build less files from a different folder (css/ not src/ folder).

huochunpeng:
From a flesh new app built form `au new`. The `import css form "text!./test.css";` works in initial build but waiting infinitely on watch mode if I change the source code `//import...`. The behaviour is same on 0.22.0 or my branch.

huochunpeng:
Some finding.

When you don't use `import css form "text!./test.css";` in javascript file, only load css in template require tag. All css files in your source folder are bundled directly by cli itself (`tasks/process-css.js`) without hitting amodro, cli also use a pre-processor (none/less/scss) defined in aurelia.json file.

When you use import css in javascript code, it goes through amodro without the pre-processing, somehow, amodro (even the unpatched version) cannot get the path of the module. 

I can reintroduce the same issue (even on cli 0.22.0) on my newly generately fresh au app by modify `tasks/build.js` like this:
```
export default gulp.series(
  readProjectConfiguration,
  gulp.parallel(
    transpile,
    processMarkup,
    // processCSS, // disable css process, not it got the same issue like my app.
                              // if your problem is related, I guess it's a tracing condition in your code,
                              // your css process is too slow to finish before transpile to hit that js file.
  ),
  writeBundles
);
```

I can also fix it by forcing processCSS before transpile
```
export default gulp.series(
  readProjectConfiguration,
  processCSS, // somehow after bundler sees it, amodro can see the path 
  gulp.parallel(
    transpile,
    processMarkup,
    // processCSS,
  ),
  writeBundles
);
```

This doesn't fix watch mode.

Have a try to see if this changes behaviour on your app.

huochunpeng:
BTW, `import css form "text!./test.css";` is not supported by standard es6 syntax, the line implies the building system is requirejs.
In principle, I think it should not be supported (or need normalized syntax) by cli since cli is going to support jspm and webpack in future.

huochunpeng:
It's a bug in amodro.

EisenbergEffect:
@huochunpeng Are we waiting on James to hear his explanation on the amodro issue? Anything else we need to do in the mean time?

huochunpeng:
Yes, waiting for James. But not necessary for cli to move forward,
```
fileExists: function(defaultExists, id, filePath) {
      // return !!bundler.getItemByPath(calculateFileName(filePath));
      return true; // force anything goes through next fileRead func. Somehow amodro's default file read doesn't work inside aurelia-cli (a problem I could not reproduce outside of cli)
    },
```
can bypass the amodro issue, if @AStoker confirms it works for his app.

If the additional change is ok for you, it would be enough for cli to work as expected.

AStoker:
I'm aware that it's not standard ES6 Syntax. But if the developer knows what they're transpiling to, and knows the loader used, they should have the freedom to use that.  
The steps that are happening on our build are such that regular CSS files get bundled into the build as wells as the compiled Sass (compiled to CSS) files. So a Sass file that's referenced via the `.css` extension should be no different than the regular `.css` files in regards to how they're stored. The trouble is coming from the actual importing method.   

I can guarantee that I'm not the only one who is pulling in CSS as text with the `import` statement, so it would be nice to continue to allow this same setup.   

The reason that amodro is tracing the css files is because they're being pulled in from the ViewModel. Amodro doesn't trace css files when they're brought in from a `<require>` tag. 
So you think that the issue is that when amodro is tracing the javascript files, it can't recognize the css path (I tried processing the css before the transpiling step, but it didn't change the outcome)? Why would this have worked previously then?

AStoker:
@huochunpeng, I should learn to refresh my webpage before posting. I missed your previous comments :P 
I tried your suggestion, but it still failed. I'm going to try and get a basic repo up for you to compare against.

AStoker:
Side note, if I comment out the css import line, and then build/watch, then while the code is still being watched, uncomment the css import line, everything traces correctly and works as expected.

AStoker:
Could have been a package mishap on my end. I just updated my node modules and it seems to be working. I want to test out a bit more to confirm though.

huochunpeng:
@AStoker is there any difference between forcing `processCSS` before and after `transpile` for your app?

AStoker:
@huochunpeng, you mean this?
``` javascript

export let buildTask = gulp.series(
  clean,
  readProjectConfiguration,
  processCSS,
  gulp.parallel(
    transpile,
    processMarkup,
    processSVG,
    processImages
  ),
  processCSS,
  writeBundles

);
```
No, that does not help.  
I really want to get this thing pushed through :) I tried making a basic setup similar to my setup, but it seemed to work fine. So there's something in my setup that is causing the problem. The reason I want to wait until that is resolved is because it's a production setup that is large scale. If something had worked in one release, but not the next, it's some kind of regression that should be noted.  

If it's due to bad code, I want to know what bad code is causing the problem so that others who might run into this know what to do instead of just blaming the CLI :)

huochunpeng:
Can you just try processCSS after tranpile under current cli 0.22.0 (remove first processCSS) from the code you showed above)?
This is not meant to be a fix, just want to confirm your issue is from from timing of transpile and processCSS. If I guess right, this will expose your issue no matter what cli version you use.

There IS a bug in amodro, I cannot imaging your clean node_modules somehow fixed it.

huochunpeng:
You can try that basic app (attampt to reproduce the issue), forcing processCSS after transpile will reproduce it.

AStoker:
@huochunpeng, this is my typical setup, the css processor does happen after the transpile step:
``` javascript
export let buildTask = gulp.series(
  clean,
  readProjectConfiguration,
  gulp.parallel(
    transpile,
    processMarkup,
    processCSS,
    processSVG,
    processImages
  ),
  writeBundles
);
```
I just tried putting the processCSS task right after transpile, and removed any other steps, and still I get the same issue.

AStoker:
It would be helpful for context to have somebody else try this PR in their production setup, so that way it's not just me and my setup :)

AStoker:
Some additional details. The last traced module that seems to be causing the failure is the module that is importing the model that imports the CSS.
Module A imports Module B, Module B imports CSS.
Amodro traces Module B first (successfully finds the CSS files), then traces Module A, where everything now just stops.

huochunpeng:
I mean try processCSS after gulp.parellel(...transpile) WITH original cli 0.22.0. This will fail on you big app and fail on you basic app too.

If you read through my longest post in this thread, There is a sentive timing issue in your app (whether it hits amodrobug or not, is purely caused by timing of processCSS and transpile which started as parellel tasks).

huochunpeng:
I will try the module A -> B -> css scenario you described, when I reach my mac today.

AStoker:
This block [here](https://github.com/aurelia/cli/pull/419/files#diff-a98cd86a736a989ffcce0c1416db2d54R675) in the starting point of where things are failing for me. Still digging into it, but it seems related to how the tracing cache is used.

huochunpeng:
Sorry, @AStoker I made one mistake, I thought I tried processCSS after transpile but I didn't, The only thing I can successfully expose the issue on 0.22.0 was hiding processCSS.

Tracing happens in `writeBundles`, not `transpile`, my bad.

Can you try `au build` on 0.22.0 with this:
```
export let buildTask = gulp.series(
  clean,
  readProjectConfiguration,
  gulp.parallel(
    transpile,
    processMarkup,
    // processCSS, // of course the app wouldn't work without this step, but the build task should finish normally
    processSVG,
    processImages
  ),
  writeBundles
);
```

It will fail on the import css line. If you comment out import css line, the `au build` passes. This is what I can reproduce on a bare minimum app.

when you say
> I tried your suggestion, but it still failed.

Did you mean modifying local `node_modules/aurelia-cli/lib/build/bundled-source.js` with this patch

    fileExists: function(defaultExists, id, filePath) {
      // return !!bundler.getItemByPath(calculateFileName(filePath));
      return true; // force anything goes through next fileRead func. Somehow amodro's default file read on text!file.css doesn't work inside aurelia-cli
    }, ...

If not, please have a try. If this patch doesn't fix your issue, then we are probably investigating two separate issues.

huochunpeng:
@AStoker good news, I repeated your A -> B -> css problem in a small app **on 0.22.0** with untouched build task.

In [this app](https://github.com/huochunpeng/au-cli-import-css), main.js -> app.js -> test.css
Tracing main... fails silently.

This is definitely different issue from what I had, and `fileExists` patch does not help. But this confirms it's not an issue introduced by my branch, so you don't have to waste time on checking my branch.

If you can reproduce it, I will create a separate issue to track it.

huochunpeng:
@AStoker I got a strange fix for your A -> B -> css problem, modify `node_modules/aurelia-cli/lib/build/bundled-source.js` to disable stub

    writeTransform: allWriteTransforms({
       // stubModules: loaderConfig.stubModules 
    })

No more silent failure!

I don't know what went wrong with the stub, I could not reproduce it in pure amodro setup (out of aurelia).

AStoker:
@huochunpeng, glad you got it working and identified that issue. I'm still worried (less so about this PR though) because this issue _didn't come up until this PR_. So something in this PR is revealing an issue that should be resolved, and should probably be related so this PR doesn't mess up others.   

I tried the setup you gave, however, and it works fine for me on 0.22.0, but when I upgrade to your PR branch, it fails as it does in my production setup.

AStoker:
When this block is skipped (https://github.com/aurelia/cli/pull/419/files#diff-a98cd86a736a989ffcce0c1416db2d54R359), everything works as expected. Can you explain to me what is going on right here so I can try and understand better?

AStoker:
It looks like the `traceCacheEntry` already has the completed contents you're trying to compile into the regular `contents` variable. This if block creates a define block with the module dependencies and name, but not the contents. When I just set the `contents = traceCacheEntry.contents`, everything seems to work correctly. I think the reason this was breaking is because you're setting the `context._cachedFileContents[url] = contents`, and when the contents are an empty module, and the module in question is actually the `text` plugin module, when require tries to load the module using the text plugin, it fails because the text plugin doesn't actually do anything... Just some thoughts.

AStoker:
The issue, I believe, is stemming from (this commit)[https://github.com/amodrojs/amodro-trace/commit/32dc5cd49f758d95433cad4ea8931bd2f9069939] where the contents of a trace are being reused. Comparing the contents of the traced file to a newly traced file, it appears that they're different, which I wouldn't expect.

huochunpeng:
That is strange. I was consistently repeating the issue on 0.22.0. And fix works in 0.22.0.

My guess is my branch made some step faster, it might be a timing sensitive issue. I will try it on different machine.

The bug maybe even deeper. I am running latest nodejs 6.9.1 under macOS 10.12.

huochunpeng:
I don't believe the commit you pointed out is the root cause, as I can repeat the issue on 0.22.0 without that commit of amodro.

It might be a false alarm for you.

AStoker:
Could be a false alarm, but when I disable that block of code that was changed, everything works as expected...

AStoker:
Any ideas why the content would be different like I mentioned though? Could that be causing the problem?

huochunpeng:
That's why I suspect the bug is deeper. I will try more machines and more nodejs versions. What's your os + nodejs versions?

huochunpeng:
Update, I can repeat the A->B->css issue on 0.22.0 on my Macbook 2015 and iMac 2010 with node 4.6.2, 6.9.1 and 7.20.

So the bug is likely either in cli code or amodro code.

When I got time, I will do more investigation, now I have to spend time on my day job...

huochunpeng:
I spent bit more time here...

In A->B->css amodro define "text" module, but cached stubbed version in content cache, when it tries to load "text!test.css" module, the `lib/amodro-trace/write/plugins.js` need original "text" module (before stubbed) to process "text!test.css" module, hence it failed.

To approve it, edit `node_modules/aurelia-cli/lib/build/amodro-trace/write/stubs.js` to add some log.
```
module.exports = function stubs(options) {
  options = options || {};

  return function(context, moduleName, filePath, contents) {
    if (options.stubModules && options.stubModules.indexOf(moduleName) !== -1) {
      if (lang.hasProp(context.plugins, moduleName)) {
        console.log('stub1:'+moduleName);
        return 'define({load: function(id){console.log("===== :-( stub1 tries to load: "+id);' +
               'throw new Error("Dynamic load not allowed: " + id);}});';
      } else {
        console.log('stub2:'+moduleName);
        return 'define({load:function(name){console.log("===== :-( stub2 tries to load: "+name);}});';
      }
    } else {
      return contents;
    }
  };

};
```

And `au build` prints
```
...
Starting 'writeBundles'...
Tracing app...
stub1:text
Tracing environment...
Tracing main...
===== :-( stub1 tries to load: test.css
```

Bingo, when tracing main, it tries to use cached "text" module to process "text!test.css". And someone just silenced the `throw new Error("Dynamic load not allowed: " + id);`.
If you look carefully, "text" module is stubbed when tracing app.js, not stubbed again when tracing main.js.

Amodro use unintended global _cachedFileContents (inside one requirejs context) across calls, so it's actually not a pure function without side effect.

I am confident that I can repeat this issue in a pure amodro setup without aurelia. I will do that and raise an issue to amodro guy.

The reason that you only see the issue on my branch not 0.22.0 maybe is the order of tracing. Can you try my au-cli-import-css sample app again?

I suspect, when you see the issue with my cli branch, the tracing is like
```
Tracing app...
Tracing main...
```
When you didn't see the issue with 0.22.0, the tracing is like
```
Tracing main...
Tracing app...
```

AStoker:
Thanks for digging into that @huochunpeng, great work! I'll try that branch momentarily :)

AStoker:
When I modify your build file to skip processing the CSS, this is what I get as an error (doesn't build)
```
TypeError: Path must be a string. Received undefined
    at assertPath (path.js:7:11)
    at Object.normalize (path.js:1177:5)
    at normalizeKey (/Users/astoker/Downloads/au-cli-import-css-master/node_modules/aurelia-cli/lib/build/bundler.js:198:15)
    at Object.getItemByPath (/Users/astoker/Downloads/au-cli-import-css-master/node_modules/aurelia-cli/lib/build/bundler.js:88:28)
    at amodroTrace.then.traceResult (/Users/astoker/Downloads/au-cli-import-css-master/node_modules/aurelia-cli/lib/build/bundled-source.js:108:28)
    at process._tickDomainCallback (internal/process/next_tick.js:129:7)
```
Tracing it down to what path is being requested at `bundler.getItemByPath(result.path)`, `result.path` is `/Users/astoker/Downloads/au-cli-import-css-master/node_modules/text/text.js`

I also tried adding those comments in the write stubs on my production environment, and I didn't get any console messages :(

huochunpeng:
That bug needs different fix, force `fileExists` to always return true bypass that bug.

After that, you might now hit a->b->css bug, which can be bypassed with disabling stub.

In summary, we needa these two patches to work around two amodro bugs.

The first patch is transparent to end user, second one needs to be highlighted in release note.

huochunpeng:
@AStoker good new, I fixed A->B->css bug.

My previous feeling was wrong.
The bug is not in amodro, amodro-trace is still a pure function without side effect.

The issue is in cli, I am talking about master branch.

In `fileRead` option func that cli supplied, it always gets file contents from cached `bundler.itemLookup` where it retained all file contents **after writeTransform (including stub)**. So cli is caching the stubbed "text" module, this confused amodro when it tries to use "text" plugin code to load additional resource.

I have pushed two fixes up in my branch, one to bypass amodro-trace#6, another one to fix the stubbed cache issue (A->B->css).

It should be all good now, no visible behaviour change for end user. Have a try.

EisenbergEffect:
Just wanted to jump in here and thank you both for working hard on this. It's going to be a great contribution!

huochunpeng:
You are welcome.

huochunpeng:
@AStoker when you do code review, please skip `lib/build/amodro-trace/index.js` and `lib/build/amodro-trace/lib/loader/Loader.js`. The two files were copied from https://github.com/amodrojs/amodro-trace/tree/issue5-trace-cache.

AStoker:
@huochunpeng, were there any changes I had to make to my code in order for this to work? The production project still fails to build when importing a css file.

huochunpeng:
No change required, unless I missed something.

AStoker:
Dang. It's still not working on my production setup. Still failing in the same location. What kind of details can I get you to help with this?

huochunpeng:
You can try put debug line in write/stub.js to see if it still uses stubbed text module. Did you rebuild node_modules folder? Do check if it uses latest code.

AStoker:
Yes, I cleaned out the node modules and rebuilt.
I added those console lines, but there's no output for errors...

huochunpeng:
Sounds like never ending bug :-)
Just for clearing doubt, can you confirm local bundled-source.js has latest code?

AStoker:
You're telling me :)
Yes, always confirm. The bundled source _does_ have the latest code.
line 92: 
``` javascript
 if (found) {
            // ensure to use contents before stub
            return found.contentsBeforeStub || found.contents;
```
It's got all the changes

huochunpeng:
Guess there is one more bug to chase.

AStoker:
Any leads on where I should start looking in the changes? I'm not as familiar with how the caching of files works.

AStoker:
If I could just get _some_ kind of output in the console, I could start figuring things out. But the complete lack of logs is having me confused on where to start.

huochunpeng:
you can try add log in fileRead to see last found or unfound id and location.

AStoker:
Well, the fun part is, the file that is importing the css (we'll call this Module B) seems to "finish" it's tracing, but then the file (we'll call this Module A) that imports the file that brings in the CSS fails.
So the last "found" file is from Module A and not related to the CSS importing file. Module B (which does the CSS importing) says that it finds the CSS file correctly. 
The last "not found" file is farther up the tracing chain and in regards to the aurelia dialog module.

AStoker:
Aha! Found what was holding it up, but I don't understand fully the reasoning (perhaps @EisenbergEffect can clarify for me). 
My plugins config was written as so:
``` javascript
"plugins": [
        {
          "name": "text",
          "extensions": [
            ".html",
            ".css",
            ".svg"
          ],
          "stub": false
        }
      ]
```
So in this particular issue, there was no text plugin stub to use, and it caused it to fail. Setting the `stub: true` allowed the project to compile.  
*However*, this is not ultimately feasible, since for some reason (unknown to me, and I'd appreciate clarification), the `stub` property has to be set to false for the text plugin in order for me to load up files as text (such as svg's and css files). Thoughts?

huochunpeng:
My understanding is If you include svg file extension in app-bundle, you can stub text.

huochunpeng:
If you put log line in node_modules/text/text.js load func, you might see some output there. Beware amodro will silence any error it throws.

My guess is, in your app, even original text module cannot load the additional resource.

AStoker:
For the SVG files, I'm using the Aurelia Loader to load files like this:
```
 this.loader.loadText(`/dist/${svg}.svg`)
```
So it's making a request for files with the `.svg` extension. A question would be does any project with Text stubbed to be false work?

AStoker:
For reference, the error thrown when stub is set to true:
```
vendor-bundle.js:40400 Uncaught Error: Dynamic load not allowed: /dist/resources/images/svg/Cancel.svg
    at Object.load (https://andrews-pc10:44301/dist/vendor-bundle.js:40400:41)
    at Module.<anonymous> (https://andrews-pc10:44301/dist/vendor-bundle.js:5315:28)
    at https://andrews-pc10:44301/dist/vendor-bundle.js:4357:23
    at on (https://andrews-pc10:44301/dist/vendor-bundle.js:4738:21)
    at Module.callPlugin (https://andrews-pc10:44301/dist/vendor-bundle.js:5176:17)
    at Module.fetch (https://andrews-pc10:44301/dist/vendor-bundle.js:5045:46)
    at Module.check (https://andrews-pc10:44301/dist/vendor-bundle.js:5077:30)
    at Module.enable (https://andrews-pc10:44301/dist/vendor-bundle.js:5396:22)
    at Object.enable (https://andrews-pc10:44301/dist/vendor-bundle.js:5777:39)
    at Module.<anonymous> (https://andrews-pc10:44301/dist/vendor-bundle.js:5381:33)
    at https://andrews-pc10:44301/dist/vendor-bundle.js:4357:23
    at each (https://andrews-pc10:44301/dist/vendor-bundle.js:4282:31)
    at Module.enable (https://andrews-pc10:44301/dist/vendor-bundle.js:5333:17)
    at Module.init (https://andrews-pc10:44301/dist/vendor-bundle.js:5009:26)
    at https://andrews-pc10:44301/dist/vendor-bundle.js:5680:36
```

huochunpeng:
Did you copy all svg files to dist?

If you bundle all svg files in aurealia.json app-bundle section, you can avoid dynamic loading, so you can stub text since you don't need it for dynamic loading.

AStoker:
Yup, all the SVG's exist in `dist`.
I had considered doing that. But we have a large amount of SVG's, and the decision was made to keep them separate to load up only when needed (saves initial load time in our app).

AStoker:
How you're describing it is exactly what I do for my smaller side projects :P

huochunpeng:
I will try to repeat a->b->css issue with stub:false.

This is crazy, bug on either side.

AStoker:
Something the heck is going on, and I don't necessarily believe it's in your PR. But for some reason I think your PR just brings to light an existing issue, which is kind of a bummer because it should be solved so we don't have people updating and having things break.

huochunpeng:
You can use dedicated bundle (or bundles to keep size small) for svg files. It should improve performance than loading every single file. Anyway this only bypasses the bug.

huochunpeng:
BTW, initial loading time is only for vendor-bundle. All other bundles are loaded on demand. I create many bundles to keep performance up.

AStoker:
I'll probably end up refactoring it all then :) In the mean time, I'd still like to find out why this fails on the stub change...

huochunpeng:
Yes of course. It might be difficult for me to reproduce.

You might modify write/plugins.js to let it show plugin error instead of silencing it.

huochunpeng:
@AStoker I could not reproduce it. Could you share your aurelia.json?

EisenbergEffect:
How are we doing on this?

EisenbergEffect:
@huochunpeng @AStoker What's the status of this?

huochunpeng:
I have not did any change for both of my PRs #419 and #431 since early Dec.
#431 is very aggressive, but it does likely avoid the annoying tracing bug in #419 thread.

AStoker:
@EisenbergEffect, the stall on this PR was that it crashed hard on what was apparently a pre-existing bug. Until that pre-existing bug is resolved, I feel like this might be a bit problematic to merge.

JeroenVinke:
Hey @huochunpeng, thanks for your great work on this. Would you mind explaining the use case for [this bit](https://github.com/huochunpeng/cli/blob/ca1f6c0d803a4e328da5b8acd2726db7a2f9d9de/lib/build/amodro-trace/lib/loader/Loader.js#L360-L366)? When removing this piece of code it builds just fine and resolves the A -> B -> CSS issue. Thanks!

huochunpeng:
That is a good catch.

That is not my code, that is original amodro trace code. It is the old simple cache (before amodro implement full cache for us). I fixed our usage of cache (new cache) for the css issue, but did not touch the old one.

Since amodro has full cache now, it makes sence to turn off the old cache code. There is more code can be deleted.

huochunpeng:
Sorry, maybe it is new code in amodro cache branch which is still provided by original author. I cannot see thw whole piece on my phone. I will checked it out later. 

All amodro code are from original author, I just copied that branch to cli.

JeroenVinke:
That would be much appreciated. Let me know if you need a second pair of eyes on something. Just send me a private message on Gitter (username is @JeroenVinke)

EisenbergEffect:
Fixed.

Issue #423, opened 2016-11-29
Emit accurate sourcemaps for application plugins.


shaunluttin:
**I'm submitting a bug report**

* **Library Version:** 0.20.0
* **Operating System:** Windows 10
 * **Node Version:** 6.9.1
* **NPM Version:** 3.10.8
* **Browser:** Chrome Version 54.0.2840.99 m (64-bit)
* **Language:** TypeScript 2.0.10

**Current behavior:**

If an Aurelia app references an Aurelia plugin, and if that plugin ships with JavaScript source maps, then the source mapping that the Aurelia bundler produces will not work with the Chrome debugger. This happens because the resultant `*-bundle.js.map` file has several problems.

* the `sources` for the plugin use full paths
* those `sources` point to JS not TS files, and
* the `sourceRoot` is not the application root

**Expected/desired behavior:**

We expect the Chrome debugger to work with the plugins source maps. In the `*-bundle.js.map` file, we expect that:

* the `sources` for the plugin use relative paths
* those `sources` point to TS files, and
* the `sourceRoot` will be the application root

The motivation is to be able to debug an application's plugins. 

**See also**

https://github.com/shaunluttin/aurelia-open-id-connect/issues/10 demonstrates how we have hand edited the `app-bundle.js.map` file to make the Chrome debugger work with our plugins source maps.

This is the `scripts/app-bundle.js.map` file that the bundler produced. It does not play well with the Chrome debugger.

```
{
  "version": 3,
  "sources": [
    "src/app.ts",
    "src/environment.ts",
    "src/home.ts",
    "src/open-id-connect-configuration.ts",
    "src/main.ts",
    "src/user-profile.ts",
    "src/resources/index.ts",
    "C:/dev/aurelia-open-id-connect/demo/node_modules/aurelia-open-id-connect/dist/amd/plugin.js",
    "C:/dev/aurelia-open-id-connect/demo/node_modules/aurelia-open-id-connect/dist/amd/open-id-connect-logger.js",
    "C:/dev/aurelia-open-id-connect/demo/node_modules/aurelia-open-id-connect/dist/amd/open-id-connect-configuration.js",
    "C:/dev/aurelia-open-id-connect/demo/node_modules/aurelia-open-id-connect/dist/amd/open-id-connect.js",
    "C:/dev/aurelia-open-id-connect/demo/node_modules/aurelia-open-id-connect/dist/amd/open-id-connect-routing.js",
[...]

PR #426, opened 2016-11-30
fix/accurate sourcemaps

shaunluttin:
This is a work in progress. Comments are welcome. Resolves https://github.com/aurelia/cli/issues/423

### DIFF
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -127,6 +127,29 @@ exports.Bundle = class {
         let currentFile = files[i];
         let sourceMap = buildOptions.sourcemaps ? currentFile.sourceMap : undefined;
 
+        function acquireSourceMapForDependency(file) {
+          if (!file || !file.path) {
+            return;
+          }
+
+          let fileDirectory = path.posix.dirname(file.path);
+          let converter = Convert.fromMapFileSource(file.contents.toString(), fileDirectory);
+          sourceMap = converter
+            ? converter.sourcemap
+            : null;
+
+          if (sourceMap != null) {
+            // for some reason, path.posix.relative(from, to) does not work here
+            sourceMap.sourceRoot = fileDirectory.substring(process.cwd().length);
+          }
+
+          return sourceMap;
+        }
+
+        if (!sourceMap) {
+          sourceMap = acquireSourceMapForDependency(currentFile);
+        }
+
         if (sourceMap) {
           needsSourceMap = true;
         }
@@ -157,7 +180,7 @@ exports.Bundle = class {
       let mapFileName = bundleFileName + '.map';
       let mapSourceRoot = path.posix.relative(
         path.posix.join(process.cwd(), platform.output),
-        path.posix.join(process.cwd(), this.bundler.project.paths.root)
+        process.cwd()
       );
 
       console.log(`Writing ${bundleFileName}...`);
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -127,6 +127,19 @@ exports.Bundle = class {
         let currentFile = files[i];
         let sourceMap = buildOptions.sourcemaps ? currentFile.sourceMap : undefined;
 
+        function fileIsDependency(file) {
+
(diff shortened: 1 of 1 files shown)


shaunluttin:
As of tag `alpha01` the Chrome DevTools display the source maps like this. That is exactly how we would like Chrome to display them.

capture

EisenbergEffect:
Thanks @shaunluttin !  @AStoker Can you take a look at this?

shaunluttin:
@EisenbergEffect @AStoker I've had today and yesterday to work on this. My schedule fills up between now and Sat 03 Dec, at which point I will have time again to work on this. :watch:

EisenbergEffect:
We really appreciate it. Hopefully @AStoker can do a review between now and then to get you some feedback.

lib/build/bundle.js
```diff
@@ -157,7 +184,7 @@ exports.Bundle = class {
       let mapFileName = bundleFileName + '.map';
       let mapSourceRoot = path.posix.relative(
         path.posix.join(process.cwd(), platform.output),
-        path.posix.join(process.cwd(), this.bundler.project.paths.root)
+        path.posix.join(process.cwd())
```
AStoker: Why does this need to be a join? There aren't two things two join together.

lib/build/bundle.js
```diff
@@ -127,6 +127,33 @@ exports.Bundle = class {
...
+          // todo: Make obtaining the directory more robust.
+          let fileDirectory = file.path.substring(0, file.path.lastIndexOf("/") + 1);
+          let converter = Convert.fromMapFileSource(file.contents.toString(), fileDirectory);
+          sourceMap = converter
+            ? converter.sourcemap
+            : null;
+
+          if (sourceMap != null) {
+            let cwd = process.cwd();
+            let fileDirectoryRelativePath = fileDirectory.substring(cwd.length)
```
AStoker: Can you not just say `fileDirectory.substring(process.cwd().length)` instead of making an additional variable?

lib/build/bundle.js
```diff
@@ -127,6 +127,33 @@ exports.Bundle = class {
...
+            sourceMap.sourceRoot = fileDirectoryRelativePath;
```
AStoker: Same goes here. `fileDirectoryRelativePath` isn't used anywhere else in this block, so maybe just do `sourceMap.sourceRoot = fileDirectory.substring(process.cwd().length)`?

shaunluttin: Arguably, using an extra variable clarifies intent. That being said, it isn't terribly important, so I changed it @AStoker.

shaunluttin:
@AStoker @EisenbergEffect Review has stalled on this one. Is it simply that things are busy?

EisenbergEffect:
@AStoker is the best person to review. He's been busy. We have one other major CLI improvement related to perf I know he's spending a lot of time on.

lib/build/bundle.js
```diff
@@ -127,6 +127,39 @@ exports.Bundle = class {
...
+        }
+
+        function acquireSourceMapForDependency(file) {
+          if (!file || !file.path) {
+            return;
+          }
+
+          let parsedPath = path.parse(file.path);
+
+          let converter = Convert.fromMapFileSource(
```
shaunluttin: `fromMapFileSource` fails with some files, because the bundle has appended a `define` statement to the end of the file.

shaunluttin:
What the status of the review? Is this PR of interest?

EisenbergEffect:
@shaunluttin Sorry for the delay on this. With the holidays and several of our core team members moving, etc. we've slowed down during December. @AStoker Will you have time to take a look at this soon?

shaunluttin:
@EisenbergEffect @AStoker Not to worry. It make sense that times are busy right now. Thank you for letting me know.

EisenbergEffect:
@shaunluttin Do you feel this is in a good place for review? @AStoker Can you take a look?

EisenbergEffect:
Looks good to me. Merging.

PR #363, opened 2016-10-07
fix(package-analyzing): allow deeper levels for source root

alpox:
As i announced, here is the PR i provided earlier.

Before, the package-analyzer looked for the node_modules folder in only exactly one path step back
to the source files.
I changed this behaviour to use a node-like lookup which looks for the node_modules folder in
the following manner:
./node_modules, ../node_modules, ../../node_modules, etc.

And this allows to put the root of the source (project.paths.root) somewhere deeper in a file-structure.

See Issue https://github.com/aurelia/cli/issues/278

This is a new PR for: https://github.com/aurelia/cli/pull/295

PR #427, opened 2016-11-30
fix(package-analyzing): allow deeper levels for source root

alpox:
@EisenbergEffect as you asked for, here is the new PR for https://github.com/aurelia/cli/pull/363.
This time it should be right :-)

Actual description:
>Before, the package-analyzer looked for the node_modules folder in only exactly one path step back
to the source files.
I changed this behaviour to use a node-like lookup which looks for the node_modules folder in
the following manner:
./node_modules, ../node_modules, ../../node_modules, etc.

>And this allows to put the root of the source (project.paths.root) somewhere deeper in a file-structure.

Remark:
>As a quick remark:
>When putting the sources in a deeper structure, one has to set the "build.targets.useAbsolutePath" to true in the aurelia.json, otherwise requirejs tries to load the scripts from a wrong directory (I couldn't track this down so well, maybe you have some insight?)
Otherwise it works fine with my test setups and additional module load tries. One has to go sure that all paths to the source files in aurelia.json are adjusted. (This could lead to some trouble if one just adjusts the "paths.root" only.

Best, alpox

### DIFF
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -35,9 +35,14 @@ exports.Bundle = class {
     return Promise.all(
       dependencies
         .filter(x => bundler.itemIncludedInBuild(x))
-        .map(dependency => bundler.configureDependency(dependency)
-        .then(description => bundle.addDependency(description)))
-    ).then(() => bundle);
+        .map(dependency => bundler.configureDependency(dependency))
+    )
+    // Add dependencies in the same order as they were entered
+    // to prevent a wrong module load order.
+    .then(descriptions => 
+      descriptions.forEach(description => 
+        bundle.addDependency(description)))
+    .then(() => bundle);
   }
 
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
@alpox Do you feel that this is ready to go now? I know you had some reservations earlier on. I just wanted to make sure you felt it was ready before I did a final review and merge. Let me know.

alpox:
@EisenbergEffect From my side its ready as is.
One little question i have though, just to go perfectly sure:

I used `Promise.all` to go sure that the dependencies are loaded in the same order as the input used to be. I didn't find that many resources about the fact that `Promise.all` provides the results in the same order as the input promises. It works nevertheless.

Can this be seen as a given or would it make more sense to write an own implementation which goes sure?

EisenbergEffect:
It should be fine as is. Thanks!

PR #431, opened 2016-12-06
Remove tracing

huochunpeng:
**This branch is not in clean state to be merged.**

As expected, `au build` is slightly faster than #419.

While working on #419 , I realise cli doesn't really need amodro-trace. What amodro-trace brings to the table are two things.
1. ensure all deps are bundled. (this can be achieved with modified aurelia.json)
2. tranform all modules, more explicitly only need `write/defines.js` and `write/stubs.js`. (even stubs.js is not really needed, if we don't provide feature to stub module)

This branch removed tracing step, replaced it with just writeTransform on "cjs", "defines" and "stubs".

To use `"aurelia-cli": "git://github.com/huochunpeng/cli.git#remove_tracing",` in your app's package.json, you need to modify your `aurelia.json` to ensure all deps are bundled without tracing.

Luckily, you only need to modify vendor-bundle section to add `"resources": ["**/*.js"]` for some aurelia packages.

For reference Here is my vendor-bundle section. I modified **"aurelia-dialog", "aurelia-templating-resources", "aurelia-templating-router", "aurelia-testing", and "aurelia-i18n"**.
```
{
        "name": "vendor-bundle.js",
        "prepend": [
          "node_modules/whatwg-fetch/fetch.js",
          "node_modules/keymaster/keymaster.js",
          "node_modules/bluebird/js/browser/bluebird.core.js",
          "scripts/bluebird_turnon_cancellation.js",
          "css/swell/js/vendor/jquery-1.11.2.min.js",
          "css/jquery.gritter/js/jquery.gritter.min.js",
          "css/swell/js/vendor/bootstrap.min.js",
          "node_modules/moment/min/moment.min.js",
          "css/swell/js/vendor/bootstrap-datetimepicker.js",
          "css/swell/js/controls.js",
          "css/bcxlib.js",
          "node_modules/highcharts/highcharts.js",
          "scripts/highcharts_options.js",
          "node_modules/requirejs/require.js"
        ],
        "dependencies": [
          "aurelia-binding",
          "aurelia-bootstrapper",
          "aurelia-dependency-injection",
          "aurelia-event-aggregator",
          "aurelia-fetch-client",
          "aurelia-framework",
          "aurelia-history",
          "aurelia-history-browser",
          "aurelia-loader",
          "aurelia-loader-default",
          "aurelia-logging",
          "aurelia-logging-console",
          "aurelia-metadata",
          "aurelia-pal",
          "aurelia-pal-browser",
          "aurelia-path",
          "aurelia-polyfills",
          "aurelia-route-recognizer",
          "aurelia-router",
          "aurelia-task-queue",
          "aurelia-templating",
          "aurelia-templating-binding",
          "text",
          {
            "name": "aurelia-dialog",
            "path": "../node_modules/aurelia-dialog/dist/amd",
            "main": "aurelia-dialog",
            "resources": [
              "**/*.js"
            ]
          },
          {
            "name": "aurelia-templating-resources",
            "path": "../node_modules/aurelia-templating-resources/dist/amd",
            "main": "aurelia-templating-resources",
            "resources": [
              "**/*.js"
            ]
          },
          {
            "name": "aurelia-templating-router",
            "path": "../node_modules/aurelia-templating-router/dist/amd",
            "main": "aurelia-templating-router",
            "resources": [
              "**/*.js"
            ]
          },
          {
            "name": "aurelia-testing",
            "path": "../node_modules/aurelia-testing/dist/amd",
            "main": "aurelia-testing",
            "env": "dev",
            "resources": [
              "**/*.js"
            ]
          },
          {
            "name": "intl",
            "path": "../node_modules/intl/dist",
            "main": "Intl.min"
          },
          {
            "name": "i18next",
            "path": "../node_modules/i18next/dist/umd",
            "main": "i18next"
          },
          {
            "name": "aurelia-i18n",
            "path": "../node_modules/aurelia-i18n/dist/amd",
            "main": "aurelia-i18n",
            "resources": [
              "**/*.js"
            ]
          },
          {
            "name": "nprogress",
            "path": "../node_modules/nprogress",
            "main": "nprogress",
            "resources": [
              "nprogress.css"
            ]
          },
          "lodash",
          "numeral",
          "node-uuid",
          "dropzone",
          {
            "name": "mock/generate",
            "path": "../mock",
            "main": "generate",
            "env": "dev"
          },
          {
            "name": "codemirror",
            "path": "../node_modules/codemirror",
            "main": "lib/codemirror",
            "resources": [
              "addon/selection/active-line.js",
              "addon/display/fullscreen.js",
              "addon/display/fullscreen.css",
              "addon/dialog/dialog.js",
              "addon/dialog/dialog.css",
              "addon/search/searchcursor.js",
              "addon/search/search.js",
              "addon/mode/simple.js",
              "mode/shell/shell.js",
              "mode/perl/perl.js",
              "mode/ruby/ruby.js",
              "mode/python/python.js",
              "mode/sql/sql.js",
              "mode/powershell/powershell.js",
              "lib/codemirror.css",
              "theme/abcdef.css"
            ]
          }
        ]
      }
```

I am running my production app against this branch.
Welcome more testing.


huochunpeng:
@AStoker there is unlikely any A->B-css bug here :-)

huochunpeng:
To clarify, I have been working on this branch for more than a week. This is not created out of frustration of A->B-css bug in #419.

EisenbergEffect:
Interesting... @AStoker What do you think of this?

AStoker:
I'm thinking that removing the tracing step would cut out the largest time hog of our building process :) _And_ it would remove the hidden problem that we were trying to solve earlier. Things at the office have picked up (and will likely remain picked up till the end of the year), so I'll look at it as soon as I can. But I don't want to be the reason that things are being held up.

AStoker:
@huochunpeng, so I guess one downside to this would be that you'd be bundling _all_ the resources of a module, which potentially could be much more than you want?

TylerJPresley:
I appreciate the work being done here. Quick suggestion. What if it didn't trace for dev. Making dev faster. I wouldn't mind waiting for the trace when it's building my production bundles.

huochunpeng:
@AStoker that is correct. With tracing, those **/*.js was picked by amodro-trace and added to other bundles rather than in vendor-bundle (since they are deps for the modules listed in vendor-bundle).
But the downside is only true if npm package author compile amd target in separate files like some aurelia package do.

huochunpeng:
Sorry, clicked wrong button.

huochunpeng:
@TylerJPresley different behavior between dev and prod might supprise users. Futhermore, we do encounter some special issue with tracing in #419.

Since most npm package did not compile amd target in separate files which defeats tracing effort, the bundle size is not a big issue.

huochunpeng:
There is one downside for tracing.

I have multiple bundles, due to how tracing works, some aurelia modules are not bundled in vendor-bundle.js, they landed in the first app bundle. This makes not only vendor-bundle a hard dep for other bundles, but also my first app bundle a hard requirement for other bundles.

The result is when I load `http://localhost/job/index` page, it will load vendor-bundle, plus app-bundle, and then job-bundle, instead of just vendor-bundle + job-bundle.

This downside of tracing can be fixed by adding `"resources": ["**/*.js"]` in those aurelia packages in aurelia.json, but this fix just gives up the benefit (minimum deps) that tracing provides.

huochunpeng:
Here is a comparison between cli@0.23.0 and this branch with no minify on my app. I would say the size increment is insignificant.

```
                       0.23.0 rm-trace     diff
app-bundle.js          435191   290064  -145127
blueprint-bundle.js    360220   360220        0
catalog-bundle.js       14844    14844        0
dashboard-bundle.js     34754    34754        0
deployment-bundle.js   224963   224963        0
job-bundle.js           41134    41134        0
vendor-bundle.js      2670985  2829911  +158926

total                 3782091  3795890   +13799 (+0.36%)
```

EisenbergEffect:
Perhaps it shouldn't trace anything that is `source` but only trace `dependencies`? and only if they have a package configuration, rather than a single file?

huochunpeng:
I don't quite understand your proposal. If only trace the dependencies, there is no way to have "tree shaking" effect. What else benefit can trace provide?

If it is just to remove `"resources": ["**/*.js"]`, I think changing the way aurelia-* libs generating AMD target could reduce the lib dependency conifg in aurelia.json into one line, just `"aurelia-dialog"`, instead of `{"name":"aurelia-dialog",...}`. Propose to generate a single file UMD target to replace both AMD & commonjs targets.

EisenbergEffect:
I'd love to pick this back up and see where we want to take it.

fabioluz:
@AStoker Is there anybody taking this one? I'm very interested in this enhancement

EisenbergEffect:
We'd love any additional help on exploring this. @fabioluz

AStoker:
Is this something we still want after the improvements to the cli tracing recently?

JeroenVinke:
I think we're OK for now with the tracing, but if we tracing becomes a bottleneck again we can always pick this back up again

JeroenVinke:
Closing this for now

PR #436, opened 2016-12-14
feat(typescript): set lib to es2017 for upcoming js features

Vheissu:
The "es2017" value encompasses ES2015 and ES2016 as well as the upcoming ES2017.

### DIFF
--- a/lib/resources/content/tsconfig.json
+++ b/lib/resources/content/tsconfig.json
@@ -10,7 +10,7 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "moduleResolution": "node",
-    "lib": ["es2015", "dom"]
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
@Vheissu What does this specifically include in the case of TS? Does this only make it provide definitions for new libraries?

Vheissu:
@EisenbergEffect This is purely a typings related change for TypeScript. Setting the lib to ES2017 provides type hinting for all Javascript features including the upcoming ES2017 specification. This brings the TypeScript CLI project inline with the same configuration the Webpack TypeScript skeleton uses: https://github.com/aurelia/skeleton-navigation/blob/master/skeleton-typescript-webpack/tsconfig.json

EisenbergEffect:
Thanks!

PR #437, opened 2016-12-14
chore(commands): add alias for cli commands

AbraaoAlves:
Resolves: aurelia#412
Add optinal abbreviation for au commands with ./commands/alias.json


EisenbergEffect:
Thanks!

PR #441, opened 2016-12-15
Add configure template processing step

bigopon:
Note: All step Ids from `cssProcessor` has been increased by 1


EisenbergEffect:
I'm confused by the creation/use of the new "markup-select" type in the workflow. Is there some reason you can't use "input-select"?

bigopon:
Oh right, I thought input-select was only for transpiler. You merged it though, so can you fix it or I can commit another one if necessary?

EisenbergEffect:
I just pushed a set of fixes. It should all be working now :)

PR #447, opened 2016-12-27
fix: user relative path for source maps with empty `mappings` property

nenadvicentic:
When `sourceMap` object with empty `mappings` property is encountered, bundled source map contains absolute instead of relative path of source file.

### DIFF
--- a/lib/build/concat-with-sourcemaps/index.js
+++ b/lib/build/concat-with-sourcemaps/index.js
@@ -77,9 +77,7 @@ Concat.prototype.add = function(filePath, content, sourceMap) {
         });
       }
     } else {
-      if (sourceMap && sourceMap.sources && sourceMap.sources.length > 0)
-        filePath = sourceMap.sources[0];
-      if (filePath) {
+      if (sourceMap && sourceMap.file) {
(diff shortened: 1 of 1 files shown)


nenadvicentic:
This is stilla but in aurelia-cli 0.24.0.

This is screenshot from Chrome to illustrate what the issue is.

aurelia-absolute-path

EisenbergEffect:
Can you submit this PR here? https://github.com/floridoo/concat-with-sourcemaps
That's the original. If they accept the PR there, then ping us by creating an issue and we'll update our copy from that source.

nenadvicentic:
Ah OK. Did not figure it's from another repository. I'll do that.

nenadvicentic:
I checked [https://github.com/floridoo/concat-with-sourcemaps](https://github.com/floridoo/concat-with-sourcemaps)

Implementation has no notion of project root or paths relative to the root. It just concatenates what's past in as input parameters. Problem is why `aurelia-cli` passes absolute path to the `concat.add`:

`
concat.add(currentFile.path, currentFile.contents, sourceMap ? JSON.stringify(sourceMap) : undefined);
`

This line is in the `lib/build/bundle.js`

PR #449, opened 2016-12-28
chore(test):Add missing .js extensions after normalization

s-hoff:
Shamelessly copied from http://stackoverflow.com/a/40246316 after I bumped into this issue myself.
In my opinion, an new cli project should support this out of the box. This would otherwise get especially confusing with TypeScript as one would need to use a .js extension in the import for a .ts file.


EisenbergEffect:
Thanks!

PR #454, opened 2016-12-31
Package importer

JeroenVinke:
This pull request aims to add package importer functionality to the cli. For a quick visualization of the changes in this PR we have made a small video which can be viewed [here](http://www.screencast.com/t/kZmu1S436Iw).

# 1. Definition of terms
**Install** - The term install refers to the process of fetching a package from a remote registry (npm) and copying it to the local storage ( `node_modules` )

**Import** - Add configuration to the aurelia.json file so that an already installed package can be used with aurelia-cli

# 2. Commands
Two new commands will be added to aurelia-cli:

1. `au import <package>`
  When running this command the importer module will import an already installed package for use.

   Example usage:
   1. `npm install aurelia-dialog` (or yarn)
   2. `au import aurelia-dialog`

2. `au install <package>`
   When running `au install <package>` the package will be installed (through npm/yarn) and will be imported afterwards.

   Example usage:
   1. `au install aurelia-dialog`

Both commands support the `--debug` flag.

# 3. Installing the fork
1. `git clone https://github.com/monterey-framework/cli.git`
2. `cd cli`
3. `git fetch`
4. `git checkout feat/import`
5. `npm install`
6. `npm install esprima babel-polyfill babel-register`
7. `npm install git+https://git@github.com/gulpjs/gulp.git#4.0`
8. `npm link`
9. `cd ..`
10. `au new` and use defaults
11. `cd aurelia-app`
12. `npm link aurelia-cli`

# 4. Flow
[Open in new tab](https://cloud.githubusercontent.com/assets/2189477/20866751/a2f0a798-ba34-11e6-837c-2fffb7f17e2f.png)
img

The importer will first determine whether to install the package or just configure it (based on whether the `au import` command or the `au install` command is used). Then the importer will run through a list of (currently) 6 importer strategies. When a strategy is unable to "handle" the configuration of the plugin, it proceeds to the next strategy on the list.

After the best possible importer strategy has been found, the importer strategy does its thing. Then, the user is notified of which strategy has been used to configure the plugin.

After the plugin has been configured, the importer will look in the importer section of the `package.json` file. If there is a `tutorial` field, the contents of the field (which is an array of strings) will be printed.

# 5. Strategies

Currently 6 different importer strategies were implemented. Four of these strategies will be integrated in the importer and cannot be modified by plugin authors. Plugin authors can create instructions for the remaining two strategies. This is explained below.

## Metadata in package.json

There are two ways for plugin authors to instruct the aurelia-cli importer.

First, in the `aurelia` section of the package.json file, plugin authors can add an `import` section. When the aurelia-cli importer detects this section, it will execute the instructions within.

Full-featured example of `package.json`:

```
{
  "name": "my-plugin",
  "aurelia": {
    "import": {
      "patches": [
        { "op": "replace", "path": "/build/loader/plugins/0/stub", "value": false}
      ],
      "dependencies": [],
      "bundles": [
        ...
      ],
      "tasks": [
        "prepare-materialize"
      ],
      "scripts": {
        "install": [
          "au prepare-materialize",
          "node node_modules/requirejs/bin/r.js -o tools/rbuild.js"
        ]
      },
      "tutorial": [
        "1. in main.js add .plugin('aurelia-materialize-bridge')"
      ]
      }
    }
  }
}
```

### Sections:

**[patches]**

Apply patches to `aurelia.json` using RFC6902 standard.

**[dependencies]**

Append new dependency configurations to the default/specified `bundle.dependencies` array.

**[bundles]**

Add new bundles or even override an entire, existing bundle.

**[tasks]**

Copy custom CLI-tasks from package folder into `aurelia_project/tasks` folder.

**[scripts]**

You can execute any number or type of node scripts as needed.

**[tutorial]**

Allows plugin authors to explain to the user how the plugin can be used.

## Custom importer strategy
Plugin authors may find that adding importer instructions to the package.json file is not sufficient. In this case the plugin author can create an `importer-callbacks.js` file.

This `importer-callbacks.js` file could look like this:

```
class ImporterCallbacks {

 static inject() { return ['ui']; };

 constructor(ui) { }

 /**
 * Determine whether this particular strategy can be used or not
 * @returns {boolean}
 */
 determine() {
  return true;
 }

 /**
 * Main entry point called by ImportEngine
 */
 execute() {
  return {
   dependencies: [{
    name: "my-plugin",
    path: "../node_modules/my-plugin",
    main: "index.js"
   }]
  };
 }

 /**
 * Short description display in CLI output
 * @returns {string}
 */
 get name() {
  return 'Custom Importer Strategy';
 }
}
```

Whenever the importer encounters an `importer-callbacks.js` file, it loads the file. If `determine()` returns `true` the importer will call the `execute()` method.

The return value of `execute()` is metadata that is equivalent to what can be defined in the `"aurelia"."import"` section of the plugin's package.json.

# 6. Changes to the CLI
1. aurelia-logger was added to `lib/cli.js`, and the existence of the `--debug` flag determines whether the loglevel is info or debug. `lib/logger.js` is the custom logger for the cli, which always uses `console.log` since `console.info` etc are unavailable.
2. A couple of methods were added to `lib/file-system.js`: existsSync, copySync, resolve, join, statSync
3. The `package-managers` folder was added, containing both the `npm` and the `yarn` implementation of `install(packages, options)`
4. The `aureliaJSONPath` is now set on the Project (`lib/project.js`)
5. The `multiselect()` function was added to `ui.js` to allow a user to select multiple CSS resources at once.
6. The following dependencies were added:
- aurelia-logging
- bluebird
- node-glob (search for .css files in a package)
- rfc6902 (to patch aurelia.json via rfc6902)
- semver (to find the closest matching version of a package configuration in the cli's registry)

# 7. Additional CLI Feature suggestions:

### 7.1 Extending `au new` wizard steps with "Choose default package manager" option:

A new field has been added to `aurelia.json` called `packageManager` of which the default value is 'npm' (and 'yarn' is supported). Currently there is no way to change this value other than opening the aurelia.json file in a text editor and manually patching the value.

A new question could be added to the `au new` wizard asking the user which package manager should be used for the new project. The default should be 'npm' here, as not everyone has `yarn` installed.

### 7.2 Using npms (https://npms.io) services to search npm registry
While this functionality works best in Monterey GUI context, where:

1. User starts typing into the `autocomplete` box, which finds the package in real time using **[npms API](https://api-docs.npms.io/)** and populates the list of available version numbers
2. User selects the version number (current version is selected by default).
3. Actions 1. and 2. together enable the `Add this pair` button
4. Click on `Add this pair` button adds a package to the list of packages to be imported.

<p align=center>
</img>
<br><br>
</p>

we plan to make a similar, properly restricted service available to Aurelia CLI application as well. See **[Provide npms based package search](https://github.com/monterey-framework/cli/issues/28)** for more information.

### 7.3 Assisting the user in the process of accessing the imported a package.

While testing the initial importer code, we made a random **[selection of packages](https://github.com/monterey-framework/cli/issues/12)** and encountered many problems trying to invoke the code from the package. So, we came up with the idea to advice / intelligence / proactive support for the developer that has "installed" and "imported" a package and then sits stupefied pondering how to invoke the code from his app.

For most packages there is a two step process for importing package:

**Loading CSS**

The importer can detect css files in packages and add them to the `resources` property in a dependency configuration. A good example is `flatpickr`:
```
{
"name": "flatpickr",
"path": "../node_modules/flatpickr",
"resources": [
"dist/flatpickr.css"
],
"main": "dist/flatpickr.js"
}
```

The user can load the css file with: `<require from="flatpickr/dist/flatpickr.css"></require>`. The importer provides this require statement so that the user only needs to copy and paste it into an html file.

**Loading Javascript**

There are many different kinds of import statements, leaving the developer to guess which one is the right one.

```
import 'flatpickr';
import flatpickr from 'flatpickr';
import * as flatpickr from 'flatpickr';
```

We would like to investigate whether it is possible to detect which import statement could be used to import the javascript portion of a package.

### 7.4 Backup and restore
Before modifying the aurelia.json file we can create a backup. Then, with `au import --revert` we can revert to the previous aurelia.json file.

# 8. Todo
1. See what services custom importers could need (and create custom importers for aurelia-i18n, aurelia-kendoui-bridge and aurelia-materialize-bridge)

2. Take a closer look at [these issues](https://github.com/monterey-framework/cli/issues/27)

# 9. Questions
1. Should the single file dependency configuration only be used when there no `resources` either? The path is calculated incorrectly since the `path` is a reference to a file and not to a directory.

2. Should we use `lib/ui.js` to log messages, or can we keep using `aurelia-logging`? The importer logs messages of type `info` and `debug`
3. Does the bundler support the system module format? A plugin can configure its `dist` directory to point to `dist/system`, and the importer will configure the dependency using the `dist/system` directory.
```
"jspm": {
   "directories": {
      "dist": "dist/system"
   }
}
```

### DIFF
--- a/lib/commands/import/command.js
+++ b/lib/commands/import/command.js
@@ -1,0 +1,26 @@
+"use strict";
+
+const ImportEngine = require('../../importer/import-engine'),
+      ArgumentParser = require('../install/package-argument-parser');
+
+module.exports = class {
+
+  static inject() { return [ImportEngine, ArgumentParser]; };
+
+  constructor(engine, argumentParser) {
+    this.engine = engine;
+    this.argumentParser = argumentParser;
+  }
+
+  execute(args) {
+    let packages = this.argumentParser.parse(args);
+
+    if (packages.length === 0) {
+      throw new Error('Expected atleast one package (au import <package>)');
+    }
+
+    return this.engine
+      .import(packages)
+      .catch(e => { console.log(e); });
+  }
+};
--- a/lib/commands/import/command.json
+++ b/lib/commands/import/command.json
@@ -1,0 +1,10 @@
+{
+  "name": "import",
+  "description": "Configures a package in aurelia.json",
+  "parameters": [
+    {
+      "name": "packages",
+      "description": "Package names, separated by spaces"
+    }
+  ]
+}
--- a/lib/commands/install/command.js
+++ b/lib/commands/install/command.js
@@ -1,7 +1,31 @@
 "use strict";
 
+const ImportCommand = require('../import/command'),
+      PackageInstaller = require('../../importer/package-installer'),
+      ArgumentParser = require('./package-argument-parser'),
+      logger = require('aurelia-logging').getLogger('Install');
+
 module.exports = class {
+
+  static inject() { return [PackageInstaller, ImportCommand, ArgumentParser]; };
+
+  constructor(packageInstaller, importCommand, argumentParser) {
+    this.packageInstaller = packageInstaller;
+    this.importCommand = importCommand;
+    this.argumentParser = argumentParser;
+  }
+
   execute(args) {
-    return new Promise((resolve, reject) => reject(new Error('Install not yet implemented.')))
+    let packages = new ArgumentParser().parse(args);
+
+    if (packages.length === 0) {
+      throw new Error('Expected atleast one package (au import <package>)');
+    }
+
+    return this.packageInstaller
+      .install(packages)
+      .then(() => logger.info('The packages were successfully installed. Going to import them now.'))
+      .then(() => this.importCommand.execute(args))
+      .catch(e => console.log(e));
   }
 }
--- a/lib/commands/install/command.json
+++ b/lib/commands/install/command.json
@@ -1,0 +1,10 @@
+{
+  "name": "install",
+  "description": "Installs and configures a package in aurelia.json",
+  "parameters": [
+    {
+      "name": "packages",
+      "description": "Package names, separated by spaces. Versioning is supported (jquery@3.0.0)"
+    }
+  ]
+}
--- a/lib/importer/strategies/custom-importer.js
+++ b/lib/importer/strategies/custom-importer.js
@@ -1,0 +1,61 @@
+"use strict";
+
+const fs = require('../../file-system'),
+      Container = require('aurelia-dependency-injection').Container,
+      logger = require('aurelia-logging').getLogger('Custom-importer');
+
+let CustomImporterStrategy = class {
+
+  static inject() { return [Container, 'package']; };
+
+  constructor(container, pkg) {
+    this.container = container;
+    this.package = pkg;
+  }
+
+  applies() {
+    let location = this.getStrategyLocation();
+
+    return fs.exists(location)
+    .then(available => {
+      if (available) {
+        let strategy = this.getStrategy();
+        let result = strategy.applies();
+
+        return result;
+      }
+
+      logger.debug(`The plugin does not have a custom importer module. Looked for "${location}"`);
+
+      return false;
+    });
+  }
+
+  execute() {
+    let strategy = this.getStrategy();
+    let result = strategy.execute();
+
+    return result;
+  }
+
+  getStrategy() {
+    let ctor = require(this.getStrategyLocation());
+    return this.container.get(ctor);
+  }
+
+  getStrategyLocation() {
+    let pjson = this.package.packageJSON;
+
+    if (pjson.aurelia && (typeof pjson.aurelia.import === 'string')) {
+      return fs.resolve(fs.join(this.package.rootPath, pjson.aurelia.import));
+    }
+
+    return fs.resolve(fs.join(this.package.rootPath, 'install', 'importer-callbacks.js'));
+  }
+
+  get name() {
+    return 'Custom Importer Strategy';
+  }
+};
+
+module.exports = CustomImporterStrategy;
--- a/lib/importer/strategy-loader.js
+++ b/lib/importer/strategy-loader.js
@@ -1,0 +1,38 @@
+"use strict";
+
+const fs = require('../file-system'),
+      Container = require('aurelia-dependency-injection').Container;
+
+let StrategyLoader = class {
+
+  static inject() { return [Container] };
+
+  constructor(container) {
+    this.container = container;
+  }
+
+  getStrategies() {
+    let files = [
+      'metadata',
+      'custom-importer',
+      'aurelia-registry',
+      'jspm-section',
+      'browser-section',
+      'amodro'
+    ];
+    let strategies = [];
+
+    for(let file of files) {
+      let path = fs.join(__dirname, 'strategies', file);
+      let ctor = require(path);
+      let strategy = this.container.get(ctor);
+
+      strategies.push(strategy);
+    }
+
+    return strategies;
+  }
+};
+
+
+module.exports = StrategyLoader;
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -5,6 +5,8 @@ const fs = require('./file-system');
 const ui = require('./ui');
 const Project = require('./project').Project;
 const CLIOptions = require('./cli-options').CLIOptions;
+const LogManager = require('aurelia-logging');
+const Logger = require('./logger').Logger;
 
 exports.CLI = class {
   constructor(options) {
@@ -27,7 +29,18 @@ exports.CLI = class {
         }
 
         return this.createCommand(command, args);
-      }).then(command => command.execute(args));
+      })
+      .then((command) => {
+        this.configureLogger();
+
+        return command.execute(args);
+      });
+  }
+
+  configureLogger() {
+    LogManager.addAppender(this.container.get(Logger));
+    let level = CLIOptions.hasFlag('debug') ? LogManager.logLevel.debug : LogManager.logLevel.info;
+    LogManager.setLevel(level);
   }
 
   configureContainer() {
--- a/lib/commands/help/command.js
+++ b/lib/commands/help/command.js
@@ -35,6 +35,8 @@ module.exports = class {
   getLocalCommandText() {
     let commands = [
       require('../generate/command.json'),
+      require('../import/command.json'),
+      require('../install/command.json'),
       require('./command.json')
     ];
 
--- a/lib/commands/install/package-argument-parser.js
+++ b/lib/commands/install/package-argument-parser.js
@@ -1,0 +1,80 @@
+"use strict";
+
+const gitRegex = /(?:git|ssh|https?|git@[\w\.]+):(?:\/\/)?([\w\.@:\/~_-]+)\.git(?:\/?|\#[\d\w\.\-_]+?)$/;
+
+class PackageArgumentParser {
+
+  parse(args) {
+    let packages = [];
+    let flag = false;
+
+    for(let i = 0; i < args.length; i++) {
+      let arg = args[i];
+
+      if (flag) {
+        flag = false;
+        continue;
+      }
+
+      // skip arguments starting with - or --
+      if (arg.startsWith('--')) continue;
+      if (arg.match(/^-.$/)) {
+        flag = true;  
+        continue;
+      }
+
+      let pkg;
+
+      if (this.isGitUrl(arg)) {
+        pkg = this.parseGitUrl(arg);
+      } else if (this.hasVersion(arg)) {
+        pkg = this.parseVersioned(arg);
+      } else {
+        pkg = {
+          argument: arg,
+          name: arg  
+        };
+      }
+      packages.push(pkg);
+    }
+
+    return packages;
+  }
+
+  parseVersioned(argument) {
+    let split = argument.split('@');
+    return {
+      argument: argument,
+      name: split[0],
+      version: split[1]
+    };
+  }
+
+  hasVersion(argument) {
+    return argument.indexOf('@') > -1;
+  }
+
+  isGitUrl(argument) {
+    return argument.match(gitRegex);
+  }
+
+  parseGitUrl(argument) {
+    let matches = argument.match(gitRegex);
+
+    // github.com/aurelia/i18n
+    let segment = matches[1];
+
+    // aurelia/i18n
+    let repo = segment.substring(segment.indexOf('/') + 1);
+
+    // aurelia-i18n
+    let name = repo.replace('/', '-');
+
+    return {
+      argument: argument,
+      name: name
+    };
+  }
+}
+
+module.exports = PackageArgumentParser;
--- a/lib/commands/new/project-template.js
+++ b/lib/commands/new/project-template.js
@@ -1,6 +1,6 @@
 "use strict";
 const ProjectItem = require('../../project-item').ProjectItem;
-const NPM = require('../../npm').NPM;
+const NPM = require('../../package-managers/npm').NPM;
 const path = require('path');
 const string = require('../../string');
 const getSupportedVersion = require('../../dependencies').getSupportedVersion;
--- a/lib/file-system.js
+++ b/lib/file-system.js
@@ -3,8 +3,14 @@ const fs = require('fs');
 const nodePath = require('path');
 const mkdirp = require('./mkdirp').mkdirp;
 
+exports.fs = fs;
+
 exports.exists = function(path) {
   return new Promise((resolve, reject) => fs.exists(path, resolve));
+};
+
+exports.existsSync = function(path) {
+  return fs.existsSync(path);
 };
 
 exports.mkdir = function(path) {
@@ -48,6 +54,22 @@ exports.readFileSync = fs.readFileSync;
 exports.readFileSync = function(path, encoding) {
   return fs.readFileSync(path, encoding || 'utf8');
 };
+
+exports.copySync = function(sourceFile, targetFile) {
+  fs.writeFileSync(targetFile, fs.readFileSync(sourceFile));
+};
+
+exports.resolve = function(path) {
+  return nodePath.resolve(path);
+};
+
+exports.join = function(...segments) {
+  return nodePath.join.apply(this, segments);
+};
+
+exports.statSync = function (path) {
+  return fs.statSync(path);
+}
 
 exports.writeFile = function(path, content, encoding) {
   return new Promise((resolve, reject) => {
--- a/lib/importer/import-engine.js
+++ b/lib/importer/import-engine.js
@@ -1,0 +1,74 @@
+"use strict";
+
+const fs = require('../file-system'),
+      UI = require('../ui').UI,
+      CLIOptions = require('../cli-options').CLIOptions,
+      Container = require('aurelia-dependency-injection').Container,
+      StrategyLoader = require('./strategy-loader'),
+      Tutorial = require('./services/tutorial'),
+      MetadataService = require('./services/metadata-service'),
+      ResourceInclusion = require('./services/resource-inclusion'),
+      Registry = require('./services/registry'),
+      logger = require('aurelia-logging').getLogger('Importer'),
+      Package = require('./package'),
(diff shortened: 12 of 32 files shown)


EisenbergEffect:
@JeroenVinke I'm very excited about this. Thanks for putting in the awesome work. I'll need a bit of time to review this before we merge. Please hang in there. Thanks!

JeroenVinke:
Sure! We are going to create a couple of custom importers so that we know what services a custom importer might need. We will let you know when that work is done.

PWKad:
Just started looking in to this, awesome work @JeroenVinke and team!

EisenbergEffect:
Can you explain the need to add a dependency on bluebird?

EisenbergEffect:
For logging, we may want to use an implementation of the log appends that pushes the data to the CLI's ui module, for consistency. The idea is that you could run the CLI hosted and then you might have a different implementation of the ui module. So, I'd prefer anything that's intended to be displayed to go through that.

EisenbergEffect:
The bundler doesn't support the system module format at present. It only supports AMD and CommonJS.

EisenbergEffect:
Yes, the single file dependency configuration should probably be used when there are no resources.

lib/commands/import/command.js
```diff
@@ -0,0 +1,39 @@
...
+  static inject() { return [Container, UI, CLIOptions]; };
+
+  constructor(container, ui, options) {
+    this.container = container;
+    this.ui = ui;
+    this.options = options;
+  }
+
+  execute(args) {
+    let analyzer = this.container.get(Analyzer);
```
EisenbergEffect: Is there some reason why this can't be injected?

lib/commands/import/command.js
```diff
@@ -0,0 +1,39 @@
...
+    return analyzer
+      .setPackage(args[0])
+      // analyze contextual information (CLI parameters, given package)
+      .execute()
+    //*/
+      // configure aurelia.json, install custom tasks, run additional scripts
+      .then(() => {
+        let engine = new ImportEngine(
+          analyzer.result.project,
```
EisenbergEffect: Is there some reason why the result isn't passed through the promise instead of making it a property on the analyzer instance?

lib/commands/import/command.js
```diff
@@ -0,0 +1,39 @@
...
+          analyzer.result.config,
+          [new ImportDefault(), analyzer.result.importer]
```
EisenbergEffect: Should the `ImportDefault` instance come from the container instead?

lib/commands/import/command.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "import",
+  "description": "Generates code for common use cases in Aurelia applications.",
+  "parameters": [
+    {
+      "name": "generator-name",
+      "description": "The name of the generator you want to run."
+    }
+  ]
+}
```
EisenbergEffect: We need to fix up the description and parameter info here so that it matches the command. This is what gets displayed when the user asks for help.

lib/commands/install/command.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "import",
+  "description": "Generates code for common use cases in Aurelia applications.",
+  "parameters": [
+    {
+      "name": "generator-name",
+      "description": "The name of the generator you want to run."
+    }
+  ]
+}
```
EisenbergEffect: We need to fix up the description and parameter info here so that it matches the command. This is what gets displayed when the user asks for help.

lib/commands/install/command.js
```diff
@@ -1,7 +1,24 @@
...
+  static inject() { return [Container]; };
+
+  constructor(container) {
+    this.container = container;
+  }
+
   execute(args) {
-    return new Promise((resolve, reject) => reject(new Error('Install not yet implemented.')))
+    let npm = this.container.get(NPM);
+    let importCommand = this.container.get(ImportCommand);
```
EisenbergEffect: Is there some reason we can't inject the `ImportCommand`?

EisenbergEffect: Should this be hard-coded to NPM?

lib/importer/import-default.js
```diff
@@ -0,0 +1,287 @@
...
+ * 1. applyPatches
+ * 2. registerDependencies
+ * 3. registerBundles
+ * 4. saveProject
+ * 5. installTasks
+ * 6. executeScripts
+ */
+module.exports = class {
+
+    constructor() { }
```
EisenbergEffect: No need for the empty constructor.

lib/importer/import-default.js
```diff
@@ -0,0 +1,287 @@
...
+        this.cliParams = options;
+        this.projectInfo = projectInfo;
+        this.ui = this.engine.cliBridge.ui;
+
+        this.bundles = this.engine.projectConfig.build.bundles;
+        if (this.bundles.length === 0) {
+            throw new Error("aurelia.json: bundles section is missing.");
+        }
+
+        this.bundleName = this.cliParams.bundle || 'vendor-bundle.js';
```
EisenbergEffect: I think what we'll want to do here is search for the bundle with the largest number of dependencies and make that the fallback bundle.

lib/importer/import-default.js
```diff
@@ -0,0 +1,287 @@
...
+     * Register dependencies within aurelia.json bundle section
+     * @return ImportStepState | void
+     */
+    registerDependencies() {
+        if (!this.bundle) {
+            this.writeLine(`[INFO] Bundle '${this.bundleName}' could not be found. Looking for default bundles...`);
+
+            // There are 2 sections by default, second is usually the vendor-bundle.js
+            // Although, some developers prefer to merge everything into a single bundle
+            let index = this.bundles.length > 1 ? 1 : 0;
```
EisenbergEffect: See note above. If there's more than one bundle, and one is not specified, the lets use the bundle with the largest number of dependencies.

lib/importer/strategy-loader.js
```diff
@@ -0,0 +1,44 @@
...
+    for (let filePath of filePaths) {
+      let baseName = path.basename(filePath, '.js'),
+          StrategyType = require(`${this.basePath}/${baseName}`);
+      this.strategies.push(new StrategyType());
+    }
+  }
+
+  load() {
+    for (let strategy of this.strategies) {
+      let chosen = strategy.determine();
```
EisenbergEffect: I'm not sure if I like the method name `determine`. What about `canExecute()` or `applies`?

EisenbergEffect:
Oops. Looks like I may have not reviewed the latest set! It was a bit confusing with all the commits. Regardless, there are probably still some applicable items in my comments and questions above. If you can address the ones that make sense, that would be great. I'll make another review later today or tomorrow.

JeroenVinke:
Will do, thanks. I can also squash all commits so it's easier to review

EisenbergEffect:
That would be great! Thanks!

JeroenVinke:
I think that the helper methods in [this class](https://github.com/monterey-framework/cli/blob/de3f250ff7a2e5b1b4b2c0a9113f35bf0ec5655d/lib/importer/services/aurelia-json-patcher.js#L16) can also be added to [project.js](https://github.com/monterey-framework/cli/blob/de3f250ff7a2e5b1b4b2c0a9113f35bf0ec5655d/lib/project.js).

What I had in mind when creating aurelia-json-patcher was to have one class responsible for changing the aurelia.json, and I didn't want all parts of the importer to have deep understanding of the structure of aurelia.json. What do you think?

lib/importer/strategy-loader.js
JeroenVinke: `applies` sounds good, will change it tomorrow

EisenbergEffect:
@JeroenVinke That makes sense. Feel free to update things to move those methods over to the project class. Thanks!

JeroenVinke:
Ok, resolved the issues and squashed the commits. I also made a custom importer locally for the aurelia-kendoui-bridge to see what that would look like: http://screencast.com/t/2oQUXr1Olpn

The code of this custom importer can be found [here](https://gist.github.com/JeroenVinke/e6fdef787becae606bab110028ea65b2). 

@zewa666 did you once create a custom importer for aurelia-i18n? If so, where can I find it? I'd like to see what services these custom importers might need

Any tips on how to improve the output? you're getting a lot of messages (even without `--debug`)

lib/importer/strategies/custom-importer.js
```diff
@@ -0,0 +1,55 @@
...
+    return result;
+  }
+
+  getStrategy() {
+    let ctor = require(this.getStrategyLocation());
+    return this.container.get(ctor);
+  }
+
+  getStrategyLocation() {
+    return fs.resolve(fs.join(this.package.rootPath, 'importer-callbacks.js'));
```
JeroenVinke: I don't like the name `importer-callbacks.js`. Currently the importer looks for `node_modules/<package>/importer-callback.js` and if that file is present it executes it as a import strategy (calling `applies()` first, then `execute()` if `applies()` returned `true`)

adriatic: why do you not like this name and what would you like instead?

JeroenVinke: I must have missed this e-mail notification. I don't quite like the `-callbacks` part. Perhaps something like `import-handler.js`. Also, I wonder if we should reference this file from package.json so that plugin authors can decide how to name the file (and where it is located)

zewa666:
@JeroenVinke I'm not sure if I'm sure what you're refering to, but if its the work that was related to aurelia-cli-pacman then you can find that in the [auto-install](https://github.com/aurelia/i18n/commits/auto-install) branch on i18n. If thats not what you're looking for please just ping me on gitter so we can address that quicker

JeroenVinke:
Together with @zewa666 we came up with some improvements:

**Installing specific branches**
I have added support for git urls to the `au install` command: `au install git://github.com/aurelia/i18n.git#auto-install`.

**Customize location of importer-callbacks.js file**
The importer now looks in the package.json of the package for the `"aurelia"."import"` key to determine where to look for a custom importer of the package.

If this is a string (a path to a .js file) it will look for the custom importer there:
```
{
  "aurelia": {
    "import": "./custom-dir/aurelia-importer.js"
  }
}
```

If `"aurelia"."import"` does not exist, the importer will look for a custom importer at `install/importer-callbacks.js`.

If `"aurelia"."import"` is an object, it will execute the import instructions
```
{
  "aurelia": {
    "import": {
      "bundles": [{
        "name": "aurelia-kendoui-bridge",
        "path": "../node_modules/aurelia-kendoui-bridge",
        "main": "index.js"
      }]
    }
  }
}
```

**Generate importer-callbacks.js file**
In order to help plugin authors get started quickly with creating a custom importer, @zewa666 came up with the idea to generate a custom-importer.js template. We could create a yeoman generator, but perhaps we can leverage aurelia-cli's generator functionality instead.


**Services**
A couple of ideas for services that would be useful for custom importers:
- patching main.js / main.ts (for adding `aurelia.plugin('my-plugin')`)
- patching html files (adding <require> statements to load .css files)

---------------------------

Should I squash the latest commits so you can do another review? Also, should we create the yeoman generator and additional services now or can we do this in a subsequent release?

EisenbergEffect:
Let's go ahead and squash the commits. Also, we can create the generator and related services at a later time after we get this merged and released.

JeroenVinke:
Ok, I have squashed the commits

zewa666:
Great work, looking forward to this one. I'll make sure that the i18n plugin is ready for prime time

EisenbergEffect:
@JeroenVinke Just wanted to let you know this isn't being ignored. It's a large bit of work, so it will take a bit of time to review. Do you feel good that it's all ready to go now?

JeroenVinke:
Yes, I believe it is good to go. We are also available to fix any bugs that may pop up and work on extending and improving the importer.

EisenbergEffect:
BOOYA! 🥇

adriatic:
A new era begins now!

zewa666:
Congratulations to all involved in this. @JeroenVinke I'd recommend as soon as the new CLI version is published we should setup i18n based on the gist you've sent out.

JeroenVinke:
👍

AStoker:
First off, this is _awesome_!
Question though, perhaps I missed it in the comments. How would this handle scenarios where the use has multiple bundles, or bundles that are named differently? Does it automatically insert the configuration into the `aurelia.json`, or would it try to figure it out on its own and then ask for clarification?

JeroenVinke:
We are very excited as well! When you use `au install` or `au import` you can supply a `-b` or `--bundle` flag. If you do then it will use that bundle, if you do not it will use the "default" bundle, which currently means that it will pick the largest bundle and it will use that.

AStoker:
Very handy. Do we have documentation describing all this? Sounds awesome.  

Also, would it make more sense to use the `loader.configTarget` bundle instead of analyzing which bundle is larger? We break up our bundles to try and be pretty granular, so we're not loading up bundles that aren't needed for particular parts of the application. That results in different size bundles. 

If the default just goes to the largest, there's nothing "static" about where the packages would go, and could cause some confusion about where things are. If on Monday I add a bundle, and it configures it in the `vendor-bundle`, but by Wednesday my `settings-vendor-bundle` is larger, and my newly imported packages goes there, there was nothing that indicated to me that the install location changed.  

Just some thoughts. Not a deal breaker, but things that potentially should be resolved.

adriatic:
@AStoker  As you know, this addition to Aurelia-CLI was not even released yet, so whatever functionality is implemented at the moment is the result of discussions within the Aurelia-UI-Toolkits and Monterey Framework members. I am pointing it out to indicate the the initial auto-configuration features set is biased towards what our team needs for our bridges (KendoUI and Materialize) and for [Monterey](https://github.com/monterey-framework) application.

I am expecting that this `package importer` component will remain "owned" by our group (Rob has the final say of course), so we plan to create a repository with samples and articles on how to use it. In addition, we will host the discussion on improvement ideas - where you may just made the first contribution.

EisenbergEffect:
@adriatic I'm happy to let you all own this inside our tool. I was pondering whether we should have a bit of a "foreign exchange program" by bringing @JeroenVinke onto the Aurelia Core Team so that he can be a member of both groups. That would enable him greater ability to evolve this feature over time in response to needs of both the broad Aurelia community as well as the communities around Monterey and UI Bridges.

adriatic:
@EisenbergEffect I am not surprised to again see a complete alignment of concepts which we did not discuss. I am very happy to see Jeroen's ascent (being a member of the core team is a big compliment in my view) and know for a fact that he sees it the same way. 

I treat Jeroen as my son that I never had (proud father of two daughters) so his career is **very** important to me.

JeroenVinke:
@AStoker 

> If the default just goes to the largest, there's nothing "static" about where the packages would go, and could cause some confusion about where things are. If on Monday I add a bundle, and it configures it in the vendor-bundle, but by Wednesday my settings-vendor-bundle is larger, and my newly imported packages goes there, there was nothing that indicated to me that the install location changed.

Good point, that would be unwanted behaviour

> Very handy. Do we have documentation describing all this? Sounds awesome.

Nothing more than what's described [here](https://github.com/aurelia/cli/pull/454#issue-198239875). I can  get started on properly documenting this. We could have two parts, one for the user of the importer and another for plugin authors

AStoker:
@JeroenVinke, in as far as documentation, sounds good. Since this is merged (and therefore will be released in the next cycle), it would be helpful to at least open up an issue for documentation of this feature, just to cover it for later. This feature is awesome, and I think will help out users greatly, but they're going to need to know how to use it, and they won't find that in the issue backlog.  

Thanks again for your work!

adriatic:
@AStoker @JeroenVinke We (aureliatools group) are very motivated to make the cli importer an example of using  world class documentation and discussion tools - as Monterey's success largely depends on that tool. So, I am on the process of building the [dedicated forum](http://forum.aureliatools.com) where we can discuss all needed elements of the importer documentation set and use.

As you could easily guess, this forum will host discussions other than cli importer - it will equally support our other projects

PR #457, opened 2017-01-05
Removed Typings dependency in favor of @types

milkshakeuk:
extended from #397


glenkleidon:
Hi MilkshakeUK/Con-ssc,  
Do you think that https://github.com/aurelia/framework/issues/683 might be implicated here?  I have the fix,  just wondering If your pull request is suitable?  When are you planning to merge back or have I missed it?

milkshakeuk:
@glenkleidon I'm not sure I understand what you are asking, this pull request is intended to replace the use of the [typings cli](https://github.com/typings/typings) in favour of TypeScripts now [prescribed consumption](http://www.typescriptlang.org/docs/handbook/declaration-files/consumption.html) of declaration files (as of typescript 2.0).

This should fix aurelia/framework#683 since it appears to be caused by the typings cli.

glenkleidon:
Thanks,

 

Yes, that answers the question.  Rob has closed.

 

Glen.

PR #469, opened 2017-01-24
feat(bundle): add support for 'append' property in bundle config

haynesjm42:
Just a copy of the 'prepend' logic to allow appending files to the end of a bundle. Useful for legacy (non-amd) Javascript library plugins which require the global object defined in the 'main' library script to be available when the plugin scripts execute.


haynesjm42:
I signed the CLA ... though the e-mail on the commit was not yet added to my github account at the time I submitted this PR, but the check didn't seem to refresh after I added it ... do I need to re-submit the PR?

PR #471, opened 2017-01-25
feat(bundle): add support for 'append' property in bundle config

haynesjm42:
Just a copy of the 'prepend' logic to allow appending files to the end of a bundle. Useful for legacy (non-amd) Javascript library plugins which require the global object defined in the 'main' library script to be available when the plugin scripts execute.

### DIFF
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -13,6 +13,7 @@ exports.Bundle = class {
     this.config = config;
     this.dependencies = [];
     this.prepend = (config.prepend || []).filter(x => bundler.itemIncludedInBuild(x));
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
Makes sense :)

AStoker:
Do we have some documentation for this feature? Would be handy :)

haynesjm42:
Created pr [#716](https://github.com/aurelia/framework/pull/716) in [aurelia/framework](https://github.com/aurelia/framework) with documentation for the append feature. I didn't include a specific example since it functions exactly the same as prepend, and I noted as such in the documentation. Thanks!

PR #474, opened 2017-01-30
Changed transpilation task for gulp-ts >= 3

con-ssc:
Used the example from the gulp-typescript page for ts 2.x


EisenbergEffect:
Thanks!

con-ssc:
That was unexpectedly unproblematic.

PR #476, opened 2017-01-31
fix(copy-files): copy files when building for production

fabioluz:
This PR resolves the problems found in https://github.com/aurelia/cli/pull/415

## Problems

**Thomas Darling wrote:**
> @fabioluz @AStoker @EisenbergEffect I have a couple of concerns about this.
> 
> > By default root is the src folder and the output is the scripts folder.
>
> I don't understand why the output is relative to the scripts folder.
> I'd very strongly prefer to have it be relative to the project output folder instead, so I don't have to write ../ when copying files that are not scripts - I think copying non-script files will be a much more common scenario for this, such as in the example above, which actually copies font files.
> 
> I'm also wondering why this task runs as part of the serve task - that's for development only, so this won't work on a build server. From my point of view, copying additional files should just be part of the build, and those files should be watched for changes, just like any other source file.
> 
> Thoughts?

**Rob Eisenberg wrote:**

> I agree with all points. Would you be willing to submit a PR to make those
> changes?

## Changes

The `input` and `output` are now relative to project folder. The setup for copying files is:

```
"bundles": [ ... ], 
"copyFiles": {
    FILE_YOU_WANT_TO_COPY_BASED_ON_PROJECT_FOLDER: DESTINATION_BASED_ON_PROJECT_FOLDER
  }
```

An example for bootstrap would be:

```
"bundles": [ ... ], 
"copyFiles": {
    "node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.woff2": "bootstrap/fonts".
    "node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.woff": "bootstrap/fonts",
    "node_modules/bootstrap/dist/fonts/glyphicons-halflings-regular.ttf": "bootstrap/fonts"
  }
```

The `copyFiles()` is now part of the `build()`, solving problem that files are currently copied only in the development process. 

The files are being watched for changes. The `gulp-changed-in-place` ensures that only changed files will be recopied.

----

I tried to follow the same strategy in the other tasks (transpile, process-markup, etc). If something is not right, please tell me.

I'll update the docs https://github.com/aurelia/framework/pull/666 when I have time.

### DIFF
--- a/lib/commands/new/project-template.js
+++ b/lib/commands/new/project-template.js
@@ -117,7 +117,6 @@ exports.ProjectTemplate = class {
       ProjectItem.resource('build.ext', 'tasks/build.ext', this.model.transpiler),
       ProjectItem.resource('build.json', 'tasks/build.json'),
       ProjectItem.resource('copy-files.ext', 'tasks/copy-files.ext', this.model.transpiler),
-      ProjectItem.resource('copy-files.json', 'tasks/copy-files.json'),
       ProjectItem.resource('run.ext', 'tasks/run.ext', this.model.transpiler),
       ProjectItem.resource('run.json', 'tasks/run.json')
     ).addToGenerators(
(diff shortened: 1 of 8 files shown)


EisenbergEffect:
Thanks!

Issue #330, opened 2016-09-22
[Feature] CLI Test Coverage


AStoker:
**I'm submitting a feature request**
- **Library Version:**
  0.19.0

**Please tell us about your environment:**
- **Operating System:**
  All
- **Node Version:**
  6.6.0
  
  
- **NPM Version:**
  3.10.3
  
  
- **Browser:**
  all
- **Language:**
  all

**Current behavior:**
No test coverage of Aurelia CLI behaviors and functions

**Expected/desired behavior:**

- **What is the expected behavior?**
  Test coverage should exist for the CLI to protect from regression issues and ensure good code quality
- **What is the motivation / use case for changing the behavior?**
  Adding new features to the CLI can cause regression issues, among other things. To protect against regression issues, and to ensure that the CLI is behaving as expected, tests should exist.

AStoker:
Thoughts from: https://github.com/aurelia/cli/pull/326

>My initial thoughts for some basic tests would be to have a few difference "scenarios" represented by various aurelia.json files. From that, we might be able to at least do some basic test cases to make sure that functions will execute correctly given various scenarios, and are returning the expected data.  
https://github.com/aurelia/cli/pull/326#issuecomment-248945529

>That makes sense. Ultimately you end up building a little mini-framework for testing, based on scenario definitions. The trick is the assertions I suppose. Lets just keep brainstorming about this. In the mean time, I'll get this merged in. Thanks for working on it!
https://github.com/aurelia/cli/pull/326#issuecomment-248947941

MeirionHughes:
this should be priority, unfortunately I couldn't mock `fs` properly and a few things broke (namely the tracing lib that is embedded in this repo).

EisenbergEffect:
We could refactor that to enable instancing a file system implementation if that would help. Let me know.

MeirionHughes:
if all the libs and utils could be made to use an instance of the FileSystem class, rather than use `require("fs")`, that probably would do it. could then put in a mock file system during testing.

s-hoff:
[...]

PR #481, opened 2017-02-03
chore(unit-tests): Enable unit testing

s-hoff:
Added a basic setup for unit testing using jasmine and mock-fs. I can drop the test:watch task and nodemon devDependency if desired.
fs.exists() is deprecated and not mocked by mock-fs. Therefore, I had to remove its usage in order to make everything testable.
I only included tests for the file-system module (as a proof of concept) and the functions or part of the functions changed for fs.exists() as I am not very experienced with the test themselves and therefore wanted to avoid testing the whole thing and having to rewrite it all ;). I would happily contribute more tests myself once this is thoroughly reviewed and merged.
Refers to #330.


EisenbergEffect:
@AStoker Can you take a look at this? It would be great to have testing in place.

s-hoff:
I forgot to mention:
  Since the esprima dependency is not in the package.json yet, one has to first run `npm i esprima`in the cli directory. Then to run the tests do `npm test`or `npm run test:watch` to watch the files during development.

AStoker:
Just starting to take a crack at this. I've been out of the loop for a bit. Anything in particular stand out that I should know, @s-hoff? Thanks for setting this up!

AStoker:
Also, if you could squash your changes to a single commit, that would be helpful.

lib/file-system.js
```diff
@@ -7,6 +7,15 @@ exports.exists = function(path) {
   return new Promise((resolve, reject) => fs.exists(path, resolve));
 };
 
+exports.stat = function(path) {
```
s-hoff: I am considering to drop a deprecation warning in here and remove completely later (as some new functions on the master branch that are not present here still use it).

spec/lib/cli.spec.js
```diff
@@ -0,0 +1,166 @@
...
+        .then(project => {
+          expect(Project.establish)
+            .toHaveBeenCalledWith(cli.ui, path.join(process.cwd()));
+          expect(project).toBe(project);
+        })
+        .catch(fail).then(done);
+    });
+
+    it('does not catch Project.establish()', () => {
+      pending('Does it even make sense to test this?');
```
s-hoff: As it says. The test would refer to the use of project.establish, returning a promise. If it were catched here, it could not be observed with a .catch() later. However, I feel like the effort of testing for these kind of things might be over the top compared to the benefits.

spec/lib/file-system.spec.js
```diff
@@ -0,0 +1,164 @@
...
+
+  describe('The mkdirp() function', () => {
+    it('makes deep directories', done => {
+      fs.mkdirp(writeDir + readDir).then(() => {
+        return fs.readdir(writeDir + readDir);
+      }).catch(fail).then(done);
+    });
+
+    it('rejects if mkdirp returns an error', done => {
+      pending('figure out how to test this.');
```
s-hoff: I was intending to mock mkdirp at first but realized that because the function is imported to be directly accessed, I was unable to do so. Doing only a require('mkdirp'); on the top and calling mkdirp.mkdirp() in the fs.mkdirp function should make this possible.

s-hoff:
Generally, I would like to have it ensured that I did not misunderstand how some of the functions I changed worked before as it is now catching for errors on fs.stat() rather then checking for true/false in the resolve.

s-hoff:
I resolved the conflict and squashed the commits. The failing tests were due to the polyfills missing because the tests do not run through the normal startup code that would load them normally. They are now loaded for all tests by a helper.

lib/file-system.js
AStoker: That sounds like a good idea, since stat is going away from `fs` anyway, right?

spec/lib/cli.spec.js
AStoker: I'd say that if you've already written the test, then it's fine. But for a first draft of tests, you might not have to get into the weeds that much.

s-hoff: Turns out there was a super easy way by faking Project.establish that I just overlooked before.

lib/file-system.js
s-hoff: Actually, `exists` is going away, and stat is one of the replacements ;).
Would a comment suffice here, or should I go all the way and include the logger to drop it on the console?

s-hoff:
Interesting, I cant reproduce the error here on my system. Does it really say nothing about an/which test it is related to?

JeroenVinke:
@s-hoff for what it's worth, here is the output when I run `npm test`: https://gist.github.com/JeroenVinke/1590e839760a8dfa96be1c9e251d7817

node: v7.4.0
npm: 4.0.5

Great work on this, can't wait to add some tests 😄

AStoker:
@s-hoff, I'm getting the same errors as @JeroenVinke here. Maybe you have some globally installed package on your machine that isn't being copied over? Have you tried running your tests on a separate machine?

JeroenVinke:
[These changes](https://github.com/JeroenVinke/cli/commit/b5797cd1868ea4b6120436ecf709be56b4f38c8e) fixed the errors for me. The output is now:

> PS C:\Development\cli> npm test

> aurelia-cli@0.26.1 test C:\Development\cli
> jasmine

>  Started
> .......................*.............

> Pending:

> 1) The file-system module The mkdirp() function rejects if mkdirp returns an error
  > figure out how to test this.

> 37 specs, 0 failures, 1 pending spec
Finished in 0.144 seconds

The one pending test is this one:

```
    it('rejects if mkdirp returns an error', done => {
      pending('figure out how to test this.');
    });
```

so that is expected

s-hoff:
I just tested it on a pretty much fresh Ubuntu vm and got the errors but not the failing tests. The changes to the tests by @JeroenVinke fixed that. I will diff the npm packages to my main machine tomorrow.
On changing to `path.posix`: It might just be me, but it feels a little wrong to have to change these just to statisfy the unit tests. So unless we want to always use posix paths anyway, an alternative might be to use `path.join()`in the tests to build the comparision strings as well rather then typing it out directly. So `.toBe('aurelia_project/generators/tests.js')` would become `.toBe(path.join('aurelia_project', 'generators', 'test.js'))`.
We could implement the last pending test by changing the `mkdirp` usage in `file-system` to just `require('./mkdirp')` and later `mkdirp.mkdirp()`. That would allow for the function to be properly mocked. Another way might be to use something like `mock-require` or `mockery`.

JeroenVinke:
Glad you are able to reproduce the errors

> On changing to path.posix: It might just be me, but it feels a little wrong to have to change these just to statisfy the unit tests. So unless we want to always use posix paths anyway, an alternative might be to use path.join()in the tests to build the comparision strings as well rather then typing it out directly. So .toBe('aurelia_project/generators/tests.js') would become .toBe(path.join('aurelia_project', 'generators', 'test.js')).

Definitely

JeroenVinke:
@s-hoff I would really like to see this PR merged, so if the `mkdirp` is blocking, i'd say go ahead and remove that test for now. We can add this test after we merge the PR.

lib/project.js
```diff
@@ -73,12 +73,13 @@ exports.Project = class {
...
   resolveGenerator(name) {
     let potential = path.join(this.generatorDirectory, `${name}${this.model.transpiler.fileExtension}`);
-    return fs.exists(potential).then(result => result ? potential : null);
+    return fs.stat(potential).then(() => potential).catch(() => null);
   }
 
   resolveTask(name) {
     let potential = path.join(this.taskDirectory, `${name}${this.model.transpiler.fileExtension}`);
-    return fs.exists(potential).then(result => result ? potential : null);
+    console.log(potential);
```
JeroenVinke: can you remove this log statement?

lib/file-system.js
```diff
@@ -29,7 +38,7 @@ exports.mkdirp = function(path) {
       else resolve();
     });
   });
-}
+};
```
JeroenVinke: could you undo this change (and similar ones below)? a different PR fixes linting issues so we should try and prevent merge conflicts here

JeroenVinke:
Could you also squash the commits? Thanks!

s-hoff:
Done, done and done ;)

EisenbergEffect:
@JeroenVinke Can you do a final review?

JeroenVinke:
Looks great, thanks @s-hoff!

There will be a merge conflict in https://github.com/aurelia/cli/pull/550 when you merge this one, but I will resolve it after this one gets merged.

EisenbergEffect:
Thanks @JeroenVinke and @s-hoff 
Let me know when #550 is ready.

Issue #451, opened 2016-12-30
Testing changes during cli development


s-hoff:
**I'm submitting a feature request**

* **Library Version:**
0.23.0

**Please tell us about your environment:**
* **Operating System:**
Linux (Ubuntu)

* **Node Version:**
6.5.0

* **NPM Version:**
3.10.0

* **Browser:**
all

* **Language:**
all

**Current behavior:**
When making changes to the cli, in order to test it, one must either

1. Edit the node_modules/aurelia-cli folder in a test project, or
2. edit the files in the global installation when making changes that affect e.g. the new command.

When trying to do an `npm link` on the dev directory, npm will complain about esprima not being installed:

```
Error: Cannot find module 'esprima'
    at Function.Module._resolveFilename (module.js:455:15)
    at Function.Module._load (module.js:403:25)
    at Module.require (module.js:483:17)
    at require (internal/module.js:20:19)
    at /home/shoff/aurelia/aurelia/cli/lib/build/amodro-trace/lib/parse.js:5:62
    at Array.map (native)
    at define (/home/shoff/aurelia/aurelia/cli/lib/build/amodro-trace/lib/parse.js:5:36)
    at Object.<anonymous> (/home/shoff/aurelia/aurelia/cli/lib/build/amodro-trace/lib/parse.js:10:1)
    at Module._compile (module.js:556:32)
    at Object.Module._extensions..js (module.js:565:10)
```

**Expected/desired behavior:**

* **What is the expected behavior?**
Using `npm link`, `npm install ./cli` or `node bin/aurelia-cli.js`, the cli should work as normal, but with the local changes. Installing esprima as a dependeny or devDependency solves this.
For some reason, using `npm i -g aurelia-cli` does not produce the same error, even although esprima cannot be found anywhere in the global installation. Adding the dependency might in fact save from future trouble as it becomes clear where it comes from.
amodro-trace, which requires this module, has esprima in its dependencies. Therefore, I assume this would also be a good place for it in the cli.

* **What is the motivation / use case for changing the behavior?**
Testing changes to the cli should be easy and not involve having to potentially even edit the global installation manually.

s-hoff:
After this, it would also be good to add documentation on how to manually test ones changes to the readme or wiki, so new cli developers can quickly get set up.

JeroenVinke:
@s-hoff since your PR has been merged, can this been closed?

s-hoff:
[...]

PR #482, opened 2017-02-03
feat(local-testing): Added esprima to dependencies

s-hoff:
Refers to #451 
With esprima declared as a dependency, an `npm link` can be used to test local changes without having to run`npm i esprima` first.
I have added it to the dependencies instead of devDependencies because it is used during runtime, even although an `npm i -g aurelia-cli` seems to magically get it from somewhere and I feel like this is safer then having some sort of 'ghost dependency' going on.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
   "dependencies": {
     "aurelia-dependency-injection": "^1.0.0",
     "aurelia-polyfills": "^1.0.0",
+    "esprima": "^3.1.3",
     "npm": "^3.10.8"
   },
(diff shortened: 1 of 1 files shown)


AStoker:
Seems reasonable. Assuming we fix that conflicting file.

s-hoff:
I fixed the conflict - looks like both my npm install and that done for the package importer moved the "npm" dependency by a few lines in the array.

AStoker:
@EisenbergEffect, can you take a peek at this? Looks good to me.

PR #486, opened 2017-02-14
fix: serving pages with Browsersync for ASP.NET Core projects

mobilemancer:
Fixed Browsersync setup for ASP.NET Core based projects.
Related to aurelia/cli#304.


mobilemancer:
Can't believe I missed that, sorry :(
Added the missing JS version now.

JamesFenton:
@JeroenVinke to try to answer your question on #304, I think the reason for the proxy because the website is served using the asp.net server, running on port 5000. 

This issue is fixed with #537, and the proxy seems like a convenience, for example it removes the need to enable CORS on the server if the website is running on port 9000 (using browsersync) and the backend is running on port 5000 (using the asp.net server).

EisenbergEffect:
@JeroenVinke Can you do a final review on this?

JeroenVinke:
I have a few concerns. Adding another version of the run task makes it more difficult to maintain as a change to one of the run tasks would have to be copied to 3 other files. I also was not able to get this working, as browsersync did not want to serve the page (it kept loading), possibly because the backend was running on port 62608 instead of port `5000` which is in the `proxy` field of browersync.

I'd like to discuss alternative solutions. For example, we could document how to enable CORS in asp.net core.

mobilemancer:
If you aren't running the ASP.NET Core host on 5000, then it won't work, no. 5000 is just the "classical" port that the Core team has been using for self hosting.

Yeah adding extra files might be cumbersome, I was just under the influence that it was a prefered way to solve this issue.

Ideally I think the CLI should prompt for the hosting port during the setup, and modify the run task according to that, for .NET Core projects. But it should also be documented how this needs to be changed when moving hosting to other ports.

EisenbergEffect:
@mobilemancer @JeroenVinke What's our status on this?

JeroenVinke:
I'd prefer to add hints in our documentation on how to configure CORS on asp.net core.

mobilemancer:
If @JeroenVinke thinks this is adding too much added complexity I understand. However I think the CLI should be setting up the ASP.NET Core projects so that Browsersync just works. :)

JeroenVinke:
We do need to solve this issue one way or another, we definitely agree on the problem. Why I feel this should be fixed serverside is mainly because of two reasons:

- the CORS issue is generally resolved by configuring the backend
- when the user deploys the frontend+backend CORS needs to be configured anyway

If we can have the cli do this automatically then that would be best. But I think it takes manual action to configure cors

JamesFenton:
We have our backend and our front end in 2 separate Visual Studio projects. The front end runs on port 9000 (Browsersync) and the backend runs on a different port, eg 5000, during development. When we deploy, we take all of the static files from the front end project and put them into the wwwroot of the backend project so they run on the same port, so CORS is not needed. 

A simpler approach that we will try is have the Aurelia CLI output the bundles into the wwwroot of the backend project. I imagine we'll need to move the index.html files into the backend project and update the paths in `aurelia.json`

JeroenVinke:
Good point @JamesFenton . Then i'd be in favor of

> Ideally I think the CLI should prompt for the hosting port during the setup, and modify the run task according to that, for .NET Core projects

Ideally we would ask the user whether to use the proxy (and what port the API is running on), put the proxy setting in aurelia.json somewhere and pass that to browersync. That way we can possibly use the [existing](https://github.com/aurelia/cli/blob/master/lib/resources/tasks/run.js) run tasks. @mobilemancer do you have time to give this a shot?

JeroenVinke:
Perhaps we can read the `properties/launchSettings.json` file and use the port defined in there as default value:

```
{
  "iisSettings": {
    "windowsAuthentication": false,
    "anonymousAuthentication": true,
    "iisExpress": {
      "applicationUrl": "http://localhost:53540/",
      "sslPort": 0
    }
  },
  "profiles": {
    "IIS Express": {
      "commandName": "IISExpress",
      "launchBrowser": true,
      "launchUrl": "api/values",
      "environmentVariables": {
        "ASPNETCORE_ENVIRONMENT": "Development"
      }
    },
    "WebApplication1": {
      "commandName": "Project",
      "launchBrowser": true,
      "launchUrl": "api/values",
      "environmentVariables": {
        "ASPNETCORE_ENVIRONMENT": "Development"
      },
      "applicationUrl": "http://localhost:53541"
    }
  }
}
```

JeroenVinke:
Just to keep you up to date, I talked to @mobilemancer and we came up with an idea on how to improve the asp.net core projects, which we'll create a proposal for

JeroenVinke:
@mobilemancer once you finished writing the proposal, I suggest to discuss it in an issue. Thanks!

Issue #465, opened 2017-01-19
vendor-bundle-xxx.js `rev` not changed when app-bundle-xxx.js `rev` changed.


xiwc:
**I'm submitting a bug report**

* **Library Version:**
0.24.0

**Please tell us about your environment:**
* **Operating System:**
Linux 

* **Node Version:**
4.4.x

* **NPM Version:**
3.10.x

* **Browser:**
all 

* **Language:**
ESNext

**Current behavior:**
i have three bundle:
app-bundle-xxx.js
deps-bundle-xxx.js
vendor-bundle-xxx.js

vendor-bundle-xxx.js `rev` not changed when app-bundle-xxx.js `rev` changed.
but app-bundle-xxx.js is loading by vendor-bundle-xxx.js.

so browser can not load changed app-bundle-xxx.js as vendor-bundle-xxx.js  was cached in index.html.

**Expected/desired behavior:**

vendor-bundle-xxx.js `rev` changed too when app-bundle-xxx.js `rev` changed.

* **What is the expected behavior?**

* **What is the motivation / use case for changing the behavior?**

edwardridge:
This happens for me as well, but only on --watch: if I run a full au build it updates the vendor bundle correctly.

sbolel:
I think this is related to issues https://github.com/aurelia/cli/issues/472, https://github.com/aurelia/cli/issues/378, and PR https://github.com/aurelia/cli/pull/427. I think it should be resolved with the next release. 

@edwardridge what's the motivation for using `rev` in an environment that use with `au run`? It should only be necessary when building minified bundles to release to production.

---

Because of this though, my team decided to put everything into a single bundle and then generate a hash manually in the `aurelia_project` build task. We don't use the `rev` option for any environment in `aurelia.json` since that's handled by our manual task. We also changed `configTarget` in `aurelia.json` to a filename that doesn't exist because that somehow prevents the order of `define` statements from changing. Bundling things into separate files has no advantage over bundling into a single file -- if anything, the browser has to make more requests and spends more time to run the same code.

AStoker:
This still appears to be happening. In particular for us, we have multiple bundles, and it's our `settings-bundle` that is changing, and the `vendor-bundle` that continues to have the same revision hash.  
[...]

PR #489, opened 2017-02-16
fix(bundle): hash filename after loader config added

AStoker:
Resolves: https://github.com/aurelia/cli/issues/465
Previously, a file hash would be generated for the config target before
the loader config was added. This would cause cache busting issues
since a vendor bundle would not have any changes, and therefore have an
identical hash as previous, but other bundles would have a different
hash. The browser cached version of the vendor bundle would now
incorrectly point to a wrong file.
By moving the hashing to after the loader config is added, we now catch
the caching issue, and generate a new (correct) hash.

### DIFF
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -179,12 +179,6 @@ exports.Bundle = class {
       let contents = concat.content;
       let bundleFileName = this.config.name;
 
-      if (buildOptions.rev) {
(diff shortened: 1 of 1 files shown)


sbolel:
This is huge, thank you for doing this!

AStoker:
@EisenbergEffect, mind taking a peek at this?

Issue #461, opened 2017-01-16
Feature Request: Show notification if trace fails


PWKad:
Currently if you reference a file by bad path the trace fails and CLI aborts the watch process.  Add a notification to show that the trace imploded to prevent any issues when wondering why changes are not showing up.

Issue #408, opened 2016-11-11
[Bug] Watcher is crashed, when nonexistent file was added to custom element


Rebulus:
**I'm submitting a bug report**

* **Library Version:**
0.22.0

**Please tell us about your environment:**
* **Operating System:**
OSX 10.11.6

* **Node Version:**
4.4.2

* **NPM Version:**
3.10.9

* **Browser:**
all

* **Language:**
TypeScript 2.1.1

**Current behaviour:**

`au run --watch` is crashed, if I use `import` of a nonexistent file. For example:
I have the custom element:

```typescript
import {bindable, computedFrom} from 'aurelia-framework';
import './some-non-exist-file';

export class CustomButton {
  @bindable type: string;
  @bindable disabled: boolean;
  @bindable text: string;

  @computedFrom('type')
  get buttonType(): string {
    return this.type ? `btn-${this.type}` : ''
  }
}
```

When I run the watcher, I see errors about `no such file or directory, open...` and watch's process is stopped.

**Expected/desired behavior:**

* add some nonexistent file in component
* **What is the expected behavior?**
The watcher is still working.

* **What is the motivation / use case for changing the behavior?**
If I add `import` before then create a file, the watcher will be crashed and I should restart it manually.

PR #498, opened 2017-02-25
fix(watch): don't stop watch due to nonexisting files

JeroenVinke:
closes https://github.com/aurelia/cli/issues/408

image

This bug is mentioned in https://github.com/aurelia/cli/issues/461 but this does not close https://github.com/aurelia/cli/issues/461. I wonder if we can somehow call [gulp-notify](https://github.com/aurelia/cli/blob/master/lib/resources/tasks/transpile.js#L22) from the CLI in cases like this.  

I'd recommend to wait for https://github.com/aurelia/cli/pull/419 to be merged so we can confirm that this bug does not reintroduce itself when traces are cached

### DIFF
--- a/lib/build/bundled-source.js
+++ b/lib/build/bundled-source.js
@@ -52,6 +52,7 @@ exports.BundledSource = class {
     let rootDir = process.cwd();
     let moduleId = this.moduleId || (this.moduleId = this.calculateModuleId(rootDir, loaderConfig));
     let that = this;
+    let modulePath = this.path;
 
     console.log(`Tracing ${moduleId}...`);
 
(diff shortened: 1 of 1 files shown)


AStoker:
This would be quite helpful. Can't tell you how many times I've accidentally put in the wrong url for a file and have the whole "watch" stop :P 
@JeroenVinke, are you able to test locally on your machine the combination of this PR and #419 ? I don't think #419 is going to go in until we catch all the edge cases that suddenly come up with it.

AStoker:
I think that `gulp-notify` isn't included because we were trying to keep the CLI's footprint small. @EisenbergEffect can comment on that if he wants. I think that there are benefits of having `gulp-notify` so that we could bubble some of these errors up and not have them lost in a sea of tracing logs. Is there a way that we can see if it's available, and if so, use it? That way it's not a mandatory dependency? Just thinking...

AStoker:
@EisenbergEffect, can we get this merged?

EisenbergEffect:
Yep 😁

Issue #500, opened 2017-02-25
[Feature Request] generate with au commands in npm scripts


sbolel:
**I'm submitting a feature request**

CLI should generate a new aurelia-app with `au` commands inserted in `scripts` of package.json

**Current behavior:**

  ```
  $ npm start
  
  npm ERR! missing script: start
  ```

**Expected/desired behavior:**
  
  ```
  $ npm start
  
  > aurelia-app@0.1.0 start /Users/sinanbolel/dev/@aurelia/aurelia-app
  > au run
  
  Starting 'readProjectConfiguration'...
  Finished 'readProjectConfiguration'
  ...etc
  ```

**What is the motivation / use case for changing the behavior?**

  I feel it's adhering to standards to include these in npm scripts. npm scripts may also be more familiar for beginner Aurelia enthusiasts.

sbolel:
Closed issue based on feedbacm

Issue #470, opened 2017-01-25
How to keep aurelia.json clean


suamikim:
**I'm submitting a feature request**

* **Library Version:**
How can I check the lib version?

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
6.9.4

* **NPM Version:**
4.1.1

* **Browser:**
Request is not browser related

* **Language:**
TypeScript 2.1.5

**Current behavior:**
When installing 3rd party scripts, I have to add them manually to `aurelia_project/aurelia.json`, which will be handled automatically by future versions of the CLI according to some Aurelia announcements.

What happens if I remove a dependency from my project? Will they also be removed automatically by the CLI? If so, is this also true for "sub dependencies" and will there be a check on whether they are required by other dependencies as well?

**What is the motivation / use case for changing the behavior?**

Short example: I've installed package [md5](https://www.npmjs.com/package/MD5) from npm and added the following lines to `aurelia.json`:

```
// ...
"dependencies": [
	// ...
	"md5",
	"crypt",
	"charenc",
	"is-buffer",
	// ...
]
// ...
```

The packages `crypt`, `charenc` & `is-buffer` are currently solely required by `md5`. If I'd like to remove `md5` sometime in the future there are some things I have to consider in order to keep the `aurelia.json` clean and not include unnecessary dependencies in my builds:

* Which dependencies can I remove with `md5`? How do I know which were the sub dependencies of `md5`?
    * Right now I keep them "logically" grouped inside `aurelia.json` but this won't allow me to refactor it in the future (e.g. arrange dependencies alphabetically)...
    * Also since I can't use comments inside JSON there's no way to tell other contributors why I've grouped the dependencies the way I did. They have to know it somehow before touching `aurelia.json`.
* How do I know if any of the sub dependencies I've removed are not needed by some other 3rd party script? Trial & error?

Will those things also handled automatically by the CLI in the future? If not, what are best practices/recommendations for the considerations I've mentioned above?

suamikim:
I'd really appreciate some input on this. How do people who already use the CLI in larger projects actually handle this?

AStoker:
[...]

PR #501, opened 2017-02-26
chore(#500): add npm scripts for au commands

sbolel:
### Summary of Changes

- implements [Feature Request] #500
- new projects are generated with a package.json that contains `au` commands in `scripts`

### DIFF
--- a/lib/commands/new/project-template.js
+++ b/lib/commands/new/project-template.js
@@ -19,6 +19,12 @@ exports.ProjectTemplate = class {
         url : '???'
       },
       license: 'MIT',
+      scripts: {
+        build: 'au build',
+        debug: 'au run --watch'
+        start: 'au run',
+        test: 'au test'
+      },
       dependencies: {},
       peerDependencies: {},
       devDependencies: {}


AStoker:
Are there any downsides to this? I see the usefulness of this. On the other hand, would it cause confusion, because "building" could now happen with either `npm` or `au`, and if somebody has a custom build event (`au someBuild`), we don't want them to think that they can just use `npm someBuild`... What are your thoughts? @EisenbergEffect, what are your thoughts on that?

sbolel:
Nice input @AStoker thanks!

- I agree it could lead to confusion. 
- Also, the two definitions could fall out of sync.

Frankly, I'm not entirely sure if I understand my own perceived issue 😅 

I enjoy using `au` as my primary cli because I can add custom commands with pretty help messages, no sweet. I'm concerned that if/when I don't use `npm` because...

I'm scared of the monster living under my bed.

sbolel:
@AStoker, your https://github.com/aurelia/cli/issues/470#issuecomment-283146716 is great input for this as well.

AStoker:
In regards to the monster under the bed, you could always just put the mattress directly on the floor, no hiding places then 😉 
In regards to the PR, I think it might be more up to @EisenbergEffect on which direction he wants to take. My [gut feeling](http://theawkwardyeti.com/comic/gut-instincts/) is that while yes, it provides some initial "lightening of the load", in the long run it might hurt more than help.

sbolel:
> you could always just put the mattress directly on the floor, no hiding places then 

That's what I keep telling her, but she's all like "you never left college." `#yolo`

EisenbergEffect:
If I can, I'd like to punt on this for now unless we hear a lot more community requests for this. I agree that it could become a source of confusion. I'm just not confident about doing it at this time.

sbolel:
Agreed 👍

Issue #510, opened 2017-03-01
Crash in CLI on syntax error


jugglingthebits:
**I'm submitting a bug report**

* **Library Version:**
0.24.0

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
7.6.0

* **NPM Version:**
4.1.2

* **Language:**
TypeScript 2.2.1

**Current behavior:**
`au build` crashes when compiling typescript with invalid syntax in application code. Not all cases of invalid syntax crash. The resulting error message contains way too much information and is confusing ([example](https://gist.github.com/jugglingthebits/6157e97ce41560cc76a16df190d21cfd)).
My main issue is that `au run --watch` also crashes which makes using the watcher much less useful.
Repro:

    git clone https://github.com/jugglingthebits/au-ts-syntax-error
    cd au-ts-syntax-error
    npm install
    au build *or* au run --watch

**Expected/desired behavior:**
`au build` should inform about the syntax error - not crash.
`au run --watch` should not crash on syntax error and ignore that file.

* **What is the motivation / use case for changing the behavior?**
`au run --watch` should not crash when application code is currently not in a valid state: Code editors may save at any point and a background watcher shouldn't just stop due to syntax errors.

AStoker:
@jugglingthebits, could you test this again with the latest CLI? We've made a few improvements that might solve your issue.
Links to relevant issues/PR's to have everything referenced:
https://github.com/aurelia/cli/pull/511
https://github.com/aurelia/cli/issues/504
https://github.com/aurelia/cli/pull/498

jugglingthebits:
Just tested with 0.26.1. `au run --watch` doesn't crash on these syntax errors any more - nice!

Now, it doesn't start the build-run-watch-loop either but just stops with the same really long error message. IMHO, it should should enter the loop and print the same nice error messages as if those error hat been introduced after it was started correctly.

`au build` displays the same error message as above now, which could be improved. IMHO. this isn't a huge problem though.

I think this issue can be closed, although the error messages are by far not as nice as they should be yet.

AStoker:
If you have some ideas on improving the logging, PR's are more than welcome! :)

Issue #504, opened 2017-02-27
Improve error logging


JeroenVinke:
**I'm submitting a feature request**
This purpose of this issue is to track all instances where errors during the bundling/tracing phase are silenced

**Current behavior:**
Some tracing/bundling related errors are currently silenced. Some examples mentioned in:

- https://github.com/aurelia/cli/issues/502
- https://github.com/aurelia/cli/pull/419
- https://github.com/aurelia/cli/issues/190

**Expected/desired behavior:**
All errors to be reported

AStoker:
My name is Andrew Stoker, and I approve of this issue. ;)
Some thoughts I had in addition to this were having optional flags to log out various levels of verbosity. e.g. `au build --verbose` would be _verbose_ in it's messages, helping out with debugging, while `au build --silent` would output only errors...

PR #511, opened 2017-03-01
Ignore errors during tracing to not crash the cli

jugglingthebits:
I am not sure this is the right position within the codebase to handle syntax errors in TS files, but as far as I can see this fix works for #510 .

### DIFF
--- a/lib/build/bundled-source.js
+++ b/lib/build/bundled-source.js
@@ -122,6 +122,8 @@ exports.BundledSource = class {
           }, this.includedBy);
         }
       }
+    }, traceReject => {
+      // Ignore failed traces
     });
   }
 


JeroenVinke:
Thanks for creating the issue and the PR! This is one of those spots where we should add some logging as part of https://github.com/aurelia/cli/issues/504. So i'm not sure if swallowing the errors here is the best solution for this problem. We definitely want to prevent the watcher from crashing, but this may swallow other errors that we do want to show

jugglingthebits:
I see. I thought that is a different issue as in my case it isn't swallowing the error at all - quite the opposite.

JeroenVinke:
Yes, you're right, it is a different issue. But we want to keep #504 in mind here as we don't want to swallow errors that we don't have to

AStoker:
Having been bit by swallowed errors before, I'm hesitant to just just straight up swallow without any kind of logging. What can happen is some bad code might be pulled down, and the compilation will fail, but not give the user any kind of indication on what just happened, or where. No bueno there 😞

lib/build/bundled-source.js
```diff
@@ -122,6 +122,8 @@ exports.BundledSource = class {
           }, this.includedBy);
         }
       }
+    }, traceReject => {
+      // Ignore failed traces
```
EisenbergEffect: Can we log the errors at least, instead of ignoring them?

jugglingthebits:
I will just close this, as #510 has been fixed elsewhere.

PR #499, opened 2017-02-25
fix(#494): bundling of imports w/ non-relative paths

sbolel:
### Summary

- Fixes #494 - enable bundling of imports with non-relative paths
- The [Bash session](https://gist.github.com/sbolel/2b8bf70db9239c980b1844f058f6df3a) during new typescript app creation
- A [demo app PR](https://github.com/sbolel/aurelia-cli-pulls-499-demo/pull/1) showing use of non-relative paths in a newly generated typescript app

#### Changes

- Add `lib`, `baseUrl`, `paths` to _lib/resources/content/tsconfig.json_
- Update Typescript dependencies in _lib/dependencies.js_
- Update `aurelia-tools` to `^1.0.0` in _package.json_ and _lib/dependencies.js_
- Change `import from  'gulp-changed'` to `import from 'gulp-changed-in-place'`
- Remove unused Protractor type definitions

#### TODOs

- [x] fix protractor type errors: `Namespace 'protractor' has no exported member 'Protractor'.`

PR #512, opened 2017-03-01
Updated aurelia dependencies + remove unused protractor typedefs

JeroenVinke:
Updated the aurelia libraries and confirmed that both ESNext and typescript projects created by the CLI still worked (au run + au test)

I did encounter a typescript error (`Namespace 'protractor' has no exported member 'Protractor'.`) for which a fix was already supplied by @sbolel in https://github.com/aurelia/cli/pull/499. So I cherrypicked that commit, and we can take a look at the rest of https://github.com/aurelia/cli/pull/499 at a later time

cc @AStoker @EisenbergEffect @AshleyGrant


sbolel:
Nice! Thanks for pulling in the commit from #499 @JeroenVinke!

JeroenVinke:
@EisenbergEffect I do wonder what the protractor typescript definition is used for since I do not see any mention of protractor elsewhere in the codebase

EisenbergEffect:
It's an incomplete feature basically.

Issue #514, opened 2017-03-03
rev issue: index.html not updated


tdamir:
**I'm submitting a bug report**

* **Library Version:**
0.25.0

**Please tell us about your environment:**
* **Operating System:**
Windows [10]

* **Node Version:**
6.9.2

* **NPM Version:**
4.3.0

* **Browser:**
all

* **Language:**
all

**Current behavior:**
If rev is enabled in aurelia.json, the bindle revisions are generated but index.html is not updated according to that.

**Expected/desired behavior:**
It should replace the index.html like in previous version.
I've looked at the source and it looks like index is replaced before the revisioned filename is generated. If I change the order, it works fine.

Original bundle.js (line: 182)
```javascript

      if (loaderOptions.configTarget === this.config.name) {
        //Add to the config bundle the loader config and change the "index.html" to reference the appropriate config bundle
        concat.add(undefined, this.writeLoaderCode(platform));
        contents = concat.content;
        let outputDir = platform.baseUrl || platform.output; //If we have a baseUrl, then the files are served from there, else it's the output
        if (platform.index) {
          this.setIndexFileConfigTarget(platform, path.posix.join(outputDir, bundleFileName));
        }
      }

      if (buildOptions.rev) {
        //Generate a unique hash based off of the bundle contents
        this.hash = generateHash(concat.content);
        bundleFileName = generateHashedPath(this.config.name, this.hash);
      }

```
Modified version working as expected:

```javascript
      if (buildOptions.rev) {
        //Generate a unique hash based off of the bundle contents
        this.hash = generateHash(concat.content);
        bundleFileName = generateHashedPath(this.config.name, this.hash);
      }

      if (loaderOptions.configTarget === this.config.name) {
        //Add to the config bundle the loader config and change the "index.html" to reference the appropriate config bundle
        concat.add(undefined, this.writeLoaderCode(platform));
        contents = concat.content;
        let outputDir = platform.baseUrl || platform.output; //If we have a baseUrl, then the files are served from there, else it's the output
        if (platform.index) {
          this.setIndexFileConfigTarget(platform, path.posix.join(outputDir, bundleFileName));
        }
      }
```

AStoker:
Dangit... This is why we need to get those tests in there. When I tested out my "fix", it must have been using a cached version, so I didn't test everything out like I thought I did. Essentially, we moved the hashing function up to solve a different bug, but failed to modify when the updated hash is written to the .html file. I'll have to get the fix in real quick

[...]

PR #515, opened 2017-03-03
fix(bundle): write index file modifications after hash generation

AStoker:
Resolves: https://github.com/aurelia/cli/issues/514
Ensures that the index file is modified (when using revisions) at the
correct time. Previously, it was modified before the hashes were
created, resulting in an index file that was not referencing the
correct file path. Now the index file is modified after the hash is
generated.

### DIFF
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -180,17 +180,25 @@ exports.Bundle = class {
       let bundleFileName = this.config.name;
 
       if (loaderOptions.configTarget === this.config.name) {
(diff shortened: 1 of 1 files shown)


AStoker:
I made a stupid mistake that wasn't properly tested in the previous release.
I'm not angry, I'm just disappointed... In myself 😞 
Anyway, @EisenbergEffect, can you take a look at this real quick and get it merged in a jiffy so we don't have a slew of issues all pointing fingers back at me? 😬

PR #518, opened 2017-03-04
Fix/base64 sourcemap

JeroenVinke:
This will prevent the build process from crashing in two ways:

- when base64 sourcemaps are used (such as in dragula.js)
- when sourcemap files are incorrectly referenced (noticed this with interactjs)

### DIFF
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -154,9 +154,16 @@ exports.Bundle = class {
             return null;
           }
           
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
@JeroenVinke Is this all ready to go? If so, I'll merge and release right away. Just let me know.

JeroenVinke:
Yes, it is good to go

Issue #400, opened 2016-10-29
Trace dependencies is never set to false for resources


rkrdovrgs:
**I'm submitting a bug report**
- **Library Version:**
  0.21.0

**Please tell us about your environment:**
- **Operating System:**
  Windows 7
- **Node Version:**
  6.2.0
- **NPM Version:**
  3.9.6
- **Browser:**
  all
- **Language:**
  TypeScript

**Current behavior:**
- Not possible to add a bundle dependency that has resources, and you want to disable `traceDependencies` by setting the pattern using `[]` [see here - line 12](https://github.com/aurelia/cli/blob/master/lib/build/source-inclusion.js#L12)
- For the following dependency:

``` json
{
   "name": "my-dependency",
   "path": "../node_modules/my-dependency/amd",
   "main": "index",
   "resources": [
      "[**/*.js]"
   ]
}
```
- Pattern is supposed to disable trace dependencies, but when it is evaluated, pattern value becomes  `"../node_modules/my-dependency/amd/[**/*.js]"`. Two things ocurred:
-  `pattern[0]` is not equal to `'['` meaning `this.traceDependencies` is set to `true`.
- The pattern will not return any matches.
- Also typo on [line 10](https://github.com/aurelia/cli/blob/master/lib/build/source-inclusion.js#L10)

**Expected/desired behavior:**
- **What is the expected behavior?**
  The pattern disables `traceDependencies` and the pattern return matches.
- **What is the motivation / use case for changing the behavior?**
  Allow disabling `traceDependencies` for dependency resource patterns.

AStoker:
Just to clarify, you're wanting to be able to trace dependencies by using the bracket (`[]`) syntax for resources, right?
The bracket syntax is specific to Aurelia, not a minimatch/glob feature. So I believe it's only good for defining source locations. Dependencies and resources I think have to be more specific. You'd probably want to be real careful trying to trace resources of dependencies, as you might add a large amount of time tracing things like jQuery or other paths. That's why you have a `deps` option for resources, so you can define what dependencies are needed for this dependency.   
Could you give a use case for wanting this kind of feature?

rkrdovrgs:
@AStoker no, I am talking about the opposite. I want to be able to disable tracing by using the brackets. This is not an enhancement, this is a bug. As you suggested, I actually was talking about a dependency:
```
{
   "name": "bundle-name.js"
   "dependencies": [
      {
         "name": "my-dependency",
[...]

Issue #342, opened 2016-09-27
Sub-dependencies of bundle dependencies are going into wrong bundle


pfurini:
**I'm submitting a bug report**
- **Library Version:**
  0.20.0

**Please tell us about your environment:**
- **Operating System:**
  OSX 10.11.6
- **Node Version:**
  6.4.0
  
  
- **NPM Version:**
  3.10.8
  
  
- **Browser:**
  all
- **Language:**
  TypeScript 2.0

**Current behavior:**
When you have the standard bundling config in `aurelia.json` (one vendor-bundle and one app-bundle), only the root file of listed `dependencies` goes into the vendor-bundle, while the rest of the package files goes into the app-bundle.
Here is an example of `bundles` node in `aurelia.json` file:

```
    "bundles": [
      {
        "name": "app-bundle.js",
        "source": [
          "[**/*.js]",
          "**/*.{css,html}"
        ]
      },
      {
        "name": "vendor-bundle.js",
        "prepend": [
          "node_modules/bluebird/js/browser/bluebird.core.js",
          "scripts/require.js"
        ],
        "dependencies": [
          "whatwg-fetch",
          "aurelia-binding",
          "aurelia-bootstrapper",
          "aurelia-dependency-injection",
          "aurelia-event-aggregator",
          "aurelia-framework",
          "aurelia-history",
          "aurelia-history-browser",
          "aurelia-loader",
          "aurelia-loader-default",
          "aurelia-logging",
          "aurelia-logging-console",
          "aurelia-metadata",
          "aurelia-pal",
          "aurelia-pal-browser",
          "aurelia-path",
          "aurelia-polyfills",
          "aurelia-route-recognizer",
          "aurelia-router",
          "aurelia-task-queue",
          "aurelia-templating",
          "aurelia-templating-binding",
          "extend",
          {
            "name": "aurelia-fetch-client",
            "path": "../node_modules/aurelia-fetch-client/dist/amd",
            "main": "aurelia-fetch-client"
          },
          {
            "name": "aurelia-api",
            "path": "../node_modules/aurelia-api/dist/amd",
            "main": "aurelia-api"
          },
          {
            "name": "jwt-decode",
            "path": "../node_modules/jwt-decode/lib",
            "main": "index"
          },
          {
            "name": "aurelia-authentication",
            "path": "../node_modules/aurelia-authentication/dist/amd",
            "main": "aurelia-authentication"
          },
          {
            "name": "text",
            "path": "../scripts/text"
          },
[...]

PR #525, opened 2017-03-07
fix(source-inclusion): detect traceDependencies from dependency resources correctly

JeroenVinke:
https://github.com/aurelia/cli/issues/400

There are actually two issues preventing https://github.com/aurelia/cli/issues/400 from working. One is fixed by this PR, and the other one seems to be https://github.com/aurelia/cli/issues/342.

To reproduce https://github.com/aurelia/cli/issues/400:
- clone [this app](https://github.com/JeroenVinke/aurelia-cli-400)
- add the following folder to `node_modules`: [my-dependency.zip](https://github.com/aurelia/cli/files/825723/my-dependency.zip)

The setup is as follows. The aurelia-cli-400 app has a dependency called `my-dependency` which has two files (`resource1.js` and `resource2.js`) where `resource2.js` is a dependency of `resource1.js`. The `my-dependency` dependency is defined in the `vendor-bundle`:
```
          {
            "name": "my-dependency",
            "path": "../node_modules/my-dependency",
            "main": "index",
            "resources": [
                "[resource1.js]"
            ]
          }
```

What you'd expect using this setup is that `resource1.js` ends up in the `vendor-bundle` but `node_modules/my-dependency/resource2.js` does not. Without the changes in this PR both `resource1.js` and `resource2.js` are in **`app-bundle.js`**. This can be explained due to the glob pattern in SourceInclusion being invalid (as mentioned in https://github.com/aurelia/cli/issues/400), which causes neither `resource1.js` or `resource2.js` to be included in the `vendor-bundle`. Due to https://github.com/aurelia/cli/issues/342 they end up in the **app-bundle** instead.

With these changes, only `resource1.js` ends in `vendor-bundle`, which is what you'd expect. Although https://github.com/aurelia/cli/issues/342 is causing `resource2.js` to be included in `app-bundle`, so https://github.com/aurelia/cli/issues/400 is not completely "fixed" by this PR

### DIFF
--- a/lib/build/dependency-inclusion.js
+++ b/lib/build/dependency-inclusion.js
@@ -1,6 +1,5 @@
 "use strict";
 const os = require('os');
-const path = require('path');
 const SourceInclusion = require('./source-inclusion').SourceInclusion;
 
 exports.DependencyInclusion = class {
@@ -20,9 +19,8 @@ exports.DependencyInclusion = class {
     if (resources) {
       let promises = [];
 
-      resources.forEach(x => {
-        let pattern = path.join(loaderConfig.path, x);
-        let inclusion = new SourceInclusion(bundle, pattern);
+      resources.forEach(pattern => {
+        let inclusion = new SourceInclusion(bundle, pattern, loaderConfig);
         promises.push(inclusion.addAllMatchingResources(loaderConfig));
         bundle.includes.push(inclusion);
       });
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
@JeroenVinke Feel free to merge when you feel it's ready.

JeroenVinke:
I'm going to take another stab at this and open a new PR

Issue #533, opened 2017-03-10
au install i18next has wrong path in aurelia.json


anthonyborell:
**I'm submitting a bug request**

**Library Version:**
0.26.1

**Operating System:**
Windows 10

**Node Version:**
6.10.0

**NPM Version:**
3.10.8

**Browser:**
all

**Language:**
all

**Current behavior:**
Similar to @AshleyGrant issue. Was following the aurelia-i18n tutorial on the docs hub but using v0.26.1 of the CLI. 

When using `au install i18next` the output was:

```
{
    "name": "i18next",
    "main": "./index.js",
    "path": "../node_modules/i18next",
    "resources": []
 }
```

however when performing a build I was getting the following error:

```
File not found or not accessible: D:/source/code/app/i18next/dist/commonjs/index.js. Requested by D:/source/code/app/src/main.js
File not found or not accessible: D:/source/code/app/i18next/dist/commonjs/index.js. Requested by D:/source/code/app/src/main.js
```

To fix the issue I had to change the path to:

```
{
    "name": "i18next",
    "path": "../node_modules/i18next/dist/umd",
    "main": "i18next",
    "resources": []
}
```

JeroenVinke:
I think this may be the same issue as https://github.com/aurelia/cli/issues/519. i18next uses CommonJS (similar to cheerio) and also uses the [.js](https://github.com/i18next/i18next/blob/master/index.js#L1) extension in the require statement (similar to cheerio). The result is that `node_modules` is missing from the path, which is also what happens in https://github.com/aurelia/cli/issues/519. If this issue can be resolved by removing the .js extension [here](https://github.com/i18next/i18next/blob/master/index.js#L1) then it is definitely the same issue

AStoker:
I have seen many libraries import modules with the `.js` extension, and it has caused many headaches for me. Can you think of any way to work around this?

JeroenVinke:
I'm not exactly sure where this fails yet. Perhaps we can strip off the .js extension somewhere, but we'd have to find the right spot

JeroenVinke:
https://github.com/amodrojs/amodro-trace/issues/7

zewa666:
Fixed with new auto-tracer

PR #543, opened 2017-03-15
fix(build): strip .js extension from moduleIds

JeroenVinke:
resolves https://github.com/aurelia/cli/issues/533

I'd like to discuss this with the Amodrotrace folks, but requirejs behaves differently when the .js extension is used: https://github.com/aurelia/cli/blob/d6f0ada18c4832bf0ec8296175ef1862047dce56/lib/build/amodro-trace/lib/loader/require.js#L1639

Stripping the .js extension in the override of amodrotrace seems to resolve the issue


JeroenVinke:
Added this to https://github.com/aurelia/cli/pull/589 since that PR adds a function to amodrotrace that does module id normalization

PR #546, opened 2017-03-16
fix(test): esling reports error that path should never be concated as string

k2s:
fixes ESLint error: Use path.join() or path.resolve() instead of + to create paths.


JeroenVinke:
Thanks! Could you squash your commits and use the appropriate commit message format (described in [this section](https://github.com/DurandalProject/about/blob/master/CONTRIBUTING.md#commit))? Some examples can be found [here](https://github.com/aurelia/cli/commits/master). The other PR also needs somes squashing and renaming

Issue #547, opened 2017-03-16
[Feature] Add view-viewmodel generator


johot:
The first thing I expected to find among the generate commands was a command that would generate a view-viewmodel pair. Is it possible we could have one added?

Thanks!

AStoker:
Creating a view and view model are pretty simple. However, I wonder how exactly we could handle this. With creating resources, we have a semi-defined structure of where things go, so we can automate that task rather well (it also doesn't have to navigate very far).  
For creating view/viewmodel pairs, there are a few different ways people do that. Some people might split up their views and view models into specific folders called `views` and `viewmodels` (this is a common scheme I've seen for small projects). Others might just have everything at the top level (again, common for small projects). And finally others have a scheme set up where things are broken down into functional areas (useful for larger projects).  
I'd be interested to know of an approach we could use that would account for all those situations... Sure, we could just generate the pair at the root level, but then the user would most likely still have to split up/move the files (at which point I'm not too sure how helpful it would _truly_ be). 

How do you envision using this kind of feature?

johot:
Hi Andrew, thank you for writing me!

I understand you concerns that different projects use different
conventions. I do however think it is important to have a "au generate
component" generator because it's the first thing you look for when dealing
with a tool such as the aurelia cli. In my case I was very confused that it
didn't exist to the point that I thought the aurelia cli shouldn't be used
for things like this and started looking for other tools.

I think we should solve it in the same way as the anglar cli does it. That
is we generate the view-viewmodel pair in the folder the user is currently
located in. Just like in the angular cli the user the should also be able
to specify relative paths for example:

*"au generate component feature/my-component-name"*

This would put the files in the "src/feature/" folder (creating the feature
folder if it does not exist). Typing *"au generate my-component-name"*
would put it in the current directory which could be the src folder and we
should even support *"au generate ../my-component-name"*.

Example from the angular docs:

https://github.com/angular/angular-cli#generating-components-directives-pipes-and-services
Section "Generating Components, Directives, Pipes and Services"

In the case of putting the view and viewmodels in diffent folder maybe we
could solve it by specifying other arguments, for example

*"au generate my-component-name view-folder view-model-folder"*

or simply not support having them in different folders, I think we should
promote "structure by features" anyways. The most important thing is that
we at least have the command since it's really missing (I think).

What do you think? :)
Thanks

/Johan




2017-03-17 14:02 GMT+01:00 Andrew Stoker <notifications@github.com>:

[...]

PR #560, opened 2017-03-21
Added default generator for creating components (view model + view)

johot:
I really think we should have a generator for creating components (view model and views. It's the first thing I started looking for when downloading the CLI. 

I discussed this here: https://github.com/aurelia/cli/issues/547 and was asked to create a PR so here it is :)

To use this feature you will write `au generate component`

* You will then input the name of the component.
* After this you will be asked if you want to add the component to any sub folders (for example your feature folders). 

If you leave it empty you component files will be added to the src folder. If you enter a sub folder (or sub folders) that does not exist they will be created.


EisenbergEffect:
I like it! @johot Can you also add the vanilla js version of the task?

johot:
@EisenbergEffect thank you!

Ah totally forgot about the js version :), it has been added now! Also tested creating a new project with "au new" both with typescript and ESNext and they au generate component seemed to work fine.

EisenbergEffect:
Thanks!

PR #564, opened 2017-03-28
fix(build): enforce strict mode

JeroenVinke:
fixes https://github.com/aurelia/app-ux-showcase/issues/16


JeroenVinke:
@EisenbergEffect you may want to create a patch release for this

Issue #567, opened 2017-03-28
Error when bundling resources from outside node_modules


avrahamcool:
**I'm submitting a bug report**

* **Library Version:**
0.27.0

* **Operating System:**
Windows 10

* **Node Version:**
v6.9.5

* **NPM Version:**
3.10.9

* **Browser:**
all 

* **Language:**
ESNext

**Current behavior:**
1. when I'm trying to bundle npm libraries from outside `node_modules` folder I get an error from `aurelia-templating-resources`.
2. when I'm trying to bundle libraries that don't come from npm - I get a bundling error.

**step to reproduce**
To showcast the problem, I created a new project and installed `jquery` via npm.
I added it to the bundle like this:

```
{
   "name": "jquery",
   "path": "../node_modules/jquery/dist",
   "main": "jquery.min"
}
```

and everything worked fine

```
Starting 'readProjectConfiguration'...
Finished 'readProjectConfiguration'
Starting 'processMarkup'...
Starting 'processCSS'...
Starting 'copyFiles'...
Starting 'configureEnvironment'...
Finished 'copyFiles'
Finished 'processCSS'
Finished 'processMarkup'
Finished 'configureEnvironment'
Starting 'buildJavaScript'...
Finished 'buildJavaScript'
Starting 'writeBundles'...
Tracing app...
Tracing environment...
Tracing main...
Tracing resources/index...
Tracing app...
Tracing aurelia-binding...
Tracing aurelia-bootstrapper...
Tracing aurelia-dependency-injection...
Tracing aurelia-event-aggregator...
Tracing aurelia-framework...
Tracing aurelia-history...
Tracing aurelia-history-browser...
Tracing aurelia-loader-default...
Tracing aurelia-logging-console...
Tracing aurelia-pal-browser...
Tracing aurelia-route-recognizer...
Tracing aurelia-router...
Tracing aurelia-templating-binding...
Tracing text...
Tracing jquery...
Tracing aurelia-templating-resources...
Tracing aurelia-templating-router...
Tracing aurelia-testing...
Writing app-bundle.js...
Writing vendor-bundle.js...
Finished 'writeBundles'
```

then I took the `jquery` directory out of `node_mdules` and put it into a directory called `external_modules` (this directory is a brother to the `node_modules` directory ).

I changed my bundle config accordingly to:
```
{
   "name": "jquery",
   "path": "../external_modules/jquery/dist",
   "main": "jquery.min"
}
```

and I get this error when building:

```
[...]

Issue #575, opened 2017-03-30
Support packages without package.json file


JeroenVinke:
original bug report: https://github.com/aurelia/cli/issues/567

**I'm submitting a feature request**

* **Library Version:**
0.27.0

**Current behavior:**
An error is thrown whenever the bundler expects a package to have a package.json file:

```
Starting 'readProjectConfiguration'...
{ uid: 1,
  name: 'readProjectConfiguration',
  branch: false,
  error:
   { Error: ENOENT: no such file or directory, open 'C:\Users\USERNAME\Desktop\cli-bundle-error-show-case\node_modules\keylines\package.json'
       at Error (native)
     errno: -4058,
     code: 'ENOENT',
     syscall: 'open',
     path: 'C:\\Users\\USERNAME\\Desktop\\cli-bundle-error-show-case\\node_modules\\keylines\\package.json' },
  duration: [ 0, 57820956 ],
  time: 1490709458797 }
{ uid: 0,
  name: '<series>',
  branch: true,
  error:
   { Error: ENOENT: no such file or directory, open 'C:\Users\USERNAME\Desktop\cli-bundle-error-show-case\node_modules\keylines\package.json'
       at Error (native)
     errno: -4058,
     code: 'ENOENT',
     syscall: 'open',
     path: 'C:\\Users\\USERNAME\\Desktop\\cli-bundle-error-show-case\\node_modules\\keylines\\package.json' },
  duration: [ 0, 63698917 ],
  time: 1490709458813 }
{ Error: ENOENT: no such file or directory, open 'C:\Users\USERNAME\Desktop\cli-bundle-error-show-case\node_modules\keylines\package.json'
    at Error (native)
  errno: -4058,
  code: 'ENOENT',
  syscall: 'open',
  path: 'C:\\Users\\USERNAME\\Desktop\\cli-bundle-error-show-case\\node_modules\\keylines\\package.json' }
```

* **What is the expected behavior?**
The build to continue without errors

AStoker:
This might help with packages that are css only too

PR #576, opened 2017-03-30
find location of packages outside of node_modules, support packages without package.json

JeroenVinke:
fixes https://github.com/aurelia/cli/issues/567

@AStoker i'd like to hear your thoughts on the following. We need the root folder of a package (kept in `description.location` and looks something like `c:/development/my-app/node_modules/my-plugin/`) for two things:

1. to look for (and read) the package.json of the package (which is [used](https://github.com/aurelia/cli/blob/master/lib/build/package-analyzer.js#L48) to determine the loader configuration)
2. to determine [what bundle](https://github.com/aurelia/cli/issues/342#issuecomment-286362197) to put files in. In other words, to prevent files from ending up in `app-bundle.js` when they should have ended up in `vendor-bundle.js` (or other)

Currently the root folder of packages inside the node_modules folder is determined as follows. You resolve the `path` of the package (as defined in aurelia.json) which could be `c:/development/my-app/node_modules/my-plugin/dist/` and traverse up the folder tree until you hit the `node_modules`. The path until the first segment after `node_modules` is then used as the package root folder (`c:/development/my-app/node_modules/my-plugin`).

When you keep packages outside the node_modules folder, this trick does not work because there is no standard folder name like `node_modules` to look for in the path. So as far as I know there is no reliable way of knowing what the root folder of the package is when the package is not inside the `node_modules` folder. 

So this PR enforces that if you define dependencies outside the `node_modules` folder, that you must define a `packageRoot` :

```
          {
            "name": "jquery",
            "path": "../external_modules/jquery/dist",
            "packageRoot": "../external_modules/jquery",
            "main": "jquery.min"
          }
```

### DIFF
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -94,7 +94,7 @@ exports.Bundle = class {
   }
 
   getDependencyLocations() {
-    return this.includes.filter(inclusion => inclusion.description)
+    return this.includes.filter(inclusion => inclusion.description && inclusion.description.location)
       .map(inclusion => {
         let normalizedLocation = path.posix.normalize(inclusion.description.location).replace(/\\/g, '\/');
         return {
--- a/lib/build/package-analyzer.js
+++ b/lib/build/package-analyzer.js
@@ -34,9 +34,13 @@ exports.PackageAnalyzer = class {
 
 function loadPackageMetadata(project, description) {
   return setLocation(project, description)
-    .then(() => fs.readFile(description.metadataLocation))
-    .then(data => {
-      description.metadata = JSON.parse(data.toString());
+    .then(() => {
+      if (description.metadataLocation) {
+        return fs.readFile(description.metadataLocation)
+        .then(data => {
+          description.metadata = JSON.parse(data.toString());
+        });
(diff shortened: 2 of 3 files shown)


AStoker:
I like that idea. I won't have time to really look into it until Monday though (gone over the weekend). So in your example, it would search up `external_modules` the same way it would search up `node_modules`?

JeroenVinke:
It wouldn't traverse the directory tree to find the package root like it does with `node_modules` as that's not necessary when the user defines the `packageRoot` in aurelia.json

AStoker:
Gotcha, I'm tracking now. Sounds reasonable to me!

JeroenVinke:
@EisenbergEffect I believe this is good to go. 

I decided to add another fix to this PR, because they depend on each other. So this fixes:
- https://github.com/aurelia/cli/issues/567
- https://github.com/aurelia/cli/issues/575

When you merge this, could you also merge https://github.com/aurelia/framework/issues/762?

AStoker:
@EisenbergEffect, sorry to bump. Could we get this in?

PR #578, opened 2017-04-01
Fix/paths aureliajson

### LINKED ISSUE #352: Tracing fails for indexes
- **Library Version:**
  0.20.1
- **Operating System:**
  OSX 10.11.6
- **Node Version:**
  v5.11.0
- **NPM Version:**
  3.8.6
- **Browser:**
  N/A
- **Language:**
  Babel ES6

**Current behavior:**

When using an import with a relative path, _au build_ works fine. But if the import uses a definition in aurelia.json:loader.config.paths, then the build works, but tracing fails.

The use case is a directory with an index which re-exports services. E.g.

```
export { Account } from './account.js'
export { Alert } from './alert.js'
export { Audit } from './audit.js'
export { Channel } from './channel.js'
```
## Simplified Test Case:

main.js:

```
import {B} from 'paks/a/index'
```

paks/a/index.js

```
export {B} from './b'
```

paks/a/b.js

```
export class B {}
```

In this scenario **au build** completes successfully both building and tracing.

However, when importing a module ID that is defined via loader.config.paths, the tracing fails to resolve. Change main.js to:

   import {B} from 'a'

Add a definition to aurelia.json to resolve 'a'

```
"loader": {
      "config": {
        "paths": {
            "a": "paks/a/index"
        }
      }
```

Then the tracing step fails on "main". It seems the config.paths are not being considered and tracing is trying to find "b.js" in the top level directory instead of under "paks/a".

```
{ uid: 7,
  name: 'writeBundles',
  branch: false,
  error:
   { [Error: ENOENT: no such file or directory, open '/Users/mob/dev/ab.1/src/b.js']
     errno: -2,
     code: 'ENOENT',
     syscall: 'open',
     path: '/Users/mob/dev/ab.1/src/b.js',
     moduleTree: [ 'a' ],
     fileName: '/Users/mob/dev/ab.1/src/paks/a/index.js' },
  duration: [ 0, 18755216 ],
  time: 1475530351775 }
```

Expect that config.paths should be uses when creating the module ID path for the import.

Have a large code base being refactored to use the CLI.

JeroenVinke:
closes #352 

Previously only `root` property of the `paths` object in aurelia.json was used by the CLI. The others were ignored. This PR fixes that

unfortunately, projects that were generated previously have incorrect paths. They have:

```
  "paths": {
    "root": "src",
    "resources": "src/resources",
    "elements": "src/resources/elements",
    "attributes": "src/resources/attributes",
    "valueConverters": "src/resources/value-converters",
    "bindingBehaviors": "src/resources/binding-behaviors"
  },
```

while all paths except for `root` should be relatively from the root path. 

So I have made sure that the CLI gracefully recovers from this situation and shows a warning message:

image

We can remove this graceful recovery logic after a while

### DIFF
--- a/lib/build/bundler.js
+++ b/lib/build/bundler.js
@@ -25,7 +25,7 @@ exports.Bundler = class {
 
     this.loaderConfig = {
       baseUrl: project.paths.root,
-      paths: {},
+      paths: ensurePathsRelativelyFromRoot(project.paths || {}),
       packages: [],
       stubModules: [],
       shim: {}
@@ -233,4 +233,27 @@ function subsume(bundles, item) {
 
 function normalizeKey(p) {
   return path.normalize(p);
+}
+
+function ensurePathsRelativelyFromRoot(p) {
+  let keys = Object.keys(p);
+  let original = JSON.stringify(p, null, 2);
+  let warn = false;
+
+  for (let i = 0; i < keys.length; i++) {
+    let key = keys[i];
+    if (key !== 'root' && p[key].indexOf(p.root + '/') === 0) {
+      warn = true;
+      p[key] = p[key].slice(p.root.length + 1);
+    }
+  }
+
+  if (warn) {
+    console.log('Warning: paths in the "paths" object in aurelia.json must be relative from the root path. Change ');
+    console.log(original);
+    console.log('to: ');
+    console.log(JSON.stringify(p, null, 2));
+  }
+
+  return p;
 }
--- a/lib/commands/new/project-template.js
+++ b/lib/commands/new/project-template.js
@@ -248,11 +248,11 @@ exports.ProjectTemplate = class {
 
     this.model.paths = Object.assign(this.model.paths, {
       root: appRoot,
-      resources: this.resources.calculateRelativePath(this.root),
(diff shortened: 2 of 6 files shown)


ghidello:
After updating to 0.28.0 the `au generate` command puts the files in the folder `<root>\resources` instead of `<root>\src\resources\`.
Digging in the `element` generator, I found out that the destination files are created in the folder specified by the [project.elements](https://github.com/aurelia/cli/blob/master/lib/resources/generators/element.js#L19) property which I think is populated by the following line of the [project constructor](https://github.com/aurelia/cli/blob/master/lib/project.js#L28):
```javascript
this.locations = Object.keys(model.paths).map(key => this[key] = ProjectItem.directory(model.paths[key]));
```
I'm totally not sure if I got it right and, in case I did, I don't know what would be the best way to solve the issue:

1. creating an helper method in the project class that adds the root folder to the specific resource type subfolder and change all the generators in order to call this function
2. change the project class constructor and prepend the root folder to all the resource folders

To me the first option looks the safest mostly because i can't tell how many regressions can be created by the second one

JeroenVinke:
@ghidello could you open a new issue for this? Thanks

ghidello:
Done #596

Issue #588, opened 2017-04-04
Static json files in dependencies breaks build


JeroenVinke:
**I'm submitting a feature request**

* **Library Version:**
0.27.0

**Please tell us about your environment:**

* **Node Version:**
all

* **NPM Version:**
all

**Current behavior:**
When a dependency uses `require('./some-json.json')`, the build produces the following error:

> Error: ENOENT: no such file or directory, open 'C:\Users\Jeroen\Downloads\test-app\src\test.json.js'

**Expected/desired behavior:**

The JSON file to be going through the JSON requirejs plugin and the build not breaking

JeroenVinke:
This one is very tricky mainly because of https://github.com/requirejs/requirejs/issues/674. SystemJS supports this nicely because you can use `meta`:

```
meta: {
  '*.json': { loader: 'json' }
}
```

Then whenever a file with the json extension is loaded, the json loader is used. According to https://github.com/requirejs/requirejs/issues/674, RequireJS does not have such option.

This means that we have to prefix all require calls (that include an extension) with a loader plugin. E.g. `require("some-dir/my-file.json")` becomes `require("json!some-dir/my-file.json")`. This needs to be done in two places: amodrotrace and our bundler.

JeroenVinke:
Another option: all paths that have an extension (other than .js), we set in the `paths` object we path to requirejs. So when `some-package` does `require('./my-file.json')` we set:

```
{
  "paths": {
      "some-package/my-file.json": "json!some-package/my-file.json"
  } 
}
`

3cp:
json is supported in latest cli-bundler without the need of mapping.

PR #589, opened 2017-04-04
Feat/support static json files

JeroenVinke:
Fixes https://github.com/aurelia/cli/issues/588

With the changes in this PR, if a `map` property is present in aurelia.json, it will be used in the requirejs config. That should allow users to have more control over the configuration of the loader

I believe that any moduleid that has an extension (except .js) should be loaded through a requirejs plugin. For example, `require('./some-file.json')` needs to be loaded through the JSON plugin, and `require('./some-file.css')` needs to be loaded through the text plugin. Quite a few packages `require` json files, so the CLI should be able to deal with those.

During the tracing step, whenever a moduleid is encountered that has an extension but no plugin loader prefix (e.g. `require('./my-file.json`), a `map` entry is added to the requirejs config. One example:

When the `my-package` dependency does an import of `require('my-file.json`)`, this `map` entry is added:
```
map: {
  '*': {
    'my-package/my-file.json': 'json!my-package/my-file.json'
  }
}
```

It uses the extension as the loader plugin. So if a package you use imports a json file, you need to make sure that the loader is set up to load json files. This resolves https://github.com/aurelia/cli/issues/588

Since quite a few packages import json files, I think we should make JSON a first class citizen by adding the json plugin to new projects by default. So i've added those changes to this PR, and we can decide whether to go through with that or not

I'd like to do a bit more testing before we merge this,

### DIFF
--- a/lib/build/amodro-trace/lib/parse.js
+++ b/lib/build/amodro-trace/lib/parse.js
@@ -122,7 +122,7 @@ define(['esprima', './lang'], function (esprima, lang) {
      * @returns {String} JS source string or null, if no require or
      * define/require.def calls are found.
      */
-    function parse(moduleName, fileName, fileContents, options) {
+    function parse(moduleName, fileName, fileContents, options, loaderConfig, context) {
         options = options || {};
 
         //Set up source input
@@ -167,6 +167,8 @@ define(['esprima', './lang'], function (esprima, lang) {
         }
 
         if (moduleDeps.length || moduleList.length) {
+            parse.registerPluginPaths(moduleDeps, loaderConfig, moduleName, context);
+
             for (i = 0; i < moduleList.length; i++) {
                 moduleCall = moduleList[i];
                 if (result) {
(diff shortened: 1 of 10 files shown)


EisenbergEffect:
Great work! Keep us posted on how testing turns out.

thomas-darling:
We _really_ need this feature - the lack of it is preventing us from migrating away from SystemJS :-(
And yep, JSON should definitely be a first-class citizen - it is so handy to be able to just import a json file :-) 

Any guess as to when this might be released?

JeroenVinke:
Hopefully we can get it in the next release.

lib/build/amodro-trace/lib/parse.js
```diff
@@ -1071,10 +1073,86 @@ define(['esprima', './lang'], function (esprima, lang) {
...
 
+    /**
+     * Adds map entries for paths that have specific extensions (currently only .json) to two requirejs configs:
+     * - the requirejs config passed to amodrotrace (so that the same map can be sent to the browser)
+     * - the requirejs config of the current context of amodrotrace
+     * 
+     * Example: require("./my-file.json") of the "my-package/index" module will result in:
+     * {
+     *  map: {
+     *    "*": {
```
zewa666: The * would indicate that this is a global, crossmodule path definition. Don't we know which module called the require and could therefore narrow down the setting to the proper module scope?

lib/build/amodro-trace/lib/parse.js
```diff
@@ -1071,10 +1073,86 @@ define(['esprima', './lang'], function (esprima, lang) {
...
+                var depMap = ctx.makeModuleMap(deps[i],
+                                ctx.makeModuleMap(parentModule),
+                                null,
+                                true);
+
+                var resolvedName = depMap.name;
+                var nameWithPluginPrefix = extension + '!' + depMap.name;
+
+                // add map to loader config that's passed to amodrotrace
+                if (!loaderConfig.map) loaderConfig.map = {};
```
zewa666: Have we checked somewhere that loaderConfig.map is always an object? What if the users sets it to `true` due to an mistake?

lib/build/amodro-trace/lib/parse.js
```diff
@@ -1071,10 +1073,86 @@ define(['esprima', './lang'], function (esprima, lang) {
...
+                if (!loaderConfig.map['*']) loaderConfig.map['*'] = {};
```
zewa666: this could fail if `map` is not typeof object

lib/build/amodro-trace/lib/parse.js
```diff
@@ -1071,10 +1073,86 @@ define(['esprima', './lang'], function (esprima, lang) {
...
+        return null;
+    };
+
+    /**
+     * Returns whether or not the path has a plugin prefix
+     * E.g. json!some-file.json -> true
+     * some-file.json -> false
+     */
+    parse.hasPrefix = function (path) {
+        var index = path ? path.indexOf('!') : -1;
```
zewa666: often people, including me, coming from JSPM or the likes, put the bang as a suffix. so above example would be `some-file.json!json`. Should the hasPrefix method potentially check for the pattern `[prefix]BANG[filename]`?

lib/build/bundler.js
```diff
@@ -26,6 +26,7 @@ exports.Bundler = class {
     this.loaderConfig = {
       baseUrl: project.paths.root,
       paths: ensurePathsRelativelyFromRoot(project.paths || {}),
+      map: project.map || {},
```
zewa666: what if map here is defined not as an object?

JeroenVinke:
I'll have to think about this one again. Since we're about to add SystemJS to the CLI's capabilities, we need to make sure that it would work with SystemJS as well.

thomas-darling:
Any new on this?

JeroenVinke:
Not yet. Perhaps it would be better to transform files that do this before we construct the bundle. There are some tasks I need to complete first before I can look into this again.

JeroenVinke:
Closing this until we have another go at it

PR #597, opened 2017-04-10
create a configuration node for BrowserSync at task/run.js

casion:
Added a node in aurelia.json for BrowserSync Configuration.
`
  "serve": {
    "localPort": 9000,
    "logLevel": "silent"
  },
`

### DIFF
--- a/lib/commands/new/project-template.js
+++ b/lib/commands/new/project-template.js
@@ -28,6 +28,8 @@ exports.ProjectTemplate = class {
       paths: {}
     });
 
+    this.model.serve = this.createServeConfig();
+
     this.postInstallProcesses = [];
 
     this.root = options.hasFlag('here')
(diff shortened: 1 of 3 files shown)


lib/commands/new/project-template.js
```diff
@@ -28,6 +28,8 @@ exports.ProjectTemplate = class {
       paths: {}
     });
 
+    this.model.serve = this.createServeConfig();
```
JeroenVinke: I think we should put the port and logLevel in the `platform` object, similar to `baseDir`

lib/commands/new/project-template.js
```diff
@@ -64,6 +66,13 @@ exports.ProjectTemplate = class {
     );
   }
 
+  createServeConfig(){
+    return {
+      localPort:9000,
```
JeroenVinke: can you make sure that the linter is OK with this? I think a space is needed between `localPort:` and `9000`

lib/commands/new/project-template.js
JeroenVinke: Perhaps a better place to set this would be in [create](https://github.com/casion/cli/blob/6fbfb3daf6d688da69dce300cbf9fc97b857a1da/lib/commands/new/project-template.js#L272) where we set `baseDir` as well

JeroenVinke:
Closing this for now. We've added webpack support since this was added, so we'd have to do this only for Requirejs and systemjs projects

Issue #562, opened 2017-03-22
Ability to specify Uglify options


zedL:
**I'm submitting a feature request**

* **Library Version:**
aurelia-cli@0.26.1

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
7.7.2

* **NPM Version:**
4.3

* **Language:**
TypeScript 2.2.1

**Current behavior:**
I use uglify-js@2.8.14 and get following error when I build with `au build --env stage`
```
Writing mdl-bundle.js...
WARN: Output exceeds 32000 characters
Writing vendor-bundle.js...
```
My options in aurelia.json are
```
"options": {
      "minify": "stage & prod",
      "sourcemaps": "dev & stage",
      "rev": "prod"
    },
```
Because uglify write the warning in err output my CI build give up with an error.

With uglify-js@2.7.5 it works like expected.

Under the [Beautifier-Options](https://www.npmjs.com/package/uglify-js#beautifier-options)
is a `max-line-len` option that points to this problem.

But how can I set the uglify options when I use aurelia CLI? There is no way of doing it.
I found out that it is hard coded [here](https://github.com/aurelia/cli/blob/45255cb692d4669d7690541eed19251265401fb4/lib/build/bundle.js#L239)

**Expected/desired behavior:**
-- higher default warning level or disabled option ([set to zero](https://github.com/mishoo/UglifyJS/tree/2bc1d02363db3798d5df41fb5059a19edca9b7eb#usage))
-- ability to set uglify option

**What is the motivation / use case for changing the behavior?**
I want to use continiouse integration without warnings/errors.

ariovistus:
for the sake of anyone stumbling on this from google, the way to turn off the warning (at least on 0.30.1 with uglifyjs 2.8.29) is

```
"options": {
  "minify": {
    "prod": {
      "output": {
        "max_line_len": 1000000
      }
    }
  }
}
```

diegohb:
@ariovistus - thanks for summing it up.. 

replacing "stage & prod" with this object with options.. im wondering how to still enable minify for stage per the original setting "stage & prod"... I don't see anywhere in the cli code that it parses those into configured environments..

is it maybe an empty object? 
```
 "minify": {
                "prod": {
                  "output": {
                    "max_line_len": 1000000
                  }
                },
               "stage": { }
         }
```

JeroenVinke:
[...]

PR #602, opened 2017-04-12
feat(build): allow minify options to be supplied

JeroenVinke:
closes https://github.com/aurelia/cli/issues/562

Currently the `build.options` section of `aurelia.json` is a bit limited, because you can only use strings and booleans as the option values. So in order to properly support https://github.com/aurelia/cli/issues/562 and future expansion, I decided to do a bit of refactoring to allow you to provide objects as well, allowing you to do this:

```
    "options": {
      "sourcemaps": "dev & stage",
      "minify": {
        "env": "dev & stage & prod",
        "options": {
          "max-line-len": 100000
        }
      },
      "rev": true
    },
```

It is backwards compatible with older versions of the CLI (and projects built by older versions of the CLI), so 

```
    "options": {
      "sourcemaps": "dev & stage",
      "minify": "stage & prod",
      "rev": true
    },
```

will still work.

### DIFF
--- a/lib/build/build-options.js
+++ b/lib/build/build-options.js
@@ -1,0 +1,60 @@
+'use strict';
+
+const CLIOptions = require('../cli-options').CLIOptions;
+
+exports.BuildOptions = class {
+  constructor(options, defaultOptions, environment) {
+    this.options = Object.assign({}, defaultOptions, options);
+    this.environment = environment || CLIOptions.getEnvironment();
+  }
+
+  getAllOptions() {
+    return this.options;
+  }
+
+  getValue(propPath) {
+    let split = propPath.split('.');
+    let cur = this.options;
+
+    for (let i = 0; i < split.length; i++) {
+      if (!cur) {
+        return undefined;
+      }
+
+      cur = cur[split[i]];
+    }
+
+    return cur;
+  }
+
+  isApplicable(propPath) {
+    let value = this.getValue(propPath);
+
+    if (!value) {
+      return false;
+    }
+
+    if (typeof value === 'boolean') {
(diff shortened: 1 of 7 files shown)


AStoker:
Awesome!
Curious, how much work would it be to expand this just a bit to allow different options for different environments?

JeroenVinke:
@AStoker interesting thought. Any idea what the config would look like?

AStoker:
Perhaps something like this.
```
"options": {
      "sourcemaps": "dev & stage",
      "minify": {
        "dev": {
          "max-line-len": 100000
        },
        "stage & prod": {
          "max-line-len": 10000
        }
      },
      "rev": true
},
```
Or, perhaps if there are multiple kinds of options needed, each environment configuration's object would have it's own configuration language.
```
"options": {
      "sourcemaps": "dev & stage",
      "minify": {
        "dev": {}, //Default
        "stage & prod": {
             "otherConfigNeeded": {
                  "some-prop": 1234
             },
        }
      },
      "rev": true
},
```

JeroenVinke:
I like that. We just need to be consistant by using the same syntax for all options (not just minify)

```
"sourcemaps": "dev & stage",
```
should be equivalent to:

```
"sourcemaps":  {
   "dev": true,
   "stage": true,
   "prod": "false"
 }
```

AStoker:
Absolutely.

JeroenVinke:
Great 👍  i'll close this PR and open up a new one

ghidello:
Sounds very nice to me too and I've already a couple of Ideas for using it. So.. Thanks in advance!

PR #613, opened 2017-04-27
chore(gitignore): remove no related project files

RomainLanz:
Hey everyone !

`.idea` is editor related and `.DS_STORE` is OS related. Both of those files/folders should be inside [a global gitignore file](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) since they are not related to the project.

`scripts` folder is recreate by the CLI when you bundle your application, there's no need to push compiled file so it's better to remove them by default.

### DIFF
--- a/lib/resources/content/gitignore
+++ b/lib/resources/content/gitignore
@@ -1,7 +1,7 @@
 # You may want to customise this file depending on your Operating System
 # and the editor you use.
 #
-# We recommend that you use a Global Gitignore files that is not related
+# We recommend that you use a Global Gitignore for files that is not related
 # to the project. (https://help.github.com/articles/ignoring-files/#create-a-global-gitignore)
 
 # OS
--- a/lib/resources/content/gitignore
+++ b/lib/resources/content/gitignore
@@ -1,3 +1,2 @@
 node_modules
-.idea
-.DS_STORE
+scripts
--- a/lib/resources/content/gitignore
+++ b/lib/resources/content/gitignore
@@ -1,2 +1,27 @@
+# You may want to customise this file depending on your Operating System
+# and the editor you use.
+#
+# We recommend that you use a Global Gitignore files that is not related
+# to the project. (https://help.github.com/articles/ignoring-files/#create-a-global-gitignore)
+
+# OS
+#
+# Ref: https://github.com/github/gitignore/blob/master/Global/macOS.gitignore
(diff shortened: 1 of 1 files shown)


JeroenVinke:
Hey @RomainLanz, thanks for this PR!

Unless there is a particular reason, i'd be in favor of adding the `scripts` folder to the gitignore as well. But i'm not sure about the editor configs. I can see how some people would like to have editor configs in the repository, but some don't.

RomainLanz:
I don't have editor configs in my repository but they are not in the project `.gitignore` since it's not related to the project itself. That's why you can create [a global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) to ignore all files related to your computer/editor (`Thumb.db`, `.DS_STORE`, `.idea`, `.vscode`, ...)

AStoker:
I'm with @JeroenVinke with adding the scripts to the list, makes sense.
In regards to the other files, I'm inclined to leave them in there, unless there's a specific reason to remove them (are they causing a problem?). I don't think it hurts having it (potentially) in two places (global and local), and it's convenient that it's just added to the project (don't have to go modifying a global). I actually usually check in my `.vscode` folder since there's things in there I want across a few private repos to keep workstations similar.

RomainLanz:
Well if you decide to keep editor/computer related files why not add `Thumb.db`, `.DS_STORE`, `.idea`, etc. ?

AStoker:
I'm personally not opposed to it.
Many people don't bother with adding a global git excludes, and so while having it here would be redundant, I see it helping more by leaving it here than by removing it. @JeroenVinke, what are your thoughts?

JeroenVinke:
> Many people don't bother with adding a global git excludes, and so while having it here would be redundant, I see it helping more by leaving it here than by removing it. 

Since the CLI is all about getting a working setup quickly (that you can push to github right away), I'd like to keep the editor/OS related files in the gitignore., so I agere with @AStoker. @RomainLanz could you add those editor/computer related files you mentioned, in addition to the `scripts` folder you already added?

RomainLanz:
Yes, so which file should I add? `.vscode`, `.idea`, `Thumb.db`, `.DS_STORE`?

Should I also write a little comment on the file linking to [this page](https://github.com/github/gitignore/tree/master/Global) if they want an advanced configuration?

JeroenVinke:
That would be good 👍

RomainLanz:
Okay.

I have updated the file, what do you think? @JeroenVinke @AStoker

PR #616, opened 2017-04-29
doc(readme): add instructions to run CLI locally

JeroenVinke:
closes https://github.com/aurelia/cli/issues/451

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,3 +13,16 @@ This library is part of the [Aurelia](http://www.aurelia.io/) platform and conta
 ## Documentation
 
 You can read documentation on the cli [here](http://aurelia.io/hub.html#/doc/article/aurelia/framework/latest/the-aurelia-cli). If you would like to help improve this documentation, the source for the above can be found in the doc folder within the framework repository.
+
+## Building
+1. Clone the aurelia-cli: `git clone https://github.com/aurelia/cli.git`
+2. Go into the cli directory: `cd cli`
+3. Run `npm install`
+4. Link the cli with: `npm link`
+5. Still in the cli directory, run `npm install git+https://git@github.com/gulpjs/gulp.git#4.0`
+6. Also in the cli directory, run `npm install babel-polyfill babel-register typescript`
+7. Create a new project with `au new` or use an existing project. The linked CLI will be used to create the project.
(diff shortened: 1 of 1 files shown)


s-hoff:
For testing with the TS setup one would need the `typescript` package.
esprima also no longer needs to be installed this way as it is already in the devDependencies.
I'm a little curious about gulp: Wouldn't it be better to have the same version in the cli's devDependencies as in the projects set up by it?

JeroenVinke:
@s-hoff thanks for looking it over! I have removed esprima and added typescript. Regarding using gulp 4.0 as the devDependency, it's a good idea, but considering all other libraries are on gulp 3.9.1 I think we should wait and move all libraries over to gulp 4 when that's released.

 I tried using gulp 4 and ran into this beauty:

```
PS C:\Development\cli> gulp prepare-release
[10:06:36] Using gulpfile C:\Development\cli\gulpfile.js
[10:06:36] Starting 'prepare-release'...
[10:06:36] 'prepare-release' errored after 4.11 ms
[10:06:36] TypeError: gulp.hasTask is not a function
    at C:\Development\cli\node_modules\run-sequence\index.js:19:22
    at Array.forEach (native)
    at verifyTaskSets (C:\Development\cli\node_modules\run-sequence\index.js:13:11)
    at runSequence (C:\Development\cli\node_modules\run-sequence\index.js:92:2)
    at C:\Development\cli\build\tasks\prepare-release.js:25:10
    at taskWrapper (C:\Development\cli\node_modules\undertaker\lib\set-task.js:13:15)
    at bound (domain.js:280:14)
    at runBound (domain.js:293:12)
    at asyncRunner (C:\Development\cli\node_modules\async-done\index.js:36:18)
    at _combinedTickCallback (internal/process/next_tick.js:73:7)
```

Indicating that run-sequence does not work with gulp 4 (yet)

s-hoff:
Sounds good to me.

ghidello:
Accordingly to [this](https://github.com/OverZealous/run-sequence#please-note) comment I don't think that the run-sequence package will be updated to gulp 4. We'll have to rewrite the pre-release task using the new gulp.series() api when we'll prepare for the jump to gulp 4.

PR #617, opened 2017-04-29
chore(circleci): resolve minimatch error and tests running on linux

JeroenVinke:
should allow circleCI to run the tests correctly

### DIFF
--- a/spec/lib/file-system.spec.js
+++ b/spec/lib/file-system.spec.js
@@ -115,10 +115,6 @@ describe('The file-system module', () => {
         return fs.readdir(writeDir + readDir);
       }).catch(fail).then(done);
     });
-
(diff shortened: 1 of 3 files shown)


spec/lib/file-system.spec.js
```diff
@@ -116,9 +116,9 @@ describe('The file-system module', () => {
       }).catch(fail).then(done);
     });
 
-    it('rejects if mkdirp returns an error', done => {
-      pending(`change file-system to mkdirp.mkdirp or more helpers.`);
-    });
+    // it('rejects if mkdirp returns an error', done => {
```
zewa666: if the test wasn't yet in place, I'd actually remove it vs commenting it out

PR #621, opened 2017-05-08
chore(dev-dependency): update to gulp4

ghidello:


ghidello:
I had to remove `bower.json` from the `bump-version` task because otherwise gulp4 src method breaks complaining that the file isn't present. I've been able to execute the prepare-release without issues: I also bumped my local CLI to 1.0.0 😄

JeroenVinke:
Other Aurelia libraries are still on gulp 3. I'd feel better to update the CLI's gulp to v4 when we update the rest too.

Issue #619, opened 2017-05-05
0.29.0 html-minifier breaks inline 'less than' ( < ) comparison operator


BiggAdd:
**I'm submitting a bug report**

* **Library Version:**
0.29.0

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
7.3.0

* **NPM Version:**
3.10.10

* **Browser:**
all

* **Language:**
ESNext

**Current behavior:**
Inline javascript containing a `<` breaks the html minify build step

`
    <div>${1 < 10 ? 'foo' : 'bar'}</div>
`

Javascript defined inside of an attribute tag compiles correctly.

It seems that the html-minifier node module is the culprit as bypassing this build step fixes the issue.

JeroenVinke:
Thank you for filing this issue @BiggAdd. The changes to the first file in https://github.com/JeroenVinke/cli/commit/860bedee01dee66a183c54b200f831099fa30692 should resolve your issue

gheoan:
Surprisingly, that is not valid HTML. Use `<div>${1 &lt 10 ? 'foo' : 'bar'}</div>` instead. The question is if we should workaround it or not.

BiggAdd:
Its not valid HTML, but the whole point of using the ${} notation is to treat anything inside it as an expression, like ES6 template literals, so in fact ${ 1 < 10 ? ...  } should be supported. 

I've not had a chance to test it, but it looks like JeroenVinke has pushed a fix for it.

thomas-darling:
@JeroenVinke I believe this fix breaks the bundler - please see https://github.com/aurelia/bundler/issues/163

PR #622, opened 2017-05-09
fix(html-minify): ignore interpolation expressions

JeroenVinke:
…essions

closes https://github.com/aurelia/cli/issues/619

### DIFF
--- a/lib/resources/tasks/process-markup-minify-max.js
+++ b/lib/resources/tasks/process-markup-minify-max.js
@@ -1,11 +1,14 @@
 import gulp from 'gulp';
 import htmlmin from 'gulp-htmlmin';
+import plumber from 'gulp-plumber';
+import notify from 'gulp-notify';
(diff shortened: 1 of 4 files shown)


lib/resources/tasks/process-markup-minify-max.js
```diff
@@ -1,11 +1,14 @@
 import gulp from 'gulp';
 import htmlmin from 'gulp-htmlmin';
+import plumber from 'gulp-plumber';
```
zewa666: are those two dependencies in the package.json? Couldn't find them.

zewa666:
Also make sure to fix the commit message, the first line is exceeding the length as such it wont be properly rendered in our release-notes

lib/resources/tasks/process-markup-minify-max.js
JeroenVinke: Yep 👍 it gets added to the package.json of the generated project

Issue #583, opened 2017-04-02
Path must be a string error when package.json of dependency has no main property


JeroenVinke:
**I'm submitting a bug report**

* **Library Version:**
0.27.0

**Please tell us about your environment:**
* **Operating System:**
OSX 10.x|Linux (distro)|Windows [7|8|8.1|10]

* **Node Version:**
any

* **NPM Version:**
any

**Current behavior:**
When a package has a package.json file, but no `main` property, it throws the following error during build:

```
PS C:\app\ui> au build
Starting 'copyAdditionalResources'...
Finished 'copyAdditionalResources'
Starting 'readProjectConfiguration'...
{ uid: 2,
  name: 'readProjectConfiguration',
  branch: false,
  error:
   TypeError: Path must be a string. Received undefined
       at assertPath (path.js:7:11)
       at Object.join (path.js:468:7)
       at determineLoaderConfig (C:\app\ui\node_modules\aurelia-cli\lib\build\package-an
alyzer.js:57:23)
       at loadPackageMetadata.then (C:\app\ui\node_modules\aurelia-cli\lib\build\package
-analyzer.js:15:19),
  duration: [ 0, 26527483 ],
  time: 1491140511846 }
{ uid: 0,
  name: '<series>',
  branch: true,
  error:
   TypeError: Path must be a string. Received undefined
       at assertPath (path.js:7:11)
       at Object.join (path.js:468:7)
       at determineLoaderConfig (C:\app\ui\node_modules\aurelia-cli\lib\build\package-an
alyzer.js:57:23)
       at loadPackageMetadata.then (C:\app\ui\node_modules\aurelia-cli\lib\build\package
-analyzer.js:15:19),
  duration: [ 0, 35025237 ],
  time: 1491140511850 }
TypeError: Path must be a string. Received undefined
    at assertPath (path.js:7:11)
    at Object.join (path.js:468:7)
    at determineLoaderConfig (C:\app\ui\node_modules\aurelia-cli\lib\build\package-analy
zer.js:57:23)
    at loadPackageMetadata.then (C:\app\ui\node_modules\aurelia-cli\lib\build\package-an
alyzer.js:15:19)
```

**Expected/desired behavior:**

It should recover from the situation because the main file is could also be defined in aurelia.json.

PR #625, opened 2017-05-10
fix(package-analyzer): infer index.js as main

JeroenVinke:
…roperty

fixes https://github.com/aurelia/cli/issues/583

### DIFF
--- a/lib/build/package-analyzer.js
+++ b/lib/build/package-analyzer.js
@@ -63,18 +63,24 @@ function determineLoaderConfig(project, description) {
         sourcePath = path.join(location, metadata.main);
       }
     } else {
-      sourcePath = path.join(location, metadata.main);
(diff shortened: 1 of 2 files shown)


lib/build/package-analyzer.js
```diff
@@ -63,7 +63,7 @@ function determineLoaderConfig(project, description) {
         sourcePath = path.join(location, metadata.main);
       }
     } else {
-      sourcePath = path.join(location, metadata.main);
+      sourcePath = path.join(location, metadata.main || 'index.js');
```
zewa666: what if even index.js is not present? How about aborting with an error message and giving up? It generally shouldn't happen because index.js is really the kinda de-facto standard, but still you know those legacy libraries don't always play nice ;)

JeroenVinke:
Great idea @zewa666. I've made these improvements

Issue #198, opened 2016-06-28
jspm support


dpinart:
Is there any way to create a new project with the cli that uses jspm/systemjs as module loader?

EisenbergEffect:
Not at this time, no.

Consider that jspm and system.js are two separate things. We will probably add support for system.js first and consider jspm at a later date. jspm is the most problematic of all build options due to its stability and issues within enterprises. If you want jspm now, the best way to go is use our skeletons or set up a custom solution.

XuJinNet:
System.js +1

JeroenVinke:
### SystemJS support

I think the best way to start working on this is to first create a new aurelia-cli project (with requirejs) and modify that project to work with systemjs. You'll probably want to follow [these instructions](https://github.com/aurelia/cli/pull/616/files) to be able to make changes to the CLI easily. As soon as the project is working with SystemJS, we can teach the project creation part of the CLI to create this exact setup.

The CLI has both a bundler and a loader. For the bundling part we use [amodro-trace](https://github.com/amodrojs/amodro-trace) which uses RequireJS internally to trace dependencies of an app. What is created by the bundler is a bundle that requirejs as well as systemjs should be able to load. So we don't need to modify the bundler part of the CLI, only the loader that is used. 

I would start by editing the aurelia.json file ([require -> systemjs](https://github.com/JeroenVinke/aurelia-cli-faker/blob/master/aurelia_project/aurelia.json#L66), [empty plugins array](https://github.com/JeroenVinke/aurelia-cli-faker/blob/master/aurelia_project/aurelia.json#L69-L78) as they are requirejs specific, remove [text plugin dependency](https://github.com/JeroenVinke/aurelia-cli-faker/blob/master/aurelia_project/aurelia.json#L121), remove [this line](https://github.com/JeroenVinke/aurelia-cli-faker/blob/master/aurelia_project/aurelia.json#L97)).

The way the current setup works is that the loader is in the vendor-bundle.js, which gets loaded through a script tag in index.html. So we need to make sure that the current bundler can bundle SystemJS, and if not what is preventing it from being bundled. The vendor-bundle is [configured](https://github.com/JeroenVinke/aurelia-cli-faker/blob/master/aurelia_project/aurelia.json#L67) via aurelia.json to contain the loader configuration which can be found at the bottom of vendor-bundle.js:

[...]

PR #631, opened 2017-05-20
feat(all): systemjs support

simonfox:
Enables support for SystemJS as module loader including new project setup via `au new`.

closes aurelia/cli#198


simonfox:
I've used this to convert my jspm/systemjs project that has approx 380 js source files and 49 package.json deps to a cli project using systemjs loader....ended up being relatively painless!!

The only issue I found was that dependencies that define a `resources` section that includes `html` files do not load correctly if included in the `vendor-bundle` (a 404 is generated for the html file even though the requested url maps to a module in the bundle). They load fine when added to a separate _plugins_ bundle or even the `app-bundle`.

I'll look into this next week, but I don't think it is a deal breaker.

EisenbergEffect:
A tremendous contribution!

JeroenVinke:
I like how you you kept the amount of changes to the cli to a minimum. Using loader specific project configurations in the `new` folder is exactly what I am doing for the webpack integration :)

I will do some testing and will report back

JeroenVinke:
We do need to look at the issue with html resources in the vendor bundle as this is a common use case. Let me know if you need help debugging.

simonfox:
@JeroenVinke will do...I'm away for next couple of days but will get back to it once I'm back.

lib/build/amodro-trace/write/stubs.js
```diff
@@ -12,7 +13,9 @@ module.exports = function stubs(options) {
   options = options || {};
 
   return function(context, moduleName, filePath, contents) {
-    if (options.stubModules && options.stubModules.indexOf(moduleName) !== -1) {
+    if (options.stubModules
```
AStoker: Do we have a PR for Amodro-Trace to update to this? We've updated this library multiple times to support features we need, I'm wondering if we should just officially fork this to do our own loading, or we should continue to open PR's for the changes we make? @JeroenVinke, do you have an opinion?

lib/commands/new/loaders/require.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+const ProjectItem = require('../../../project-item').ProjectItem;
+
+module.exports = function(project) {
+  project.loader = project.model.loader.id;
+  project.model.loader = undefined; // remove loader from model as it is actually a property of model.build
```
AStoker: Do we need that property on model then? Is it just noise that isn't actually used?

lib/build/amodro-trace/write/stubs.js
simonfox: Yes I wasn't sure whether that was necessary....this change was made to allow the plugin to be defined as a package in the bundle dependencies - see [the config for systemjs-plugin-text](https://github.com/aurelia/cli/pull/631/commits/a818cd8591daf5ac73908f6b108ddbf4e53e8bb2#diff-239f8d9c8592bbc58e040901f2100f2bR8)

lib/commands/new/loaders/require.js
simonfox: The model is setup based on the result from WorkflowEngine [here](https://github.com/aurelia/cli/pull/631/files#diff-90607133b006fdb0120c6a67bb7ae68aR26). I remove that property so there is no loader node at the top level of `aurelia.json`. Make sense?

lib/build/loader.js
```diff
@@ -76,8 +77,38 @@ exports.createRequireJSConfig = function createRequireJSConfig(platform, bundler
   return config;
 };
 
-exports.createSystemJSConfig = function createSystemJSConfig(platform) {
-  throw new Error('SystemJS is not yet supported');
+exports.createSystemJSConfig = function createSystemJSConfig(platform, bundler) {
+  let loaderOptions = bundler.loaderOptions;
```
zewa666: I don't want to be nitpicky, but most of those lets could easily be consts

lib/build/loader.js
```diff
@@ -76,8 +77,38 @@ exports.createRequireJSConfig = function createRequireJSConfig(platform, bundler
...
+  let bundles = bundler.bundles;
+  let configName = loaderOptions.configTarget;
+  let includeBundles = shouldIncludeBundleMetadata(bundles, loaderOptions);
+  let bundleMetadata = {};
+  let config = {
+    map: {}
+  };
+  let location = platform.baseUrl || platform.output;
+  
+  for (let i = 0; i < bundles.length; ++i) {
```
zewa666: this section does work, but switching it to a functional method chaining -> bundles.map.filter.reduce might make this section more readable. Although this is definitely a nitpick ;)

lib/commands/new/loaders/require.js
zewa666: yeah how about `delete project.model.loader`?

simonfox: good idea..thanks @zewa666 !! will sort when I get back to this tomorrow.

simonfox:
@JeroenVinke right I've found the issue with html resources in the vendor bundle. When aurelia/loader-default is instantiated it sets the [System registry entry for `text` to the inlined text plugin here](https://github.com/aurelia/loader-default/blob/eb18cd43fae48125f73f14548e1125952605f678/src/index.js#L198). That means that the registry key `text` is set and [future requests for the `text` plugin will use that module](https://github.com/systemjs/systemjs/blob/978cc37fb6d83afde8de45d345dcae4baebafeac/src/resolve.js#L201).

When a package with an html resource is included in the vendor bundle, it is registered with SystemJS before the aurelia loader has been setup so there is not yet a registry entry for `text` so [back in SystemJS `coreResolve` resolution continues](https://github.com/systemjs/systemjs/blob/978cc37fb6d83afde8de45d345dcae4baebafeac/src/resolve.js#L201) and `text` gets finalized as `baseUrl + text` (so in the default config of a new cli project - `http://localhost:9000/text`). Which does not map to a module name in the bundle. So for an html resource like `aurelia-autocomplete/autocomplete.html`, the final key becomes `aurelia-autocomplete/autocomplete.html!http://localhost:9000/text` when it needs to be `aurelia-autocomplete/autocomplete.html!http://localhost:9000/text`.

Does that make sense?

If the 'text' key is registered with SystemJS before the html resources are defined, we can get around this (knowing that `loader-default` will replace the module at that key when it is created). So if I add `SystemJS.set('text', SystemJS.newModule({}));` to the `_aureliaConfigureModuleLoader` function then it all works again. That feels a little dirty though...thoughts? any other ideas?

JeroenVinke:
Great find @simonfox. I was able to confirm it. 

I was wondering why it would work for the skeletons, because the concept is the same. The skeleton also registers all views with systemjs before it loads aurelia-loader-default. I noticed in `config.js` that it has a map entry for the text loader, which we do not. After I added `"text": "text"` to the `map` object of the systemjs config, it loads as expected.

JeroenVinke:
Also, the systemjs option is kind of hidden in `au new`, which does not feel right ;). I'd like to see if it can be more prominently visible, keeping in mind that we'll also have a webpack option soon.

simonfox:
@JeroenVinke oh man, I'm sure I tried a map!!!!! must have had something else wrong at the same time. Thank you!

With `au new` did you just want to bring the loader selection step further forward i.e. after the transpiler option instead of right at the end? or did you mean by using colors or something like that?

JeroenVinke:
Currently you have to select "custom" if you want to use systemjs, right? I was thinking if it would be better to have the user select a loader first, followed by the "default esnext", "default typescript", "custom" question.

When we add webpack, you would be able to pick requirejs, systemjs or webpack as one of the first questions. What do you think?

simonfox:
@JeroenVinke sounds good!

lib/build/amodro-trace/write/stubs.js
simonfox: @JeroenVinke did you have thoughts on this?

simonfox:
@JeroenVinke have fixed the review comments and made the changes discussed...

lib/build/amodro-trace/write/stubs.js
JeroenVinke: I've asked the guy behind Amodrotrace if he'd be open to us starting pull requests for features that we need

JeroenVinke:
Thank you. Do all commands work, e.g. `au test` ?

simonfox:
@JeroenVinke yes all working.

`au test` just works without needing to change any of the test config because global require/define has been replaced by SystemJS.amdRequire/amdDefine.

I did notice that with the default config from `au new`, `paths.resources` and `au generate` don't align with the default folder structure of `src/resources`. But that is another issue.

JeroenVinke:
Perfect! The issue you're mentioning is fixed in master, I think you need to rebase onto master to get this fix.

I'm getting the following error while running `au new --here` and select ASP.NET Core:

```
Error locating resource: content/index.html
{ Error: Cannot find module './content/index.html'
    at Function.Module._resolveFilename (module.js:470:15)
    at Function.resolve (internal/module.js:27:19)
    at exports.locateResource (C:\Development\cli\lib\resources\index.js:5:20)
    at Function.resource (C:\Development\cli\lib\project-item.js:180:45)
    at Object.configureVisualStudioStructure (C:\Development\cli\lib\commands\new\project-template.js:80:21)
    at module.exports (C:\Development\cli\lib\commands\new\platforms\aspnetcore.js:4:11)
    at ui.clearScreen.then.then.then.answer (C:\Development\cli\lib\workflow\activities\project-create.js:41:11)
    at process._tickCallback (internal/process/next_tick.js:109:7) code: 'MODULE_NOT_FOUND' }
```

It looks like [this file](https://github.com/simonfox/cli/blob/bf03b5069169f6007470d6480a41a205484521d8/lib/commands/new/project-template.js#L80) needs to be changed to either `index.requirejs.html` or `index.systemjs.html`

After that, I think we're good to go!

simonfox:
@JeroenVinke sorry about that, missed that branch of `au new`. Should be good now...

Issue #293, opened 2016-08-26
Error in buildTypeScript: A project cannot be used in two compilations at the same time


hanssens:
**I'm submitting a bug report**
- **Library Version:**
  0.18.0 (CLI)

**Please tell us about your environment:**
- **Operating System:**
  OSX 10.11.6
- **Node Version:**
  6.1.0
- **NPM Version:**
  3.8.6
- **Browser:**
  all
- **Language:**
  TypeScript 1.8.10

**Current behavior:**
Frequently, when running `au run --watch` the TypeScript compile seems to hang. 

In the Activity Monitor I can see that the aurelia process suddenly keeps using lots of CPU:

![activity_monitor__all_processes_](https://cloud.githubusercontent.com/assets/493382/18006431/b93465fa-6b9f-11e6-9956-c4e42e9d0702.png)

Also, the Terminal window seems to hang. Meaning I'm unable to kill the process with Ctrl+C, where it leads to a long stacktrace like:

```
File Changed: src/views/devices/device-overview.ts
Starting 'readProjectConfiguration'...
Starting 'readProjectConfiguration'...
Finished 'readProjectConfiguration'
Finished 'readProjectConfiguration'
Starting 'processMarkup'...
Starting 'processCSS'...
Starting 'processMarkup'...
Starting 'processCSS'...
Starting 'configureEnvironment'...
Starting 'configureEnvironment'...
Finished 'configureEnvironment'
Starting 'buildTypeScript'...
Finished 'configureEnvironment'
Starting 'buildTypeScript'...
{ uid: 292,
  name: 'buildTypeScript',
  branch: false,
  error:
   { Error: gulp-typescript: A project cannot be used in two compilations at the same time. Create multiple projects with createProject instead.
       at compile (/Users/Hanssens/Work/git/xxx/app/node_modules/gulp-typescript/release/main.js:72:19)
       at buildTypeScript (/Users/Hanssens/Work/git/xxx/app/aurelia_project/tasks/transpile.ts:32:15)
       at bound (domain.js:280:14)
       at runBound (domain.js:293:12)
       at asyncRunner (/Users/Hanssens/Work/git/xxx/app/node_modules/async-done/index.js:36:18)
       at _combinedTickCallback (internal/process/next_tick.js:67:7)
       at process._tickDomainCallback (internal/process/next_tick.js:122:9)
     domain:
      Domain {
        domain: null,
        _events: {},
        _eventsCount: 0,
        _maxListeners: undefined,
[...]

Issue #265, opened 2016-08-08
[Feature Request] 'watch' option for the 'build' task


avrahamcool:
**I'm submitting a feature request**
It will be helpfull if we could use `watch` also in the `build` command. (currently available only for `run`)
- **Library Version:**
  v0.17.0
- **Operating System:**
  Windows 10
- **Node Version:**
  v4.4.7
- **NPM Version:**
  3.8.9
  
  
- **Browser:**
  all
- **Language:**
  all

**Current behavior:**
we can't use the `watch` option in the build command.
so changes in the code don't reflect back to the app.

**Expected/desired behavior:**
- **What is the expected behavior?**
  allow the `watch` option also for the `build` command, thus allowing the developer to choose how to serve his app, while gaining the benefits of synchronization between the code and the app.
  the `watch` should behave just like it does in the `run` command. (listen to changes in the code, and rebundle the affected files).
  [the developer will need to manualy refresh the app in the browser to see the changes, because `browserSync` will not be active]
- **What is the motivation / use case for changing the behavior?**
  the motivation behind this request is that we already have a server who takes care of serving the app. this server also server a webApi on the same address as the web project.
  so we need to be able to build our code in such a way that let us choose our own methode of serving the app, while actively listening to the changes in the code.

FYI:
as a workaround, we are currently using `run --watch`, so the app is served by the `aurelia-cli` with `watch` on his default port, and then we also serve the same app with our server on our port.
but we really could use an option to `build` with `watch` without `serve`.

zamotany:
It's a recuring theme here. `aurelia-cli` when executing tasks, searches for them in `aurelia_project/tasks`, so you can tailor them for your own needs. Those task are just simple `gulp` task.
For example to add `--watch` option to `build` command replace `build.js` with this:
```
import gulp from 'gulp';
import transpile from './transpile';
import processMarkup from './process-markup';
import processCSS from './process-css';
import {build} from 'aurelia-cli';
import project from '../aurelia.json';
import {CLIOptions} from 'aurelia-cli';

function readProjectConfiguration() {
  return build.src(project);
}

function writeBundles() {
  return build.dest();
}

function onChange(path) {
  console.log(`File Changed: ${path}`);
}

let buildTask = gulp.series(
  readProjectConfiguration,
  gulp.parallel(
    transpile,
    processMarkup,
    processCSS
  ),
  writeBundles
);

let watch = function() {
  gulp.watch(project.transpiler.source, buildTask).on('change', onChange);
[...]

PR #632, opened 2017-05-25
fix/enhancement(cli): improve --watch flag

jwx:
The improved --watch flag causes fewer crashes, performs better and is now also available on the build command.

Closes aurelia/cli#293 and aurelia/cli#265.


lib/cli-options.js
```diff
@@ -5,6 +5,13 @@ exports.CLIOptions = class {
     exports.CLIOptions.instance = this;
   }
 
+  taskName() {
```
JeroenVinke: Would you mind writing some tests for this?

lib/resources/tasks/watch.ts
```diff
@@ -0,0 +1,100 @@
...
+import transpile from './transpile';
+import processMarkup from './process-markup';
+import processCSS from './process-css';
+import copyFiles from './copy-files';
+
+const debounceWaitTime = 100;
+let isBuilding = false;
+let pendingRefreshPaths = [];
+let watches = {};
+let watchCallback = () => { };
```
JeroenVinke: perhaps `noop` is a better name

lib/resources/tasks/run.ts
```diff
@@ -32,42 +24,31 @@ let serve = gulp.series(
...
 
-    //see if there are static files to be watched
-    if (typeof project.build.copyFiles === 'object') {
-      const files = Object.keys(project.build.copyFiles);
-      gulp.watch(files, refreshCb).on('change', onChangeCb);
-    }
-  };
-};
+function reload() {
+  log("Refreshing browser");
```
JeroenVinke: `Refreshing the browser` would be better I think

package.json
```diff
@@ -43,10 +43,12 @@
     "nodemon": "^1.11.0",
     "aurelia-tools": "^0.2.4",
     "babel-eslint": "^7.1.1",
+    "debounce": "^1.0.2",
```
JeroenVinke: do we need these dependencies here or is it sufficient that the CLI adds these deps to the package.json that it creates during `au new`?

lib/resources/tasks/build.js
```diff
@@ -1,26 +1,40 @@
...
-  ),
+    copyFiles),
   writeBundles
 );
 
+let main;
+
+// Only watch if we're the actual task run, not if we're 
+// called from run or other task
+if (CLIOptions.hasFlag("watch") && CLIOptions.taskName() === "build") {
```
JeroenVinke: if you swap these two expressions then you could remove the comment

lib/resources/tasks/build.js
```diff
@@ -1,26 +1,40 @@
...
+  main = gulp.series(
+    build,
+    (done) => { watch(); done() }
```
JeroenVinke: nitpicky, but `done()` lacks a semicolon

lib/resources/tasks/build.js
```diff
@@ -1,26 +1,40 @@
...
-import {build} from 'aurelia-cli';
+import watch from './watch';
 import project from '../aurelia.json';
 
-export default gulp.series(
+let build = gulp.series(
   readProjectConfiguration,
-  gulp.parallel(
-    transpile,
+  gulp.parallel(transpile,
```
JeroenVinke: sorry for being nitpicky, but could you undo this change (and the one at `copyFiles)`) so that the codestyle matches the rest of the gulp tasks?

lib/resources/tasks/watch.ts
```diff
@@ -0,0 +1,100 @@
...
+let refresh = debounce(() => {
+  if (isBuilding) {
+    log("Watcher: A build is already in progress, deferring change detection...");
+    return;
+  }
+
+  isBuilding = true;
+
+  let paths = pendingRefreshPaths.splice(0);
+  let refreshTasks = [];
```
JeroenVinke: would `buildTasks` and `pendingBuildTasks` be a better name?

lib/resources/tasks/run.js
```diff
@@ -1,22 +1,10 @@
 import gulp from 'gulp';
 import browserSync from 'browser-sync';
 import historyApiFallback from 'connect-history-api-fallback/lib';
+import { CLIOptions } from 'aurelia-cli';
```
JeroenVinke: i'm nitpicking again, but could you remove the spaces around `CLIOptions`?

lib/cli-options.js
jwx: Sure, I'll see what I can do (that makes sense).

lib/resources/tasks/watch.ts
jwx: I don't think so, because it really is the watch callback (it's just initialized to a noop).

lib/resources/tasks/watch.ts
jwx: Well, it is the paths (files) pending to be refreshed and it is the tasks that's going to refresh them so I don't really think so. If you feel the names aren't clear enough I could try coming up with some better ones.

package.json
jwx: Hm. They might not be needed. I'll check.

lib/resources/tasks/build.js
jwx: Sure. (This is just formatting done by VS Code.)

lib/resources/tasks/build.js
jwx: I added the comment there to emphasize that the build task can be executed by the run task, but I'll happily remove it. For me the order of the conditions doesn't really affect the comment and this order mirrors the one in run.ts, but I'll swap them if you want.

lib/resources/tasks/run.js
jwx: VS Code again. I'm curious, though, why not use spaces within the {} when it's an import statement? They are within {} everywhere else, so is this some convention of the import statement?

JeroenVinke: We don't use spaces elsewhere in the framework ([example](https://github.com/aurelia/framework/blob/master/src/aurelia.js#L5)) so we should follow that standard

lib/resources/tasks/build.js
JeroenVinke: Yes, please swap them. It's purely for readability, A comment is fine :+1:

lib/resources/tasks/watch.ts
JeroenVinke: Right, I see what you mean. When I read "refresh" I thought of refreshing the browser, which happens only once after all changes have been built.

lib/resources/tasks/run.js
jwx: So it's an Aurelia standard? Hm, I wonder if the Aurelia VS Code plugin would be willing to accommodate this so that VS Code's autoformat feature will still work...

gheoan: An lint rule could be added to https://github.com/aurelia/tools/blob/e457d2d753e0abdc1a8f75d5d9e79ccb464f4f7c/.eslintrc.json in order to standardise the spacing.

JeroenVinke:
@jwx great work! I did encounter one issue when running `au test --watch`:

```
C:\Development\aurelia-cli-watch-flag>au test --watch
C:\Development\aurelia-cli-watch-flag\aurelia_project\tasks\test.js:47
  unit = _gulp2.default.series(_build2.default, _gulp2.default.parallel((0, _run.watch)(_build2.default, onChange), karma));
                                                                                       ^

TypeError: (0 , _run.watch) is not a function
    at Object.<anonymous> (C:/Development/aurelia-cli-watch-flag/aurelia_project/tasks/test.js:29:7)
    at Module._compile (module.js:571:32)
    at loader (C:\Development\cli\node_modules\babel-register\lib\node.js:144:5)
    at Object.require.extensions.(anonymous function) [as .js] (C:\Development\cli\node_modules\babel-register\lib\node.js:154:7)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at process.nextTick (C:\Development\cli\lib\commands\gulp.js:28:43)
    at _combinedTickCallback (internal/process/next_tick.js:73:7)
    at process._tickDomainCallback (internal/process/next_tick.js:128:9)
```

Could you look into this? I assume it could be the same logic as `au build --watch`

PR #642, opened 2017-06-06
feat(build): allow minify options to be supplied

JeroenVinke:
closes https://github.com/aurelia/cli/issues/562

with these changes you can do this to supply options to the uglifier:

```
    "options": {
      "minify": {
       "dev": false,
       "default": {
        "fromString": true
       },
       "stage & prod": {
        "max-line-len": 100000
       } 
      },
      "sourcemaps": "dev & stage"
    },
```

@AStoker what do you think?

### DIFF
--- a/spec/lib/configuration.spec.js
+++ b/spec/lib/configuration.spec.js
@@ -1,0 +1,295 @@
+'use strict';
+
+const Configuration = require('../../lib/configuration').Configuration;
+const CLIOptionsMock = require('../mocks/cli-options');
+
+describe('the Configuration module', () => {
+  let cliOptionsMock;
+
+  beforeEach(() => {
+    cliOptionsMock = new CLIOptionsMock();
+    cliOptionsMock.attach();
+  });
+
+  afterEach(() => {
+    cliOptionsMock.detach();
+  });
+
+  it('overrides default options with customized options', () => {
+    let config = new Configuration({ fromString: true }, { fromString: false });
+    expect(config.getAllOptions().fromString).toBe(true);
+  });
+
+  describe('the getAllOptions() function', () => {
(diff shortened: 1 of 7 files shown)


AStoker:
For history, related to: https://github.com/aurelia/cli/pull/602

So the `default` option, that is used across the environments, and then each environment could have its own config? Am I reading that right?
For the sake of documentation, I'd put `default` as the first option just so that it's more clear.

spec/lib/configuration.spec.js
```diff
@@ -0,0 +1,295 @@
...
+        foo: {
+          bar: 'abcd'
+        }
+      }, 'dev');
+      expect(options.getValue('foo.bar')).toBe('abcd');
+    });
+
+    it('supports one level options', () => {
+      let options = new Configuration({}, { foo: 'someValue' }, 'dev');
+      expect(options.getValue('foo')).toBe('someValue');
```
huochunpeng: @JeroenVinke the question is for 
```javascript
new Configuration({}, {foo: 'dev'}, 'dev');
new Configuration({}, {foo: 'dev'}, 'stage'); // very tricky
new Configuration({}, {foo: 'dev & stage'}, 'dev'); // this is easier
```
What we want for `getValue('foo')`.

For case#2, it looks quite tricky, we need to known all available env types to determine whether `'dev'` is a condition or plain string.

huochunpeng: Right now, #662 has some lines to bypass this issue when using "minify" options.

JeroenVinke: Yep, that's a bit tricky. I think it's OK to check whether or not the value is an object or a string, because only the code that uses the Configuration class knows what it could/should be

Issue #648, opened 2017-06-14
Unit test helpers not loaded correctly when using SystemJS


simonfox:
@JeroenVinke I have found a case where `au test` will break when using SystemJS......but I also have the solution, just wanted to run past you...

If you define a module in the test folder (for example a mock) and then import that into your test, it does not get loaded (404's). e.g. add `/test/unit/helper.js` then in `app.spec.js` add `import {MyMock} from './helper';`. This is because it is not defined in the bundle, and `/base/test/unit/helper` does not resolve to anything.

All that is needed is a SystemJS package definition added to config that sets the `defaultExtension` property:

```
SystemJS.config({
  "packages": {
    "base/test": {
      "defaultExtension": "js"
    }
  }
});
```
Currently `aurelia-karma.js` has `normalizePath` and `patchRequireJS` which aren't actually used by SystemJS and this file seems like the appropriate place to add the required SystemJS config. Was thinking I could just add `aurelia-karma.system.js` and `aurelia-karma.require.js` as I did with index.html so that `au new` adds the appropriate setup.....thoughts?

JeroenVinke:
I wonder why this isn't needed for the skeletons. Is it because of `  defaultJSExtensions: true,` ?

simonfox:
Yes....`defaultJSExtensions` is removed in SystemJS `0.20.0`. 

However we need the ability to define by package because the amodoro bundle contains module names that do not include the extension - everything in `/src`. We still want src imports in test modules to load from bundles i.e. `import {App} from '../../src/app` so we dont want `defaultExtension` to apply to that path.

Make sense?

JeroenVinke:
It does. Could you put together a PR?

simonfox:
Yes, will do.

PR #651, opened 2017-06-14
fix(systemjs-loader): systemjs config for karma test runner

simonfox:
An import of a unit test _helper_ module from within a spec was failing (e.g. a module that defines a mock). This was due to the import statement normalization result not having a .js extension. This change splits aurelia-karma into two loader specific implementations, with the SystemJS version adding a package configuration for the base/test path that sets `defaultExtension` to js. `au new` adds an implementation based on the selected loader.

closes aurelia/cli#648


lib/commands/new/loaders/require.js
```diff
@@ -8,7 +8,10 @@ module.exports = function(project) {
   project.loaderTextPlugin = 'text';
   project.loaderScript = 'node_modules/requirejs/require.js';
   project.addToContent(
-    ProjectItem.resource('index.html', 'content/require.index.html')
+    ProjectItem.resource('index.html', 'content/require.index.html'),
+    project.tests.add(
```
JeroenVinke: what happens when the user said he didn't want to use Karma?

simonfox: Good catch...I will move it back into the `karma` step.

simonfox:
@JeroenVinke should be good now.

PR #660, opened 2017-06-21
fix(dependencies): update gulp-sass version

fabioluz:
gulp-sass is dependant on node-sass and the current version (2.3.1) tries to download an old version of node-sass that apparently is not available on GitHub anymore. Changing the gulp-sass version to latest version (3.1.0) solves the problem.

To reproduce the problem just install a new project using Sass as the CSS Pre-processor. You'll see that gulp-sass cannot find node-sass version, thus cannot be installed.

Environment: CLI 0.30.0; NPM 5.0.3; Node 8.1.2; SO Windows 10

### DIFF
--- a/lib/dependencies.json
+++ b/lib/dependencies.json
@@ -31,7 +31,7 @@
   "gulp-sourcemaps": "^2.0.0-alpha",
   "gulp-less": "^3.1.0",
   "gulp-postcss": "6.1.1",
-  "gulp-sass": "^2.3.2",
+  "gulp-sass": "^3.1.0",
   "gulp-stylus": "^2.5.0",
   "gulp-typescript": "^3.1.4",
   "gulp-tslint": "^5.0.0",


JeroenVinke:
Hey @fabioluz! I'm unable to reproduce the issue. What error are you getting?

We need to update to gulp-sass@3.1.0 at some point in the future anyway, so i'd be fine with that. Is it backwards compatible?

fabioluz:
The error I'm getting is:

> Installing project dependencies.
> 
> node-sass@3.13.1 install C:\Users\Fabio Luz\Documents\DevProjects\testCLI\node_modules\node-sass
> node scripts/install.js
> 
> Downloading binary from https://github.com/sass/node-sass/releases/download/v3.13.1/win32-x64-57_binding.node
> Cannot download "https://github.com/sass/node-sass/releases/download/v3.13.1/win32-x64-57_binding.node":
> 
> HTTP error 404 Not Found

See that there's no `win32-x64-57_binding.node` available at http://github.com/sass/node-sass/releases?after=2356. 

About the backward compatibility:

- According to that node-sass release notes, the version 4.X is backwards compatible with 3.13.1 (https://github.com/sass/node-sass/releases?after=2356). 

- According to that gulp-sass release notes, the version 3.X is backwards compatible with 2.X (https://github.com/dlmanning/gulp-sass/releases)

Issue #435, opened 2016-12-10
Change the main property in a dependency configuration to be optional


JeroenVinke:
**I'm submitting a bug report**

* **Library Version:**
0.23.0

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
6.3.1

* **NPM Version:**
3.10.7

* **Browser:**
all

* **Language:**
all 

**Current behavior:**
Kendo-UI has a bunch of controls that are exported as individual files. There is no specific "main" file, so in dependencies section of aurelia.json we do not need to configure the "main" property either:

```
{
    "name": "kendo-ui-core",
    "path": "../node_modules/kendo-ui-core"
},
```

However, this results in the following error:

![image](https://cloud.githubusercontent.com/assets/2189477/21074754/f4d0b74e-bf00-11e6-8e5a-49eaaf03bfbc.png)

Luckily the application still runs and it works as it should (the mapping works fine).

**Expected/desired behavior:**

* **What is the expected behavior?**
The main property should be optional, so that simple path mappings can be declared as well.

* **What is the motivation / use case for changing the behavior?**
To be able to use libraries that do not have a specific main file.

Assuming that there is no other way to configure packages without main files, I can create a PR to resolve this error

original issue: https://github.com/aurelia-ui-toolkits/aurelia-kendoui-bridge/issues/660
repro: https://github.com/JeroenVinke/aurelia-app-optional-main

EisenbergEffect:
@JeroenVinke Please feel free to submit a PR for this.

zedL:
Any news here?

JeroenVinke:
@zedL I will take a look at this shortly

StrahilKazlachev:
What about being explicit, setting it to `false` - `"main": false`, or something? My idea is to show intention.

JeroenVinke:
That's probably a better solution. Right now `main` is optional, but then it assumes that the `path` is pointing to an existing file. I'll update the PR

zewa666:
This would be great in order to create a minimal import of Rx.js. I like the `main: false` approach as it is very descriptive :+1:

JeroenVinke:
:+1:

I have updated the PR: https://github.com/aurelia/cli/pull/673

PR #673, opened 2017-07-03
fix(bundler): don't trace dependencies which have no main file

JeroenVinke:
closes https://github.com/aurelia/cli/issues/435#issuecomment-293850306

I need to do some final tests and rebase it on master

### DIFF
--- a/lib/build/dependency-inclusion.js
+++ b/lib/build/dependency-inclusion.js
@@ -8,7 +8,10 @@ exports.DependencyInclusion = class {
     this.description = description;
     this.items = [];
     this.traceDependencies = true;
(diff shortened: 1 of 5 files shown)


JeroenVinke:
https://github.com/aurelia/cli/issues/435#issuecomment-312632699 is a better solution. I'll update this PR

zewa666:
Wow, that fix, so simple yet so effective.

Fresh TS scaffold with:
```
{
  "name": "rxjs",
  "path": "../node_modules/rxjs",
  "main": "Rx"
}
```
648kb vendor file

and with `main: false` 410kb.

of course its just a demo using only minimal Rx dependencies, but I still expect at least 100-150kb savings on a typical RxJS project.

You better get that merged asap :)

Issue #671, opened 2017-06-30
Double `dist` folder in `aurelia.json` after package installation


nawijn:
**I'm submitting a bug report**

* **Library Version:**
0.30.1

**Please tell us about your environment:**
* **Operating System:**
Linux (Fedora 25)

* **Node Version:**
6.10.3

* **NPM Version:**
3.10.10

* **Browser:**
all 
* **Language:**
all

**Current behavior:**
After `au install bootstrap-v4-dev` the Bootstrap section in`aurelia` reads
`
          {
            "name": "bootstrap-v4-dev",
            "main": "js/bootstrap",
            "path": "../node_modules/bootstrap-v4-dev/dist",
            "resources": [
              "dist/css/bootstrap-grid.css",
              "dist/css/bootstrap-grid.min.css",
              "dist/css/bootstrap-reboot.css",
              "dist/css/bootstrap-reboot.min.css",
              "dist/css/bootstrap.css",
              "dist/css/bootstrap.min.css"
            ]
          }
`
When building/running the app (cleanly generated with `au new`) the compiler complains:

`Error: File not found with singular glob: /tmp/au-test/ui/node_modules/bootstrap-v4-dev/dist/dist/css/bootstrap-grid.css`
 
As you can see there the `dist` folder is appearing twice, while the actual folder structure
only has a single `dist`.

**Expected/desired behavior:**
No compilation error after a `au install bootstrap-v4-dev`

gishmel:
This is a mistake with your including dist inside of the resources. It is using the path attribute as the base folder and of course you will get two dist folders as its requested in the resources at their beginning. Under resources you should only have them listed as "css/bootstrap-grid.css" and make them like that.

gishmel:
Feel free to check out my latest PR on the Aurelia CLI as I believe it has solved this problem. Sorry for not understanding before that it was aurelia CLI's problem with overlapping the path with the resources path. Hopefully it can get merged in shortly. https://github.com/aurelia/cli/pull/674

PR #674, opened 2017-07-04
fix(importer): prevent duplicate dist

gishmel:
closes https://github.com/aurelia/cli/issues/671

### DIFF
--- a/lib/importer/services/resource-inclusion.js
+++ b/lib/importer/services/resource-inclusion.js
@@ -58,7 +58,14 @@ let ResourceInclusion = class {
 
   getResources(globExpr) {
     return this.glob(globExpr, { cwd: this.package.rootPath })
-    .then(files => files.map(file => path.posix.join(file)));
+    .then(files => files.map(file => {
+			let directoryFromPath = this.package.path.substring(this.package.path.lastIndexOf('/') + 1);
+			let directoryFromFile = file.split('/').shift();
+			if (directoryFromPath == directoryFromFile) {
+				file = file.substring(file.indexOf('/') + 1)
+			}
+			return path.posix.join(file);
+		}));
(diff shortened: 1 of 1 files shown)


JeroenVinke:
@gishmel could you rename this commit so that it follows our [guidelines](https://github.com/DurandalProject/about/blob/master/CONTRIBUTING.md#-git-commit-guidelines)? That will cause it to show up in our release notes

gishmel:
Hows that?

JeroenVinke:
Could you give https://stackoverflow.com/a/20853093 a shot? The goal is to have an updated commit message in git, which is different from the title of the pull request in github. The commands in the stackoverflow post needs to be done on the branch that you created the PR for. When that is done you should see the updated commit message in https://github.com/aurelia/cli/pull/674/commits. Let me know if you run into any issues

gishmel:
Yep sorry for the confusion done. Was thinking about amending my commit but didn't realize that's what you wanted. Hopefully its all good now?

JeroenVinke:
Much better! It might be a bit too long. If you could shorten it, then it's perfect. Something like `fix(importer): prevent duplicate dist` should be good

gishmel:
Okay I just wanted to illustrate it would cause any duplicate paths between package.path and resources path to be removed but yeah I guess its fine as it will still do just that.

JeroenVinke:
Does this also work for a deeper folder structure? E.g. output/dist/test.css

gishmel:
No I did not change the underlying how it got access to the folder structure as its a fairly simplistic regular expression using the glob package. I can modify if need be just I don't believe that is at all addressed in the current implementation of the resource-inclusion package; at least from my reading and understanding of the code.

PR #686, opened 2017-08-02
Added customBaseUrl to Require Loader

magnusdanielson:
Add customBaseUrl to the paths section in aurelia.json. When using this commit it will add the customBaseUrl before the current baseUrl.

The purpose is to allow loading the scripts from other domain than the webpage is served. If for example the HTML page is hosted in SharePoint online and you want to load the script from say Azure, you need a custom baseUrl.

`"paths": {
    "root": "src",
    "resources": "resources",
    "elements": "resources/elements",
    "attributes": "resources/attributes",
    "valueConverters": "resources/value-converters",
    "bindingBehaviors": "resources/binding-behaviors",
    "customBaseUrl": "http://localhost:9002/"
  },`

### DIFF
--- a/lib/build/loader.js
+++ b/lib/build/loader.js
@@ -72,6 +72,11 @@ exports.createRequireJSConfig = function createRequireJSConfig(platform, bundler
     config.paths[currentBundle.moduleId] = location + '/' + currentBundle.moduleId + (buildOptions.isApplicable('rev') && currentBundle.hash ? '-' + currentBundle.hash : '');
   }
 
+  if(bundler.project.paths.customBaseUrl)
+  {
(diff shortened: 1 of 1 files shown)


magnusdanielson:
Some might argue that the paths section is not correct place for this configuration. Please update me if you want to move it.

magnusdanielson:
Also notice that I have not updated the code if you are using SystemJS instead of Require. I know far to little about SystemJS right now.

JeroenVinke:
@jwx do you think that we could provide environment specific baseUrls when we improve the build targets?

EisenbergEffect:
@JeroenVinke Was your thought that we should skip this and instead handle it through the in-progress build target work?

JeroenVinke:
merged https://github.com/aurelia/cli/pull/807

Issue #682, opened 2017-07-21
au build improperly links app-bundle


ariovistus:
**I'm submitting a bug report**

* **Library Version:**
0.30.1

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
8.1.2

* **NPM Version:**
5.0.3

* **Browser:**
all 
* **Language:**
TypeScript

**Current behavior:**
when I set ```"rev": true``` in aurelia.json, au build generates vendor-bundle and app-bundle properly, but embeds the path "wwwroot/scripts/app-bundle.js" in vendor-bundle rather than "wwwroot/scripts/app-bundle-(rev).js", preventing the web app from loading.

**Expected/desired behavior:**
app should load

* **What is the expected behavior?**

* **What is the motivation / use case for changing the behavior?**

ariovistus:
steps to reproduce:
```
au new --here
```
in the prompt, select
```
2. systemjs
2. aspnet core
2. typescript
1. default template
1. no css preprocessor
2. skip unit testing
1. create
1. install deps
```

then edit aurelia.json with 
```
"rev": true
```
then run
```
au build
```

ariovistus:
choosing systemjs seems to be what triggers this

PR #692, opened 2017-08-12
fix(loader): rev should be applied to bundle for systemjs

JeroenVinke:
closes https://github.com/aurelia/cli/issues/682


zewa666:
Great catch :+1:

Issue #663, opened 2017-06-22
Mixing sass+css and run --watch


tdamir:
**I'm submitting a bug report**

* **Library Version:**
0.30.0

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
6.9.1

* **NPM Version:**
4.5.0

* **Browser:**
all 

* **Language:**
all 

**Current behavior:**
I'm not sure if this is a bug but it worked in in previous versions. If I set multiple globs for cssProcessor in aurelia.config, the changes are not detected anymore. 
````
  "cssProcessor": {
    "id": "sass",
    "displayName": "Sass",
    "fileExtension": ".scss",
    "source": [
      "src/**/*.scss",
      "src/**/*.css"
    ]
  },
````
**Expected/desired behavior:**
The changes were detected in previous versions. Bundling/tracing/transforming is still working as expected so only the watch is involved.

**Comment by JeroenVinke** [comment]
@jwx could you take a look? Could this be related to the new gulp tasks?

**Comment by jwx** [comment]
Very likely related to https://github.com/aurelia/cli/issues/675.

(@JeroenVinke Sorry, looking into this got lost.)

**Comment by 4imble** [comment]
I had something similar in .less

Is it that the files are not actually changed and so ignored?

Changing my process-css.ts to this worked for me.

```
export default function processCSS() {
  return gulp.src(project.cssProcessor.source)
    //.pipe(changedInPlace({firstPass:true})) //<--- this filters unchanged files
    .pipe(plumber({ errorHandler: notify.onError('Error: <%= error.message %>') }))
    .pipe(sourcemaps.init())
    .pipe(less())
    .pipe(build.bundle());
}
```

PR #706, opened 2017-08-21
fix(watch): support array of source files

JeroenVinke:
closes https://github.com/aurelia/cli/issues/663


zewa666:
oh yeah that ones great, had to modify my local watch.js in similar manner to get that to work. Code looks good 👍

Issue #453, opened 2016-12-30
[Feature Request] Add linting to build tasks


s-hoff:
**I'm submitting a feature request**

* **Library Version:**
0.23.0

**Please tell us about your environment:**
* **Operating System:**
Linux (Ubuntu)

* **Node Version:**
6.5.0

* **NPM Version:**
3.10.3

* **Browser:**
all

* **Language:**
all

**Current behavior:**
During build, any linting rules are ignored.

**Expected/desired behavior:**

* **What is the expected behavior?**
Linting should either be done in the build process, the transpile task, or there shoud be a task set up do run the tools manually.

* **What is the motivation / use case for changing the behavior?**
There are already linting rules in the default projects, so it makes sense to use and enforce them. With this set up, one does not have to manually set it up when creating a new project.

JeroenVinke:
https://github.com/aurelia/cli/pull/718

Issue #452, opened 2016-12-30
[Feature Request] Add default tslint setup


s-hoff:
**I'm submitting a feature request**

* **Library Version:**
0.23.0

**Please tell us about your environment:**
* **Operating System:**
Linux (Ubuntu)

* **Node Version:**
6.5.0

* **NPM Version:**
3.10.3

* **Browser:**
all

* **Language:**
TypeScript

**Current behavior:**
There is a tslint.json for TypeScript projects and an eslintrc.json for babel. The eslint has no configuration in it, while the eslintrc is extending the aurelia tools setup.

**Expected/desired behavior:**

* **What is the expected behavior?**
The tslint.json should prodive some kind of default. We could base it on the eslint setup from aurelia-tools or maybe aurelia-ux (as it is the only TS project I know of among the official projects).

* **What is the motivation / use case for changing the behavior?**
Linting provides good code quality. It should be ready to go for both ESNext and TypeScript projects, and have similiar defaults.

JeroenVinke:
https://github.com/aurelia/cli/pull/718

PR #718, opened 2017-08-24
feat(tasks): add lint tasks

zewa666:
adds lint tasks for js and ts using respectively eslint or tslint

### DIFF
--- a/lib/resources/tasks/lint.ts
+++ b/lib/resources/tasks/lint.ts
@@ -1,0 +1,11 @@
+import * as gulp from 'gulp';
+import tslint from 'gulp-tslint';
+import * as project from '../aurelia.json';
+
+export default function lint() {
+  return gulp.src([project.transpiler.source])
+    .pipe(tslint({
+      tslint: require("tslint")
+    }))
+    .pipe(tslint.report("prose"));
+}
--- a/lib/commands/new/buildsystems/cli/transpilers/babel.js
+++ b/lib/commands/new/buildsystems/cli/transpilers/babel.js
@@ -12,7 +12,8 @@ module.exports = function(project) {
     ProjectItem.resource('.eslintrc.json', 'content/eslintrc.json'),
     ProjectItem.resource('.babelrc', 'content/babelrc')
   ).addToTasks(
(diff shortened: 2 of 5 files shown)


zewa666:
this one just adds the new lint tasks but its not yet wired up to be copied during scaffolding of a new app.

Related issues https://github.com/aurelia/cli/issues/453 and https://github.com/aurelia/cli/issues/452

lib/resources/tasks/lint.ts
```diff
@@ -0,0 +1,11 @@
+import * as gulp from 'gulp';
+import tslint from 'gulp-tslint';
+import * as project from '../aurelia.json';
+
+export default function lint() {
+  return gulp.src([project.transpiler.source])
```
JeroenVinke: source can be an array, will that cause issues?

JeroenVinke:
Great! We can add the tasks [here](https://github.com/aurelia/cli/blob/master/lib/commands/new/buildsystems/cli/transpilers/babel.js#L11-L28) and also any dependencies that the task has. And the equivalent for [typescript](https://github.com/aurelia/cli/blob/master/lib/commands/new/buildsystems/cli/transpilers/typescript.js)

lib/resources/tasks/lint.ts
zewa666: Nope this works ok. The reason is that I personally prefer to lint my unit tests as well so I'd typically add them here too. Not sure though if that counts for the masses. What would you say?

zewa666:
Great, I'll make an update to include that.

gheoan:
Running `aurelia help` does not show `lint` as a valid command. How can that be fixed?

JeroenVinke:
@gheoan the lint.json is in the cli but doesn't get copied to the project directory. So we need a [similar line](https://github.com/zewa666/cli/blob/e16c2166dc962640b8041ca230fa00f5f61bd84a/lib/commands/new/buildsystems/cli/index.js#L120) in [here](https://github.com/zewa666/cli/blob/e16c2166dc962640b8041ca230fa00f5f61bd84a/lib/commands/new/buildsystems/cli/transpilers/babel.js#L16)

PR #720, opened 2017-08-25
fix(webpack): remove comma from arguments

afm-sayem:
version: 0.31.2

Trying to create a new project using webpack fails with following error:
```
ERROR [ProjectCreation] Failed to create the project due to an error: Unexpected token )
INFO [ProjectCreation] /home/sayem/.nvm/versions/node/v7.10.1/lib/node_modules/aurelia-cli/lib/commands/new/buildsystems/webpack/index.js:41
  ).addToContent(
  ^
SyntaxError: Unexpected token )
```


JeroenVinke:
I didn't have this issue during testing. It must be related to the nodejs version. Could you try and do the same with node 8? Anyway, the PR looks good

PR #725, opened 2017-08-26
fix(npm script): add required dependency to start npm task

afm-sayem:
In the generated project, the `npm start` task runs `nps` but the dependencies don't have that package so the task fails.

To be foolproof, the executable should be run from `./node_modules/.bin`, but I haven't changed the task because all the tasks in `package-scripts.js` assume that the users will have `node_modules/.bin` added to their path.


JeroenVinke:
Thank you @afm-sayem

PR #748, opened 2017-09-06
Add option to pass SystemJS configuration

PisaJan:
SystemJS configuration is could be usefull for many cases, for example in my case i need to add authorization headers to request done by SystemJS.


JeroenVinke:
Thank you @PisaJan

Issue #659, opened 2017-06-21
Getting "Unexpected token <" error since using SystemJS


RomainLanz:
Hey :wave:

<hr>

**Operating System** : macOS Sierra
**Node.js version** : v8.1.2
**npm version**: 5.0.3

<hr>

I have switched my loader from `RequireJS` to `SystemJS` this morning.
To make the change I have install a new aurelia application and I have compare `package.json`, `tasks/*`, `aurelia.json` and `index.html` files and made the change needed.

The application bundle correctly, but when I try to launch it in the browser I have the error `Unexpected token <` at random place.

Note that everything was working fine before making the change.

<center><img width="511" alt="screen shot 2017-06-21 at 11 48 03" src="https://user-images.githubusercontent.com/2793951/27379185-a7474198-567a-11e7-9704-d677f2d90a4f.png"></center>

If we take a look to the `vendor-bundle.js` file at the line `11707` we arrive on the `loadModule` function.

```js
DefaultLoader.prototype.loadModule = function (id) {
  var _this3 = this;

  return System.normalize(id).then(function (newId) {
    var existing = _this3.moduleRegistry[newId];
    if (existing !== undefined) {
      return Promise.resolve(existing);
    }

    return System.import(newId).then(function (m) {                  // 11707
      _this3.moduleRegistry[newId] = m;
      return ensureOriginOnExports(m, newId);
    });
  });
};
```

The package `@fivb/sdk` is a TypeScript package compiled and transpiled to ES5 in different module formats (`amd`, `commonjs`, `es2015`, `umd`).

In my `aurelia.json`, I load it as an `amd` package.

```json
{
  "name": "@fivb/sdk",
  "path": "../node_modules/@fivb/sdk/dist/amd",
  "main": "index"
},
```

Since SystemJS should handle `commonjs` as well, I have tried to use this version.

```json
{
  "name": "@fivb/sdk",
  "path": "../node_modules/@fivb/sdk/dist/commonjs",
  "main": "index"
},
```

Now the error `Unexpected token <` is gone for `@fivb/sdk` but happen for `axios`.

```
{
  "name": "axios",
  "path": "../node_modules/axios/dist",
  "main": "axios"
},
```

<img width="500" alt="screen shot 2017-06-21 at 12 16 07" src="https://user-images.githubusercontent.com/2793951/27379423-7506fc86-567b-11e7-9610-cf95bb12a59b.png">

The line 3388 is the definition of `SystemJS`.

```js
/*
[...]

Issue #624, opened 2017-05-10
Sourcemap improvements


JeroenVinke:
There are a couple things we should fix regarding sourcemaps.

- [x] **Console.log gets confused of the origin** (https://github.com/aurelia/cli/issues/592)
When webpack encounters a transpiled javascript file with no sourcemap, it creates a sourcemap for the transpiled file. The CLI does not do this, it only adds sourcemaps to the .map file when it finds a sourcemap. If we do what webpack does it'll resolve https://github.com/aurelia/cli/issues/592 and will give a much nicer debug experience for the vendor-bundle, because chrome will know which part of the vendor-bundle belongs to which library, and all libraries will show up in the sources panel of Chrome. [This](https://github.com/gulp-sourcemaps/gulp-sourcemaps/blob/master/src/init/index.js#L58-L74) is what we could use to generate source maps for transpiled files.

- [x] **Multiple instances of sourceMappingUrl** (https://github.com/aurelia/cli/issues/532)
The CLI uses the sourceMappingUrl comments in javascript sources to create a .map file containing all sourcemaps. At the bottom of the bundle a new sourceMappingUrl comment is appended pointing to this .map file. This is the only instance of sourceMappingUrl that should be in the bundle, as otherwise IE and Edge gets confused. So we should strip all other sourceMappingUrl comments.

- [ ] **Base64 sourcemaps are not included**
The if-statement [here](https://github.com/aurelia/cli/blob/master/lib/build/bundle.js#L166-L170) causes base64 sourcemaps not to be included in the generated .map file.

The [write function](https://github.com/aurelia/cli/blob/master/lib/build/bundle.js#L115-L283) of the Bundler has too many responsibilities. We should extract source map generation into a different module, if possible

EisenbergEffect:
Sounds good.

jholland918:
Is debugging with sourcemaps supported when using WebPack? Or is it only supported for projects created using the requirejs/systemjs options? I ask because I'm trying out unit testing for a project using the WebPack option and I'm only able to debug through the transpiled JavaScript.

Alexander-Taran:
@jholland918 it is supported. depends on how you configure typescript/babel and webpack
HMR though breaks the maps.. after 1st hot injection

Alexander-Taran:
@3cp I think we're mostly ok here..

3cp:
The cli-bundler is quite big without enough unit test coverage. I would not attempt fixing the source map :-)

[...]

PR #751, opened 2017-09-12
feat(source-maps): bundling improvements

dima-v:
multiple fixes to source maps:
- fixes offset problems caused by:
	- prepended scripts;
	- duplicate dependency main module;
- remove sourece map file comments;
- add EOL separator for the anonymous module wrapper;

closes aurelia/cli#659, related to aurelia/cli#624

### DIFF
--- a/lib/build/amodro-trace/write/packages.js
+++ b/lib/build/amodro-trace/write/packages.js
@@ -29,7 +29,7 @@ function packages(options) {
                                    filePath, contents, options);
 
     if (packageName && !hasPackageName) {
-      contents += ';define(\'' + packageName + '\', [\'' + moduleName +
+      contents += '\n;define(\'' + packageName + '\', [\'' + moduleName +
                   '\'], function (main) { return main; });\n';
     }
 
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -112,14 +112,13 @@ exports.Bundle = class {
   }
 
   getBundledFiles() {
-    return this.includes.reduce((a, b) => a.concat(b.getAllFiles()), []);
+    return uniqueBy(this.includes.reduce((a, b) => a.concat(b.getAllFiles()), []), (file) => file.path);
   }
 
   write(platform) {
     if (!this.requiresBuild) {
       return Promise.resolve();
     }
-
 
     let work = Promise.resolve();
     let loaderOptions = this.bundler.loaderOptions;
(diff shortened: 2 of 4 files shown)


simonfox:
@dima-v have been discussing this with @JeroenVinke over on Gitter...couple of queries

1. what was the reason for modifying the determination of the source root
2. why the `columnOffset` calculation is moved to apply to all files

Below is @JeroenVinke comment with details...

I found one issue though, when I tried to use aurelia-configuration with these changes.
It looked like this:

image


The index.ts, aurelia-configuration.ts and deep-extend.ts should not be in the src directory, so I dug into the code.
The variables [here](https://github.com/dima-v/cli/blob/8d4777d7f0cf3a2450e97ba92556331ddb999e3f/lib/build/bundle.js#L193-L195) have the following values:
aurelia-configuration
/node_modules/aurelia-configuration/dist/amd
/node_modules/aurelia-configuration
Thus the source root is set to `/node_modules/aurelia-configuration`. If we look at the sourcemap, [this one](https://github.com/Vheissu/aurelia-configuration/blob/master/dist/amd/aurelia-configuration.js.map) then the sources are defined as `../../src/aurelia-configuration.ts`, so when the sourceRoot is `node_modules/aurelia-configuration` then Chrome will think that those sources belong in the `src` folder.
So I changed the code to:
```
if (sourceMap !== null) {
  let sourceRoot = parsedPath.dir.substring(process.cwd().length);
  sourceMap.sourceRoot = sourceRoot.replace(/\\/g, '\/');
}
```
and that fixed it. Then it works perfectly. I'd like to know the reason for changing [this code](https://github.com/aurelia/cli/pull/751/files#diff-0d757a7f03cc6a6a3151e773bd9b0ca0R103) though, and how I can reproduce the issue that you had so I can confirm the fix.
Much thanks!
If you could explain the issues that you've had for which you changed those two bits of code (the offsets) and the sourceRoot, then that would help me a lot

dima-v:
@JeroenVinke thank you for your suggestion on how to fix the source root issue.
It works for us and I updated the pull request with related changes.

Regarding to the column offset calculation ...

with the following configuration for a bundle file:
- prepend files or/and dependency package with resources (such as html, css or js files)

we found that resulting map file for the bundle will have incorrect mapping and will cause unpredictable behavior when debugging. The column offset needs to consider all files in the bundle, not just those which have a sourcemap associated.

we used that tool http://sokra.github.io/source-map-visualization/#custom to find the issue.

Please let me know if you need a working example.

JeroenVinke:
Thanks for the quick response @dima-v. Nice work

Issue #752, opened 2017-09-12
Extend the Environment-Specific Dependencies functionality to 'prepend' and 'append'


arnederuwe:
**I'm submitting a feature request**

**Current behavior:**
The prepend/append section is only configurable with paths to js libraries

* **What is the expected behavior?**
Make it possible to only load certain js libraries depending on the selected environment, like this:
```
"prepend": [
          "scripts/jszip.min.js",
          "node_modules/bluebird/js/browser/bluebird.core.js",
          {
                path: "node_modules/aurelia-cli/lib/resources/scripts/configure-bluebird.js",
                env: "stage & prod"
          },
          {
                path: "node_modules/aurelia-cli/lib/resources/scripts/configure-bluebird-dev.js",
                env: "dev"
          },
          "node_modules/element-closest/element-closest.js",
          "node_modules/requirejs/require.js"
        ],
```

* **What is the motivation / use case for changing the behavior?**
Today I had a headache trying to improve the startup performance in IE, which was terrible...
Turns out it was BlueBird related, and it was previously fixed: https://github.com/aurelia/templating/issues/448
Since the CLI has moved the BlueBird config to an internal js file, this issue is present again as the file doesn't incorporate the fix committed by @EisenbergEffect in the issue above.
As you do want these stack traces in dev mode, this is an ideal use-case for the change I propose. If you could load a different BlueBird config depending on your environment flag, this would be fixed elegantly.

I would be happy to make a PR for this, just point me in the direction where I should start making these changes.

JeroenVinke:
I like it, and a PR would be much appreciated. I think it would have to be done [in this area](https://github.com/aurelia/cli/blob/master/lib/build/bundle.js#L17-L18)

arnederuwe:
Great! I'm on it

JeroenVinke:
https://github.com/aurelia/cli/pull/761

PR #761, opened 2017-09-28
feat(bundle): prepend and append now support environment specific dependencies

arnederuwe:
This implements feature request https://github.com/aurelia/cli/issues/752
I added an extra bluebird config file without long stacktraces and changed the default generated aurelia.json file to selectively pick the right settings depending on the chosen environment

### DIFF
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -14,8 +14,8 @@ exports.Bundle = class {
     this.bundler = bundler;
     this.config = config;
     this.dependencies = [];
-    this.prepend = (config.prepend || []).filter(x => bundler.itemIncludedInBuild(x));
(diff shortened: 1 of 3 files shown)


JeroenVinke:
Sweet! I wonder though, does this actually use the `env` property on there to filter which files have to be prepended and appended?

arnederuwe:
Yes!
To my surprize that was already implemented :)
`bundler.itemIncludedInBuild(x)` does exactly that, the only thing I had to do was to make sure the output list are all strings and not a mix between strings and objects, like the dependencies list can be.

I tested this change by generating a new project and playing with both prepend and append, so it should be good to go.

JeroenVinke:
awesome, thank you for the confirmation

Issue #763, opened 2017-09-29
Bundle does not use platform baseDir


christophano:
**I'm submitting a bug report**

* **Library Version:**
0.31.3

**Please tell us about your environment:**
* **Operating System:**
OSX 10.12.6

* **Node Version:**
7.2.1

* **NPM Version:**
4.1.1

* **Browser:**
all

* **Language:**
ESNext

* **Loader/bundler:**
SystemJS

**Current behavior:**
When defining the `platform.baseDir` property in config the `baseDir` value is not used.

* **What is the expected behavior?**
I expected that the cli would use the `baseDir` as the value to define the base directory in which to write the bundles and find the index file.
For example: if I define the `baseDir` as "wwwroot", the index as "index.html" and the output as "scripts" then I would expect the bundler to write the bundle (e.g. "vendor-bundle.js") to "wwwroot/scripts/vendor-bundle.js" and to update the script src in "wwwroot/index.html" to point to "scripts/vendor-bundle.js".

* **What is the motivation / use case for changing the behavior?**
For my use case I am using Aurelia served from an aspnet core project, which provides the rest api. I want to be able to keep the client source in a single child folder (named "Client" in my case) and have the cli spit out the bundles into the wwwroot folder.

I have a fix for this, and am happy to submit a PR, but I wanted to first confirm that this is a bug and that I'm not just misunderstanding the intended use case for the `baseDir` property.

jwx:
@christophano I've added a PR, https://github.com/aurelia/cli/pull/765, that will implement build targets. Do you think it will cover your needs? And if not, what's missing?

christophano:
@jwx that looks like it's 90% of the way there. What I need specifically is to be able to specify the base path for the output (which should support paths outside of the application root) and have the index and output as relative from there. I'm not entirely certain that the target root will cover this?

jwx:
@christophano Target root can be outside of project root ("../sibling-folder-to-project/other-project-root") and target index and output are indeed relative to target root so that shouldn't be a problem at all.

jwx:
Just re-read your use case and just setting root: "../wwwroot" should do the trick (provided wwwroot is a sibling folder to Client).

christophano:
That being the case, I'm happy that this PR fixes my issue - thanks!

Issue #200, opened 2016-06-28
wrong bundle path when setting relative tagret in config


Alexander-Taran:
I like the idea of clean output of CLI workflow.
I also am in the process of enriching a web site with aurelia.enchance
I figured that I don't want to pollute already crowded folder of a mvc5 app and had set up the aurelia cli based app in an adjacent folder.

Then I instructed CLI to output bundles to the scripts folder of mvc app

```
 "build": {
    "targets": [
      {
        "id": "default",
        "displayName": "Default",
        "output": "../mvc.app/scripts"
      }
    ], 
```

now bundle metadata in vendor-bundle contains additional path for app-bundle
`"app-bundle":"../../mvc.app/scripts/dynamic-comments-bundle"`

EisenbergEffect:
I'll be honest, we hadn't considered that type of organization at all. We'll need to see how we can address this but it may take some time, since we need to focus on issues related to more standard setups. Feel free to make recommendations for a fix or submit PRs. As usual, community help is welcomed and encouraged.

flytzen:
I've just run into this problem as well. As far as I can tell, this is essentially the same as this PR: https://github.com/aurelia/cli/issues/173. 
The work-around described in there fixed it for me.

Alexander-Taran:
To be fair, I worked that around.
For development I used symlinks from au project /dist to /scripts folder or something in mvc project
and for deployment I created another cmd file that published /dist or export to subpath on IIS..

If this scenario is not interesting to anybody else, this can be closed

Alexander-Taran:
related question on SO
https://stackoverflow.com/questions/43857681/aurelia-cli-build-to-parent-folder
can't answer it )-:

Alexander-Taran:
Workaraundable..

Issue #724, opened 2017-08-25
[feature requests] Multiple bundles/targets & platforms


diegohb:
I've been trying to figure out how to support multi-spa and leverage build targets and platform in aurelia.json to support a more sophisticated build setup.

attempts:
1. just standard splitting bundles to break off dependencies (each SPA may have different 3rd-party vendor libraries): https://github.com/diegohb/samples-aurelia-vanilla_cli/tree/master-task017
--this broke the [karma tests ](https://github.com/aurelia/cli/issues/689) but otherwise worked.
2. tried to trick the au bundler to update a razor partial that each razor view references so that both root container SPA pages get the bundler rewrite: https://github.com/diegohb/samples-aurelia-vanilla_cli/tree/master-task018
3. Now i'm trying to specify SPA-specific bundles and hopefully share the common aurelia & systemjs bundle in both.. 

**The requests:**
1. ability to specify array of strings to build.targets[x].index . eg ["index1.html", "index2.html"]

2. change updateIndexForConfig() method to work against any type of file, just look for <script> tag within file and update it.

3. change build.loader.configTarget property to reference build.targets.id  to load it's .index property. This will help avoid having an additional place to change the value of index page to be updated.

4.  add property "buildTarget" to bundle object that works like dependencies[x].env property (can specify multiple target concatenated with ampersand).

5. add property "platform" to buildTarget to specify what targets are meant for which platforms and then allow parameter "platform". eg. au build --platform "netcore2.0" or "electron" etc..

I wonder if anyone can provide links to source file(s) related to each one of these requests ? I'd love to look at helping out with some.. maybe the first one for starters.. 😃

diegohb:
i see it's related to https://github.com/aurelia/cli/issues/442

diegohb:
I'd love to know why this isn't working as expected to get a better understanding of the internals of the cli and current limitations .. https://github.com/diegohb/samples-aurelia-vanilla_cli/commit/20e557b5acb73e04a7f63d915494da27125cf41b 🆘

JeroenVinke:
@diegohb i'd like to help you debug but i''m a bit behind on all my other work. If you can wait until next weekend I can help you out

diegohb:
@JeroenVinke - thanks a lot! 👍 👍 👍

johan-v-r:
Adding the following code to the top of my `build.js` file I can specify `au build --target mobile`
```js
import project from '../aurelia.json';

let target = "web";
[...]

Issue #677, opened 2017-07-13
Add ability to add a baseUrl per environment to aurelia.json


DavidColemanUK:
**I'm submitting a feature request**

* **Operating System:**
Windows 10

* **Node Version:**
6.10.0

* **NPM Version:**
3.10.10

* **Aurelia CLI Version**
CLI 0.27.0 

* **Language:**
TypeScript 2.3.2

**Current behavior:**
Currently, there is no way to change the baseUrl per deployment environment.

Our current work around is to bruteforce the line using powershell to modify, during the release process to each server.  

**Expected/desired behavior:**
Ability to provide baseUrl per environment in aurelia.json file, or in environment.js file. Failing that, then ability to pass in a value via CLI parameter, so that the value can more easily be substituted on deployment

* **What is the motivation / use case for changing the behavior?**
Make it easier and more versatile to be able to deploy the same build definition to different environments such as in a CI/CD pipeline.

**Example:**
See below for the workaround we are currently having to do in our store multiple value in aurelia.json and then use Powershell to change the value.

![baseissue](https://user-images.githubusercontent.com/783486/28171309-e3e4b0f6-67df-11e7-9e8e-fe6f3ac78508.png)

magnusdanielson:
I have created a pullrequest that can solve this issue if you are using require. https://github.com/aurelia/cli/pull/686

Add the following to the build task in your project.

`const env = CLIOptions.getEnvironment();
  const envContent = fs.readFileSync(`aurelia_project/environments/${env}.ts`, 'utf8');
  const providerHostUrl = /providerHostUrl: '(.*)'/ig.exec(envContent)[1];
  project.paths.customBaseUrl =  providerHostUrl;`

Then just add below to you environment file
`providerHostUrl: 'https://dualite/',`

To be clear. The pull request makes i easy to use absolut url that is different from the hosting page. The change of the build task above makes it possible to override the url for different environements.

magnusdanielson:
So one of my environment files looks like below.
The providerHostUrl is the interesting configuration.

![image](https://user-images.githubusercontent.com/12081481/28891851-2894e652-77cc-11e7-9aab-657ed4c11569.png)

magnusdanielson:
And yes, before anyone asks. We use another task to move the output files to a ASP.Net MVC project with below task.

`import * as gulp from 'gulp';
import * as changed from 'gulp-changed';
import * as project from '../aurelia.json';

export default function sp() {
[...]

PR #765, opened 2017-10-04
feat(cli build): add build targets

jwx:
Configure build target folders in aurelia.json that builds multiple, complete distribution folders.

Changes:
- complete build is made into target root folder based on env (default: build/dev)
- source files in src are never touched
- source file environment.js/ts is never copied into src
- files index.html and favicon.ico should always be in root directory
- build command have a new --target flag 

aurelia.json:
- build.targets changed from array to object with target names as keys
- target object in build.targets has the following properties:
    + platform (string): Specifies platform (currently web or aspdotnet)
    + displayName (string): Friendly target name
    + root (string): Target root, relative to project root, default: 'dist/${TARGET}/${ENV}'
    + index (string): Name of target index file, default: 'index.html'
    + output (string): Target folder for bundles, relative to target root, default: 'scripts' (leave it alone)
    + indexSource (string): Name of source index file for target, default: uses target's "index" property (above)
    + resources (array): Unbundled resources copied from project root to target root, default:
      [
        'favicon.ico',
        'scripts/**', // This folder isn't used by the bundles anymore, so these are any additional, non-bundled scripts
        'images/**',
        'styles/**'
      ],
    + environments (object): Used to override target settings for environment (dev/stage/prod), default: {}
- if target object in build.targets is a string instead of an object, it serves as an alias for &-separated targets ("ALL": "target1 & target2")
- build.defaultTarget added that specifies default target when no --target flag is specified
- build.defaultEnvironment added that specifies default environment when no --env flag is specified

Replaces PR aurelia/cli#644.
Closes aurelia/cli#200, aurelia/cli#442, aurelia/cli#563, aurelia/cli#627 and aurelia/cli#677.
Partially closes aurelia/cli#763 and aurelia/cli#724.

### DIFF
--- a/lib/cli-options.js
+++ b/lib/cli-options.js
@@ -1,5 +1,8 @@
 'use strict';
 
+const path = require('path');
+const readFileSync = require('./file-system').readFileSync;
+
 exports.CLIOptions = class {
   constructor() {
     exports.CLIOptions.instance = this;
@@ -12,13 +15,78 @@ exports.CLIOptions = class {
     return parts.join('.');
   }
 
+  getProject() {
+    return JSON.parse(readFileSync([process.cwd(), 'aurelia_project', 'aurelia.json'].join('/')));
+  }
+
   getEnvironment() {
+    const project = this.getProject();
     let NODE_ENV;
-    let env = this.getFlagValue('env') || (process.env.NODE_ENV ? NODE_ENV = process.env.NODE_ENV : undefined) || 'dev';
+    let env = this.getFlagValue('env') || (process.env.NODE_ENV ? NODE_ENV = process.env.NODE_ENV : undefined) || project.build.defaultEnvironment || 'dev';
     if (NODE_ENV) {
       console.log(`The selected Node Environment (${NODE_ENV}) is not a preconfigured option ('dev', 'stage', and 'prod')`);
     }
     return env;
+  }
+
+  getTarget(targetId, env) {
+    const project = this.getProject();
+    const targets = project.build.targets;
+    targetId = targetId || project.build.defaultTarget || 'web';
+    env = env || this.getEnvironment();
+
+    let target;
+    if (Array.isArray(targets)) {
+      target = targets[0];
+      target.legacy = true;
+    }
+    else {
+      if (!targets[targetId]) {
+        console.log(`No target "${targetId}" found in configuration!`);
+      }
+      target = Object.assign({}, targets[targetId], targets[targetId].environments[env]);
+    }
+
+    if (target.root) {
+      target.root = target.root.replace('${ENV}', env).replace('${TARGET}', targetId);
+    }
+    else {
+      target.root = path.join(target.output, '..');
+    }
+
+    return target;
+  }
+
+  getTargets() {
+    const project = this.getProject();
+    const targets = project.build.targets;
+    const target = this.getFlagValue('target') || project.build.defaultTarget || 'web';
+    const env = this.getEnvironment();
+    if (Array.isArray(targets)) {
+      return [ this.getTarget() ];
+    }
+
+    let targetIds = this.resolveTarget(target);
+    return targetIds.map((value) => this.getTarget(value, env));
+  }
+
+  resolveTarget(targetId) {
+    const project = this.getProject();
+    const targets = project.build.targets;
+
+    if (!targets[targetId]) {
+      console.log(`No target "${targetId}" found in configuration!`);
+      return [];
+    }
+
+    const target = targets[targetId];
+    if (typeof target !== 'string') {
+      return [ targetId ];
+    }
+
+    let parts = target.split('&').map(x => x.trim());
+    let targetIds = parts.reduce((accumulated, current) => accumulated.concat(this.resolveTarget(current)), []);
+    return targetIds.filter((current, index, arr) => arr.indexOf(current) == index);
   }
 
   hasFlag(name, shortcut) {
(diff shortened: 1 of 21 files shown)


lib/cli-options.js
```diff
@@ -12,15 +15,80 @@ exports.CLIOptions = class {
...
     return env;
   }
 
+  getTarget(targetId, env) {
+    const project = this.getProject();
+    const targets = project.build.targets;
+    targetId = targetId || project.build.defaultTarget || 'web';
+    env = env || this.getEnvironment();
+
+    let target;
```
johan-v-r: Perhaps I'm missing something, but is this always going to return the first target if the `project.build.targets` is in fact an array?

Do you perhaps have an example of what the structure should look like?
My `aurelia.json` currently looks like this:
```
"build": {
    "targets": [
      {
        "id": "web",
        "displayName": "Web",
        "index": "index.html",
        "baseDir": ".",
        "output": "scripts"
      },
      {
        "id": "kiosk",
        "displayName": "Kiosk",
        "index": "kiosk.html",
        "baseDir": ".",
        "output": "kiosk_scripts"
      }
    ]
```

lib/cli-options.js
```diff
@@ -12,15 +15,80 @@ exports.CLIOptions = class {
...
+    if (Array.isArray(targets)) {
+      target = targets[0];
+      target.legacy = true;
```
johan-v-r: Perhaps I'm missing something, but is this always going to return the first target if the `project.build.targets` is in fact an array?

Do you perhaps have an example of what the structure should look like?
My `aurelia.json` currently looks like this:
```
"build": {
    "targets": [
      {
        "id": "web",
        "displayName": "Web",
        "index": "index.html",
        "baseDir": ".",
        "output": "scripts"
      },
      {
        "id": "kiosk",
        "displayName": "Kiosk",
        "index": "kiosk.html",
        "baseDir": ".",
        "output": "kiosk_scripts"
      }
    ]
```

jwx: @johan-v-r You're not missing something; that is the current behaviour (which is kept so that it's a non-breaking change).

Here's a (somewhat long) example:
```
    "targets": {
      "web": {
        "id": "web",
        "displayName": "Web",
        "index": "index.html",
        "output": "scripts",
        "root": "dist/${TARGET}/${ENV}",
        "resources": [
          "favicon.ico",
          "images/**",
          "styles/**"
        ],
        "environments": {
          "dev": {
            "root": "build/dev"
          }
        }
      },
      "aspnetcore": {
        "id": "aspnetcore",
        "displayName": "ASP.NET Core",
        "index": "index.html",
        "output": "scripts",
        "root": "wwwroot",
        "resources": [
          "favicon.ico",
          "images/**",
          "styles/**"
        ],
        "environments": {
          "prod": {
            "root": "../dotnetcore/wwwroot"
          }
        }
      },
      "cordova": {
        "id": "cordova",
        "displayName": "Cordova",
        "index": "index.html",
        "indexSource": "index.cordova.html",
        "output": "scripts",
        "root": "../cordova/www",
        "resources": [
          "favicon.ico",
          "images/**",
          "styles/**"
        ]
      },
      "EXTERNAL": "aspnetcore & cordova",
      "ALL": "web & EXTERNAL"
    }
```
In this example the Aurelia project also contains a dotnet core web server/project (not really a good idea, but it's just an example) that will be targeted for dev and stage builds. In addition, the Aurelia project has two "sibling projects", one Cordova and one that's, again, dotnet core (which is targeted for the prod build). The Cordova project has additional requirements for index.html, so it's a separate file in the Aurelia project (that will be renamed to index.html for the target).

Running 
`au build --target ALL --watch`
will rebuild and update all three targets as soon as a file is changed.

johan-v-r: Okay yeah thanks this is awesome, will 100% work for me. This also replaces my PR #731   

Just a final note from my side, looks like `root` property is more like an output directory..? Can/should this name perhaps be changed?

jwx: @johan-v-r Well, it is the target root and the base for the other target settings. (And, depending on them, nothing might actually end up there.) Plus, I didn't want it mixed up with the current `output` (which is where the script bundles are put). But yeah, perhaps it should be changed (but then `output` also needs to be changed). Naming things is often the hardest part of this stuff...

EisenbergEffect:
Ping @JeroenVinke to review.

The1nternet:
I'm highly interested in this feature and I want it released!  :)

lib/file-system.js
```diff
@@ -46,6 +46,23 @@ exports.mkdirp = function(path) {
   });
 };
 
+exports.mkdirSync = function (path) {
```
JeroenVinke: does this create a folder structure if necessary? Nice

lib/file-system.js
```diff
@@ -46,6 +46,23 @@ exports.mkdirp = function(path) {
...
+  if (fs.existsSync(path)) {
+    return;
+  }
+  let arr = path.split(/[\\\/]/);
+  let folder = arr.shift();
+  if (folder.indexOf(':') !== -1) {
+    folder += '/';
```
JeroenVinke: could you add some comments here?

lib/file-system.js
JeroenVinke: also, have you tested this on linux as well? If not I can do that

Jenselme:
Any progress on this? I'd like to see this feature.

avrahamcool:
@JeroenVinke 
any news on this?
if the whole PR is still not ready to merge, can you take the part that use the `environment` file without copying it to the `src` folder?
it feels like there a new version of the cli in the horizon, it can help a lot to have it there.

jwx:
I'll try and take a look at comments and merge conflicts this weekend.

3cp:
Closed the stale PR. Please reopen when it's needed.

PR #780, opened 2017-10-28
fix(gitignore): track vscode workspace files

gheoan:
Update according to https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore.


JeroenVinke:
Thank you @gheoan

PR #785, opened 2017-11-10
fix(javascriptservices): update configuration to set public path for static resources

grufffta:
The webpack build moves files into the /dist/ folder and resources are loaded from the root '/' publicPath
This adds a change to the webpack.netcore.config.js file that customizes webpack to work with javascript services by adding the publicPath to the url and file loaders.

Relates to a discussion on a previous PR #741


JeroenVinke:
thanks again @grufffta. Apologies for the very slow response on this

PR #793, opened 2017-12-06
Add new switch to create plugin

manuel-guilbault:
Add new switch to create plugins:
```shell
au new --plugin
```

### DIFF
--- a/build/tasks/generate-projects.js
+++ b/build/tasks/generate-projects.js
@@ -3,7 +3,8 @@ const gulp = require('gulp');
 const Container = require('aurelia-dependency-injection').Container;
 const definition = require('../../lib/commands/new/new-application.json');
 const WorkflowEngine = require('../../lib/workflow/workflow-engine').WorkflowEngine;
-const ProjectCreate = require('../../lib/workflow/activities/project-create');
+const ApplicationCreate = require('../../lib/workflow/activities/application-create');
+const PluginCreate = require('../../lib/workflow/activities/plugin-create');
 const UI = require('../../lib/ui').UI;
 const ConsoleUI = require('../../lib/ui').ConsoleUI;
 const fs = require('fs');
@@ -60,8 +61,10 @@ function createProject(answers) {
   container.registerInstance(UI, new ConsoleUI());
   container.unregister('project-install');
   container.registerInstance('project-install', new AutoProjectInstall());
-  container.unregister('project-create');
-  container.registerSingleton('project-create', AutoProjectCreate);
+  container.unregister('application-create');
+  container.registerSingleton('application-create', AutoApplicationCreate);
+  container.unregister('plugin-create');
+  container.registerSingleton('plugin-create', AutoPluginCreate);
   container.unregister('input-text');
   container.registerInstance('input-text', new AutoInputText(answers));
   container.unregister('input-select');
@@ -129,7 +132,13 @@ class AutoInputText {
   }
 }
 
-class AutoProjectCreate extends ProjectCreate {
+class AutoApplicationCreate extends ApplicationCreate {
+  projectConfirmation() {
+    return Promise.resolve({ value: 'yes' });
+  }
+}
+
+class AutoPluginCreate extends PluginCreate {
   projectConfirmation() {
     return Promise.resolve({ value: 'yes' });
   }
--- a/lib/commands/new/application/project-template.js
+++ b/lib/commands/new/application/project-template.js
@@ -1,0 +1,190 @@
+'use strict';
+const ProjectTemplate = require('../project-template').ProjectTemplate;
+const ProjectItem = require('../../../project-item').ProjectItem;
+const path = require('path');
+const fs = require('../../../file-system');
+const CLIOptions = require('../../../cli-options').CLIOptions;
+
+exports.ApplicationTemplate = class extends ProjectTemplate {
+  constructor(model, options, ui) {
+    super(model, options, ui);
+
+    this.resources = ProjectItem.directory('resources');
+    this.elements = ProjectItem.directory('elements');
+    this.attributes = ProjectItem.directory('attributes');
+    this.valueConverters = ProjectItem.directory('value-converters');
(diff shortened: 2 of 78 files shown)


JeroenVinke:
Yeah!! This is awesome, thanks so much for working on this. I'll have a close look next week

manuel-guilbault:
Once this is merged, it might be interesting to add official documentation on plugin authoring (unless there's already existing documentation which not aware of). Any idea on the subject?

EisenbergEffect:
We definitely want this feature, but I wonder if there's any way to consolidate the code with the app creation pieces. This seems like a lot of duplication.

manuel-guilbault:
I wanted to keep my initial PR as less intrusive as possible with the existing code, that's the reason for the code duplication, but I absolutely agree. I can refactor things around to consolidate similar parts of the code.

avrahamcool:
any news on this? can help a lot for creating new plugins

JeroenVinke:
@manuel-guilbault sorry for being away. I can take a look next weekend to see where we might be able to reduce duplication

manuel-guilbault:
@JeroenVinke no problem, I was also busy like crazy those past months. If you can take a look, that would be awesome, as I'm gonna have some bandwidth in a week or two, so I could get this work done.

JeroenVinke:
The workflow json is already quite complex so I wouldn’t be opposed to adding one specifically for creating plugins. It would be nice if we can reuse the ProjectTemplate and files that are in the buildsystems folder. If we need to deviate from the standard application setup, we could probably check for a certain property on the model which would be set by the plugin workflow json

If we need a different config file for lets say tsconfig.json then we can use the template functionality of the cli that we already use for webpack.conf.js, which uses prepocess. In the template we can perform an if-check on a property on the model, which again is set by the plugin workflow json

Let me know if you have any questions

Jule-:
Any update on this one?

Would love to use `au new myPlugin --plugin`! 😃

manuel-guilbault:
@Jule- I finished a project and then started a new job in the last months, so I was pretty busy and couldn't find time to work in this. Now that things are settling down a bit, I'll try to wrap this all up in the next week or so.

@JeroenVinke There's a `type` property on the model defined by the workflow json and passed to the `ProjectTemplate` class. This property is set to `project:application` in the `new-application.json` workflow file, and I set it to `project:plugin` in the plugin workflow file.

I think it would be possible to reuse the files in the buildsystems folder without too much problem, by checking this `type` property when distinct actions are required for app vs plugin. However, I feel like the `ProjectTemplate` class would end up full of `if`s, as it contains a lot of logic specifically related to application creation. I think one better solution would be to extract all support features (ex: methods such as `addToContent`, `addToSource` or `addToResources`) from `ProjectTemplate` in a base class, and extend it in new `ApplicationProjectTemplate` and `PluginProjectTemplate` classes, which would each drive the process of creating a single type of project.

What do you think?

Jule-:
@manuel-guilbault no problem, no pressure, just asking! 😉

But I really think we totally need a feature like this in order to maintain plugin homogeneity and quality. 😎

manuel-guilbault:
@Jule- I completely agree :)

Alexander-Taran:
thing is you just need a consistent starting point..
then you will surely deviate (-:

pick any
https://discourse.aurelia.io/t/aurelia-library-template/938

manuel-guilbault:
@JeroenVinke I just pushed an update with some refactoring.

I couldn't reuse as much code as I initially thought I would. I didn't reuse the files from buildsystems, because most of the files ended being like this:

```
if (project.type === 'project:plugin') {
  //configure options for a project
} else {
  //configure options for an application
}
```

As such, I didn't see much value in merging the plugin files with the buildsystems files.

I also couldn't reuse many of the files in resources/tasks, because the build process for a plugin is pretty different than that for an application:

* no bundler;
* multiple output formats (amd, commonjs, etc.) for both code (js, ts) and other assets (css, html).

As such, merging the tasks for applications with those for plugins was producing the same code pattern with if/else blocks, so I chose to leave them separated.

I did however extracted common behavior from ProjectTemplate into a base class, which I extended into an ApplicationTemplate and a PluginTemplate.

Let me know what you think.

manuel-guilbault:
I also have an issue where some dependencies added using the `addToDevDependencies` method don't get added to the package.json output. I have no clue as to most get added and some other are omitted.

JeroenVinke:
Not yet, my schedule is crazy at the moment. I’ll try to give you feedback next weekend

JeroenVinke:
Very grateful for the work you put in though! I’m still excited about this feature

JeroenVinke:
@manuel-guilbault i've had a look and I agree with your assessment. There's another large PR on the way so I need a bit of time to get this merged and released (I expect there to be some conflicts), I'll keep you up to date

manuel-guilbault:
@JeroenVinke If you want, I can take some time to rebase my work on master and make sure everything still works correctly. Maybe this could remove some work from your plate :)

jbockle:
wouldn't it be better to include a cli script for building a feature folder as a plugin? that way you have a dev environment included with your plugin

huochunpeng:
@jbockle https://github.com/aurelia-contrib/aurelia-getting-started/blob/master/guides/how-to-use-cli-to-develop-plugin.md
That works. And it can be a very small addon to current cli code base.

zewa666:
I remember having read this somewhere else but its definitly a good idea what @jbockle mentioned. Could have used this very well for the official plugins.

cottrellio:
I've been following this feature for over 2 years across multiple GH issues... Any chance we can +1 to get this pushed over the edge? It would be extremely useful to have this feature for the projects I'm working on as well as for the Aurelia community to more easily share code and be productive.

huochunpeng:
Thank @manuel-guilbault for the pioneering work on this. We now have a simplified implementation landed in cli master.
@EisenbergEffect this one can be closed in favour of simpler #1075.

manuel-guilbault:
@huochunpeng thanks!

PR #803, opened 2018-01-19
fix(dependencies) add gulp and typescript to the dependencies

### LINKED ISSUE #801: gulp is a devDependency, when it should be a regular or peerDependency

* **Library Version:**
0.30.1

* **Operating System:**
OSX 10.12.6

* **Node Version:**
8.9.4

* **NPM Version:**
5.6.0

* **Browser:**
not important

* **Language:**
not important

* **Loader/bundler:**
not important

**Current behavior:**
when installing aurelia-cli, gulp is not installed alongside it, because it is listed in the `devDependencies`.
running commands like `au build` doesn't work in that environment, because they require gulp

gulp should be a regular dependency, or at least a peerDependency, because it is required to run commands like `au build`, and not only for the development of the aurelia-cli.

I have a specific postinstall-script that moves and links dependencies depenending on what is mentioned in the package.json of local modules. because of that, and because the aurelia-cli doesn't state gulp as a dependency, the aurelia-cli ends up using a gulp-version of a different local module that is incompatible.

also: the aurelia-commands seem to require gulp version 4.0.0, but version ^3.9.1 is listed in the devDependencies.

heikomat:
the aurelia-cli cannot be used in its entirety without having `gulp@^4.0.0` and `typescript` installed. The aurelia-cli depends on them at least at these places during `au build`:

**gulp** (fails with `Error: Cannot find module 'gulp'`)
https://github.com/aurelia/cli/blob/ceb78602240e521c9592ceef537ca5a47ccfb256/lib/commands/gulp.js#L19

**gulp@^4.0.0** (fails with `TypeError: Cannot read property 'apply' of undefined`)
https://github.com/aurelia/cli/blob/ceb78602240e521c9592ceef537ca5a47ccfb256/lib/commands/gulp.js#L72

**typescript** (fails with `Error: Cannot find module 'typescript'`)
https://github.com/aurelia/cli/blob/ceb78602240e521c9592ceef537ca5a47ccfb256/lib/project.js#L221

The dependencies of the package.json should reflect, that the aurelia-cli requires these dependencies to be used. This is what is implemented in this branch.

This fixes #801

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,16 +33,17 @@
     "aurelia-polyfills": "^1.0.0",
     "esprima": "^4.0.0",
     "glob": "^7.1.1",
+    "gulp": "^4.0.0",
     "npm": "^3.10.8",
     "npm-which": "^3.0.1",
     "preprocess": "^3.1.0",
     "rfc6902": "^1.2.2",
-    "semver": "^5.3.0"
+    "semver": "^5.3.0",
+    "typescript": "^2.0.0"
   },
   "devDependencies": {
     "aurelia-tools": "^0.2.4",
     "babel-eslint": "^7.1.1",
-    "gulp": "^3.9.1",
     "gulp-bump": "^2.7.0",
     "gulp-conventional-changelog": "^1.1.3",
     "gulp-eslint": "^3.0.1",


AshleyGrant:
@JeroenVinke any chance we can get this merged?

JeroenVinke:
I'd like to discuss this a bit. I might be overlooking something, but without having typescript and gulp installed (globally) I can run a new typescript app. Gulp and typescript are getting installed locally which seems to be enough.

That said, I have seen those errors before (`TypeError: Cannot read property 'apply' of undefined` and it not being able to find any module). I never really looked into it further because for me it only occurred when I `npm link`'d the aurelia cli, which is not something that you would do as a user. These errors are actually the reason for steps 5 and 6 in the readme [here](https://github.com/aurelia/cli#building).

Could you explain the specifics of the post install task that you're using? Perhaps we can get to the root cause of this

heikomat:
@JeroenVinke I can.

**TL;DR**: I don't install the aurelia-cli to the aurelia-project, but instead install it in a folder further up, and link it to the node_modules of the aurelia-project, therefore making the aurelia-cli unable to find the correct gulp-version, because it is not installed alongside the aurelia-cli, because it is not listed in its dependencies.

**Detailed version**:
The specific use case i have is as follows:

We have a modular project that consists of about 70 repositories. To manage all of them we have one repository that uses [meta](https://github.com/mateodelnorte/meta) [minstall](https://github.com/heikomat/minstall) to clone, install, link and build all these 70 repositories so that a developer can change something in one of these locally cloned folders and have the change take effect in our application.

This meta-repository is (in other words) the developer-setup for our modular application.

One of these 70 repositories is an aurelia-application, which is the only one that can't be built after the dependency-installation through minstall. The reason for that is, that in order to save (a **lot**) of time, minstall hoists the dependencies as far up as possible without creating conflicts. It also installes every dependency only once per conflicting version. The result is, that every dependency of every of these 70 modules is installed at least once somewhere: Either in the meta-repo-folder itself, or in one of the module-folders.

After the installation phase, minstall checks what dependencies are installed where, and what dependencies are required where, and links them accordingly. The result is, that every one of these modules has at the very least a node_modules folder, where all its dependecies are linked into.

What happens now with the aurelia-project is, that the aurelia-cli is not installed to the node_modules-folder of the aurelia-project, but to the node_modules of the meta-repo-folder, and then linked from there to the aurelia-project-node_modules. Because of that, the npm-commands in our package.json [here](https://github.com/process-engine/bpmn-studio/blob/develop/package.json#L14) see an aurelia-cli in their local node_modules, use that, but because that aurelia-cli is actually in a different folder, that aurelia-cli is **not** using the gulp-version of our aurelia-project, but some other gulp-version that just happened to be installed to the meta-repo-node_modules.

If the aurelia-cli would explicitly state that it requires these dependencies to work, then installing it to a different folder and then linking it wouldn't be a problem, because the `npm install aurelia-cli`-command would make sure, that some correct version of gulp and typescript are installed along with the aurelia-cli

JeroenVinke:
Thank you for the detailed description. I'll need to give this some thought

JeroenVinke:
let's do it. I've opened https://github.com/aurelia/cli/pull/857 which has a couple of additional changes

PR #804, opened 2018-01-20
fix(importer): support scoped packages

Alexander-Taran:
packages with names like @aspnet/signalr-client were treated like versioned

### DIFF
--- a/lib/commands/install/package-argument-parser.js
+++ b/lib/commands/install/package-argument-parser.js
@@ -42,16 +42,17 @@ class PackageArgumentParser {
   }
 
   parseVersioned(argument) {
-    let split = argument.split('@');
+    let versionIndex = argument.lastIndexOf('@')
+    
     return {
       argument: argument,
-      name: split[0],
-      version: split[1]
+      name: argument.substr(0,versionIndex),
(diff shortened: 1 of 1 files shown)


zewa666:
Great, so this PR should now properly support scoped packages. Do you have an example which you tested it with?

Alexander-Taran:
I was toying with `@aspnet/signalr-client`. 
And `au install` went perfectly well, but import failed. Because resolved package name was an empty string.
So I went on and ran cli with debugger and stepped through until I've found the culprit.
First thing that went wrong is the detection of a versioned package.
Whenever there is a @ symbol in package name  (even if it is the first symbol) - `hasVersion()` returns true. 
But changing only that didn't suffice because `parseVersioned()` would fail for versioned and scoped packages. So I changed the logic there to find the last occurrence of @ and split the string there.

Alexander-Taran:
Something else I can do to help merge this?

JeroenVinke:
If yoo could rename the commit to “fix(importer): support scoped packages” or something in that format, then it will show up in the changelog that is generated, and also in the release blog post. I’ll review it this week or next week, I appreciate your patience and contribution

Alexander-Taran:
done

Issue #312, opened 2016-09-10
some bundled imports behave differently in aurelia-karma vs. browser


fkleuver:
**I'm submitting a bug report**
- **Library Version:**
  0.19.0

**Please tell us about your environment:**
- **Operating System:**
  Windows 10
- **Node Version:**
  6.5.0
  
  
- **NPM Version:**
  3.10.6
  
  
- **Browser:**
  all
- **Language:**
  TypeScript 2.0.2

**Current behavior:**
CodeMirror underlying imports work fine with `au run` but with `au test` they error out with "Uncaught ReferenceError: CodeMirror is not defined". This prevents the test runner from starting.

It's this piece of code where it goes wrong:

```
if (typeof exports == "object" && typeof module == "object")
    mod(require("../../lib/codemirror"));
else if (typeof define == "function" && define.amd)
    define(["../../lib/codemirror"], mod);
else // **it enters here when running test, and it crashes**
    mod(CodeMirror);
```

When running in the browser, it enters the second statement `define(["../../lib/codemirror"], mod);` and it works.
When running the test, it enters the third statement `mod(CodeMirror);` and errors out.

For some reason `define` or `define.amd` is not defined (or not of the correct type) in the scope of those imports in the test runner, causing it to fall through to the third statement which does not work.

Manually replacing the piece of code above with simply this line directly in the bundle, fixes the problem:
`define(["../../lib/codemirror"], mod);`

**Expected/desired behavior:**

I expect my imports to behave the same way with `au run` as they do with `au test`. This particular scenario should either work with `au test`, or fail completely in both cases.
Probably `define` and `define.amd` should be defined so the test runner works.

Or alternatively - if I'm missing something obvious here - the docs need to include the piece of information that would allow me to properly configure my aurelia.json and/or tests.
- **Steps to reproduce:**

I have attached a minimal project that illustrates the problem:
[aurelia-testing-test.zip](https://github.com/aurelia/cli/files/465671/aurelia-testing-test.zip)
1. npm install
2. au run --env dev
3. open it in the browser -> the editor will display, no script errors occur
4. au test -> crashes
- **What is the motivation / use case for changing the behavior?**

The ability to have tests work out-of-the-box in more scenarios when aurelia.json is properly configured and works in the browser.

EisenbergEffect:
If you have a recommendation for how we can improve the Karma configuration, that would certainly help. I'm not sure off-hand why this would be happening.

freddykleuver:
[...]

PR #806, opened 2018-01-23
Update Karma to fix intermitten unit test failure

bowozer:
Because it contains import for aurelia-polyfills that essential to execute any of the app files. Or else, it may break with "Reflect.getOwnMetadata is not a function" error.
Especially when the unit test is huge.

The require js files seems to load in parallel and asynchronously. So when there
are so many unit tests, sometimes Karma load other js files first before
the setup.ts file, which make further execution may fail (but
sometimes successful). More unit tests will have higher chance for Karma
to be failed due to setup.ts is loaded/required later.

### DIFF
--- a/lib/resources/test/require.aurelia-karma.js
+++ b/lib/resources/test/require.aurelia-karma.js
@@ -70,7 +70,7 @@
 
   function requireTests() {
     var TEST_REGEXP = /(spec)\.js$/i;
-    var allTestFiles = ['/base/test/unit/setup.js'];
+    var allTestFiles = [];
 
     Object.keys(window.__karma__.files).forEach(function(file) {
       if (TEST_REGEXP.test(file)) {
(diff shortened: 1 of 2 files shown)


bowozer:
This is the first time i made Pull Request for this project. I have read the contribution guideline, but if i miss something or lacks anything, please let me know so that i could fix it.

Alexander-Taran:
@bowozer congratulations on your contribution being merged (-:

JeroenVinke:
@bowozer thank you! one tip, when creating the commit try to use the following format(s) for the commit message:
- `fix(<file>): <the change>` if you've fixed an issue
- `feat(<file>): <the change>`  if you added a new feature
while keeping the length of this message to a minimum. By using this format your change will show up in the changelog.md file and in the release notes on the blog.

Thanks again :)

Alexander-Taran:
@JeroenVinke is this by any chance related?
https://github.com/aurelia/cli/issues/312

JeroenVinke:
@Alexander-Taran the first one works with the current version of the CLI, the second one is a different problem I think. Thanks for helping out :+1:

Alexander-Taran:
bah.. karma.. cli.. mumbo jumbo.. (-:
I was just passing by 1st when you merged this one..
then did a search for karma.. in issues.. and picked on bundling issues with require..

de rien

PR #807, opened 2018-01-26
fix(rjs-ts): add baseUrl to tsconfig

AshleyGrant:
This allows non-relative paths to work. `baseUrl` is already set when you choose webpack and typescript.


AshleyGrant:
@JeroenVinke any chance we can get this fix merged so I can push my webpack analyzer feature?

JeroenVinke:
Yep, I've confirmed it works

PR #819, opened 2018-02-05
feat(webpack): add analyze flag to enable webpack bundle analyzer

AshleyGrant:
I haven't tested the code as written, I've extracted the necessary configuration from my own webpack build, so please run the necessary tests to make sure I haven't borked everything before merging!


Alexander-Taran:
If it will make it to default project structure, make sure that analyser won't be in interactive mode for production builds at least. Hangs CI

AshleyGrant:
It will only run if the `--analyze` flag is included in the command that launches the aurelia CLI.

PR #820, opened 2018-02-06
Fix/sort bundle prevent hash changes

### LINKED ISSUE #816: Bundled files are not alwasy concatenated in the same order causing rev to change

* **Library Version:**
0.32.0

* **Operating System:**
Linux (Fedora 27)

* **Node Version:**
v8.9.4

* **NPM Version:**
5.6.0

* **Browser:**
NA

* **Language:**
ESNext

* **Loader/bundler:**
RequireJS

**Current behavior:** When building a project with [build revision](http://aurelia.io/docs/build-systems/aurelia-cli#build-revisions) enabled, each time I build my project, I get bundles with differrent hashes (eg `dist/vendor-bundle-51213c0533.js` and `dist/vendor-bundle-f32b0be81c.js`) even if nothing changed in my project.

To reproduce:
- Clone the project `git clone --branch feat/enable-rev https://gitlab.com/arenaoftitans/arena-of-titans.git`
- Install dependencies: `yarn install`
- Build the project multiple times by running `npm run build -- --env staging` multiple times
- Look in the `dist` folder, you will see bundles with different hashes

I expect the hash of the bundles to be the same if my code doesn't change.

This is a more logical behavior and allow caching of bundles that haven't changed.

* **Analysis of the problem**
I managed to look into this problem and I found out that the bundles files of my application are not always appended in the same order (with sources defined like `"[**/services/*.js]",`). I saw that by diffing two bundles. I guess this is because the globs aren't always resolving files in the same order. I also found out that the list of the files included in a bundle is not always in the same order, causing the hash of the vendor bundle to change erratically.

I managed to correct `lib/build/bundle.js` to correct this behavior here: https://github.com/Jenselme/cli/tree/fix/sort-bundle-prevent-hash-changes I can open a PR if you want.

Jenselme:
Close #816

### DIFF
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -136,7 +136,18 @@ exports.Bundle = class {
       });
     }
 
-    work = work.then(() => files = files.concat(this.getBundledFiles()));
+    // Sort files by path to be sure they will alway be concatenated in the same order.
+    let bundleFiles = this.getBundledFiles()
+      .sort((a, b) => {
+        if (a.path > b.path) {
+          return -1
+        } else if (b.path > a.path) {
+          return 1
+        } else {
+          return 0
+        }
+      });
+    work = work.then(() => files = files.concat(bundleFiles));
 
     if (this.append.length) {
       work = work.then(() => addFilesInOrder(this, this.append, files));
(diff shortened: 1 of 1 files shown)


JeroenVinke:
Thanks a lot!

Jenselme:
@JeroenVinke you're welcomed. Thanks for the merge.

PR #831, opened 2018-02-27
WIP fix(bundler): fix commonjs node id compatibility.

huochunpeng:
Commonjs thinks "foo" and "foo.js" are same id, but AMD thinks they are two different ids. Passing `nodeIdCompat: true` to requirejs config to improve compatibility to whole range of commonjs packages. Ref: https://github.com/amodrojs/amodro-trace/issues/7

### DIFF
--- a/lib/build/bundler.js
+++ b/lib/build/bundler.js
@@ -29,7 +29,8 @@ exports.Bundler = class {
       paths: ensurePathsRelativelyFromRoot(project.paths || {}),
       packages: [],
       stubModules: [],
-      shim: {}
+      shim: {},
+      nodeIdCompat: true
     };
     Object.assign(this.loaderConfig, this.project.build.loader.config);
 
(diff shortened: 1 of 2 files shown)


huochunpeng:
This solves date-fns bundling issue.

There is another issue related to date-fns.
`au import date-fns` doesn't work, the `importer/strategies/amodro` rejected it because of missing main in `package.json`.

I think it is quite safe to use `"index.js"` as the default value of `package.main`, because that's how nodejs deals with missing `"main"` anyway. @JeroenVinke I can make another PR if you agree to do so.

huochunpeng:
Sure, to ease my setup, I will create PR for that after this one is merged.

huochunpeng:
nodeIdCompact only fixes requrejs setup, not systemjs.

https://github.com/systemjs/systemjs/issues/1807

huochunpeng:
There is complication in test. All apps need to update aurelia-karma.js in order to run test normally.
Which means this is a breaking change.

Plus this doesn't help on systemjs setup.

I am working on a better solution.

huochunpeng:
obsoleted by #862

PR #838, opened 2018-03-06
fix(cli build): reassign global.define.amd after assigning karma over…

ryan-smith-virteom:
…ride function

aurelia/testing ComponentTester wasn't working. Bootstrapper failed to return with an error that there was no loader defined. Tracked the error down to a dependency reassigning window.require because define.amd was undefined.

aurelia/testing issue https://github.com/aurelia/testing/issues/71


JeroenVinke:
thank you!

PR #842, opened 2018-03-10
fix(importer): support nodejs module default main file "index.js" when "main" is missing in package.json

huochunpeng:
When "main" is missing in package.json, nodejs uses
default main file "index.js", it will complain when
default main file is missing.
This fix supports nodejs module with missing "main"
in package.json. It solves importing issue on nodejs
module "date-fns". #831 is related, that PR solves
tracing issue on "date-fns".


huochunpeng:
This fix also means amodro strategy will import absolutely any nodejs module. There would not be any nodejs module in the field with missing default main file, because those multifunctional modules would not reach your ears.

Whether we can trace what we imported, that is another problem.

PR #853, opened 2018-03-22
fix: removing "Tracing..." console logging

genne:
Unnecessary logging that clutters the console and makes it harder to spot the real issues you might have.

For example, we have ~200 modules in our project, meaning we get ~200 lines of log every time the bundles are regenerated, which happens every time you change a file (triggered by `gulp.watch`).

### DIFF
--- a/lib/build/bundled-source.js
+++ b/lib/build/bundled-source.js
@@ -64,8 +64,6 @@ exports.BundledSource = class {
     let that = this;
     let modulePath = this.path;
 
-    console.log(`Tracing ${moduleId}...`);
-
     for (let i = 0, ii = loaderPlugins.length; i < ii; ++i) {
       let current = loaderPlugins[i];
 


zewa666:
I wouldnt remove it as its sometimes quite useful. Instead this logs should be configurable via aurelia.json

genne:
Ok, I can probably fix that.

Out of curiosity, what is the use case for when you need that logging? Only for debugging purposes?

zewa666:
Yep solely debugging, like figuring out where it stops or what parts are missing and not properly traced

JeroenVinke:
@genne I like the idea of being able to configure the output of `au build`. Keep us up to date on the progress!

huochunpeng:
@JeroenVinke this one can be closed.
Now the log is behind `au build --debug` or `au run --debug`.

PR #857, opened 2018-03-26
fix(package.json): added missing deps

JeroenVinke:
closes https://github.com/aurelia/cli/issues/801
supesedes https://github.com/aurelia/cli/pull/803


huochunpeng:
I think you need to also trim package.json for new project template, as gulp/typescript/... are not needed anymore. And also to avoid future version conflicts on thoses packages.

huochunpeng:
For #864 for uglify-es. Same thing, new project template needs cleanup.

PR #861, opened 2018-03-29
fix(webpack): remove json loader from webpack config

### LINKED ISSUE #860: Json loader failed to import aws-amplify json files

* **Library Version:**
0.33.1

* **Operating System:**
Ubuntu 16.04

* **Node Version:**
8.10.0

* **NPM Version:**
3.10.4

* **Browser:**
all
* **Language:**
TypeScript 2.7.2

* **Loader/bundler:**
Webpack

**Current behavior:**
You can create a new cli project and choose typescript and webpack defaults.

Install aws-amplify, ```yarn add aws-amplify```

In ```main.ts```, ```import Amplify from 'aws-amplify'```.

And call ```Amplify.configure({...})```.

Running ```au run``` fails with json errors,
```
ERROR in ./node_modules/aws-sdk/lib/region_config_data.json
Module parse failed: Unexpected token m in JSON at position 0
You may need an appropriate loader to handle this file type.
SyntaxError: Unexpected token m in JSON at position 0
    at JSON.parse (<anonymous>)
    at JsonParser.parse (node_modules/webpack/lib/JsonParser.js:15:21)
```
Following the above steps should compile the app correctly.

aws-amplify should work with aurelia cli.

So I removed the json loader line from ```webpack.config.js```, the app compiled and I could use the amplify library. Webpack docs also say that json-loader isn't required for webpack 2 and above. If this is the correct fix, I can send a PR. Otherwise, pls suggest what I might be missing. Thanks.

lovetostrike:
closes #860

### DIFF
--- a/lib/resources/content/webpack.config.template.js
+++ b/lib/resources/content/webpack.config.template.js
@@ -122,7 +122,6 @@ module.exports = ({production, server, extractCss, coverage, analyze} = {}) => (
       // @if transpiler.id='typescript'
       { test: /\.tsx?$/, loader: "ts-loader" },
       // @endif
-      { test: /\.json$/i, loader: 'json-loader' },
       // use Bluebird as the global Promise implementation:
       { test: /[\/\\]node_modules[\/\\]bluebird[\/\\].+\.js$/, loader: 'expose-loader?Promise' },
       // embed small images and fonts as Data Urls and larger ones as files:


AshleyGrant:
A simple fix, but super necessary due to changes by the webpack folks. Let's get this merged!

Issue #581, opened 2017-04-02
failure to handle the "browser" field in dependencies' package.json


gravsten:
**I'm submitting a bug report**

* **Library Version:**
0.27.0

**Please tell us about your environment:**
* **Operating System:**
Windows 8.1

* **Node Version:**
6.10.0

* **NPM Version:**
3.10.10

**Current behavior:**
Several libs show a "browser" field in their package.json, which can be of several forms :
1) alternative "main", e.g. "index-browser"
2) arguments, e.g. leveldown = false

aurelia-cli ignores this information, thus requiring the same dependencies as the node version... which inevitably fails.
As a result, users have to edit aurelia.json to add a "main" section, or worse edit the library to hard-code any arguments that change its behaviour for the browser.

**Expected/desired behavior:**
The "browser" section of package.json should be taken into account when loading and tracing a dependency module.

* **What is the motivation / use case for changing the behavior?**
Make the inclusion of certain libraries (e.g. pouchdb-adapter-memory) seamless, save the developers' time, and more importantly bring peace and happiness among all nations and people.

gravsten:
About 10% of the dependencies in a typical setup, and
**exactly 5,7% of the dependencies in the base Aurelia project** created with 'au new' and the TypeScript option, present a "browser" section in their package.json

aurelia-cli 0.27.0 fails to handle the "browser" section and, considering the above, this post is more like a **bug report** than an enhancement request.

JeroenVinke:
Thanks for opening these issues, all of them should be supported by the CLI. Just because they are flagged as enhancements, does not necessarily mean that they do not take priority over issues flagged as bugs. In fact, improving the bundler is a priority of mine, and the issues you've opened will help a lot, so thanks!

JeroenVinke:
Example package: https://github.com/mzabriskie/axios

Deterus:
Adding another example package in case multiple instances were useful. https://www.npmjs.com/package/uuid

PR #862, opened 2018-03-30
feat(bundler): auto tracing for requirejs/systemjs

### LINKED ISSUE #579: failure upon require('./file.js') or require('module/')

* **Library Version:**
0.27.0

* **Operating System:**
Windows 8.1

* **Node Version:**
6.10.0

* **NPM Version:**
3.10.10

**Current behavior:**
Tracing fails to find appropriate file when using require('./file.js') or require('module/')

aurelia-cli should add '.js' only when omitted in the library code, and remove any trailing '/'.

The purpose of using node/npm is to reuse and automatically update libraries in one's project.
With the aforementioned bug, one needs to edit all offending libraries for 'au run' to work flawlessly (while other bundlers show no such problem).

huochunpeng:
remove all lib/importer files.
remove most of amodro-tracing.
manual efficient js and html tracing for aurelia.
remove irrelevant tracing cache.
cover all aurelia-webpack-plugin covers, plus more edge cases.
more permissive, usage of PLATFORM.moduleName is NOT required.
almost zero manual maintenance of aurelia.json dependencies.
fix node id compatibility issue #831 for both requirejs and systemjs.
fix bundles moduleIds format for modules need plugin.
fix compatibility issue with yarn workspaces.
support npm package "main":"./lib" when main file is lib/index.js.
improve backwards compatibility by ignoring main:false.
simplify and reduce source code, but increase test coverge.
add "au run --watch --auto-install" for frictionless dev flow.
best guess on package manager 'yarn' to avoid 'npm i' error on yarn
controlled node_modules.
overtake webpack on simplicity, usability, compatibility and flexibility,
but probably not on performance.
perform slightly faster than existing amodro-trace + cache.

Closes #579, #581, #872
Supersede #853, #842, #831
BREAKING CHANGE: require minor user code change to support non-js main, like "main": "font-awesome.css"
BREAKING CHANGE: remove support of undocumented "main": false, replace with generic "lazyMain": true. But this is handled transparently without breaking user's existing app.


huochunpeng:
### Appreciate if you wanna help testing auto tracing. Please report failures as well as success.

#### 1. For existing app (with `requirejs`/`systemjs`)

* update `package.json` with `"aurelia-cli": "git+ssh://git@github.com/huochunpeng/cli.git#at2",`.
* fresh rebuild of your `node_modules`.

> at2 is a tag, I will keep updating it with future bug fixes.

At this point, you app should work without any modification on `aurelia.json` file. Unless you used some features mentioned in breaking changes (more info at end of this post).

* test `au run --debug`, `au run --watch --debug` and `au build --debug`.

Now have a look of your `aurelia.json` `"dependencies"`. You only need to keep following, as they are not explicitly required by code. Plus anything that use `"deps"` and(or) `"exports"`, those are for shim on old js libs.

```js
"aurelia-bootstrapper",
"aurelia-loader-default",
"aurelia-pal-browser",
{
  "name": "aurelia-testing",
  "env": "dev"
},
"text",
// or for systemjs
// {
//   "name": "text",
//   "path": "../node_modules/systemjs-plugin-text",
//   "main": "text"
// }
```

This doesn't mean you really need to keep all manual config for shim. For instance, you can **totally remove `bootstrap` v4, `jquery` and `popper.js`**. Yes, dump them! Because they are very disciplined js libs, there are enough information in their code to enable auto tracing and shim.

For other shims, you can test one by one.

I recommend **DON'T use shim at all**, use `prepend` section before `requirejs` to deal with legacy js libs. Shim config is not worth you time, `prepend` is way easier.

#### 2. For testing `au new`

* `npm i -g git+ssh://git@github.com/huochunpeng/cli.git#at2`
* run `au new a_proj` with custom mode to choose either `requirejs` or `systemjs`.
* don't let cli to install npm packages, since you need to touch `package.json` file.
* update `package.json` with `"aurelia-cli": "git+ssh://git@github.com/huochunpeng/cli.git#at2",`.
* now `npm i` or `yarn`.

Now you have chance to test the sickest feature of them all, auto-install.

> auto-install also works on your existing app. Beware you cannot test auto-install with npm link, as any install will destroy your local symlink on aurelia-cli.

Run `au run --watch --auto-install --debug`.

You can now update `app.html` to include `<require from="bootstrap/css/bootstrap.css"></require>` (or "bootstrap/dist/css/bootstrap.css", both work). Watch it brings up npm package `bootstrap`, and see the font change on "Hello World!".

At this point, `jquery` and `popper.js` are not installed. Because `bootstrap` css file doesn't require them to work. By default, main file (`bootstrap/dist/js/bootstrap.js`) is lazy loaded.

You can now update `app.js` to add `import 'bootstrap';`. Auto tracing will find and install `jquery` and `popper.js` requested by `bootstrap` js file.

You can try `au.use.plugin('bla')`, you don't even need to use `PLATFORM.moduleName()` (it works too).

I call this **frictionless dev flow**. You can try any npm package compatible to run in browser, including those won't work with old cli.

---

#### 3. Notes on breaking changes

3.1 If you used undocumented `"main": false` for package `foo`, just remove the whole config for `foo`.

The default behaviour on main file:

* lazy loading for any package without explicit config.
* eager loading for package with explicit config.
* you can use `"lazyMain": true` in explicit config to force lazy loading.

3.2 If you used non-js main file, like font-awesome v4 https://stackoverflow.com/questions/39271458/how-can-i-add-font-awesome-to-my-aurelia-project-using-npm

```js
{
  "name": "font-awesome",
  "path": "../node_modules/font-awesome/css",
  "main": "font-awesome.css"
}
```
The following require tag would not work. If auto-install is on, it got side effect to install npm package `font-awesome.css` (which actually exists!).
```html
<require from="font-awesome.css"></require>
```

Not because I cannot make it work, it's because I really don't want to make it work.

To fix it:

* option 1, keep the dep config, change require to `<require from="font-awesome/font-awesome.css"></require>`. You need to update `copyFiles` to adjust fonts folder.
* option 2 (better), remove the dep config, change require to `<require from="font-awesome/css/font-awesome.css"></require>`. You need to update `copyFiles` to adjust fonts folder.
* option 3 (better), if you use scss, let scss to deal with it `@import "../node_modules/font-awesome/css/font-awesome";`, this happens before bundler. You still need `copyFiles` for fonts.
* option 4 (best), use cdn, stop thinking about build and where are the fonts.

huochunpeng:
### Technical Details

_Acknowledge `aurelia-webpack-plugin` and `amodro-trace` on guiding me what I need to do with tracing aurelia and tracing js._

First, most important, you don't need any deep knowledge to understand how this is possible.

We had explicit dependencies config, plus amodro-trace in cli. If you use your instinct to think about them for a while, it's not hard to realize they are mathematically redundant. Because, tracing, by definition, is to find depedencies automatically.

We can either removing tracing, which I did before #431. Or removing explicit config, which I did now. Obviously, this one is better.

#### all lib/importer are gone.

Webpack doesn't need them. Neither do I.

I only use `"main"` in `package.json`, no more need of `"browser"`, `"jspm"`, `"aurelia"` sections. This means we use `CommonJS`/`UMD` all way through. cli wraps them into `AMD` when needed.

#### amodro-trace is gone, only kept parser and transformers. Requirejs runtime is gone.

amodro-trace doesn't match our use case. Even with trace cache, there is still lots of unneeded repeated deep tracing.

I implemented tracing using amodro-trace parser, plus my own parser to deal with aurelia js/html files. cjs transformer is kept to translate `CommonJS` into `AMD` format. Write transformers `defines` and `stubs` are kept, but modified to not rely on help from requirejs runtime.

#### `lib/build/find-deps.js` finds all aurelia dependencies

Basically, it does what aurelia-webpack-plugin does. Plus forgive users didn't use `PLATFORM.moduleName()`.

#### dep config path and main is simplified

When there is no explicit config, dep config is built simply as

```js
{
  "name": "foo",
  "path": "../node_modules/foo",
  "main": "dist/commonjs/foo"
}
```

Notice I kept `dist/commonjs` in the `"main"`, not playing trick on `"path"`.

The `dist/commonjs` is handled later by alias.

```js
define('foo/foo', ['foo/dist/commonjs/foo'], function(m){return m;});
```

#### bundle/bundler/dependency-inclusion/source-inclusion/bundled-source are simplified

They are bit too complex to my taste, very difficult to test. The only reason I kept them, is that I want to maintain backward compatibility. Test coverage is increased.

Usage of `[]` in pattern `[**/*.js]` is only kept for backward compatibility, the brackets don't change any logic, it doesn't turn of deep tracing.

#### manual aliases for nodejs id compatibility #831

Requirejs has a option `nodeIdCompat` to treat `foo/bar.js` and `foo/bar` as same module, but systemjs doesn't have similar option for `AMD` code.

Instead of using requirejs `nodeIdCompat`, I manually create aliases like:

```js
define('foo/bar.js', ['foo/bar'], function(m){return m;});
```

Improved nodejs compatibility for both requirejs and systemjs.

huochunpeng:
TL;DR I don't like webpack much, so I wrote a replacement with far less code.

### My feeling towards webpack, tree-shaking, and tradition of engineering

I had rough concept of auto-tracing in cli for very long (check how old is #431). And I was never shy on expressing my dislike of webpack.

But as a very lazy person, I did nothing. Until webpack became cli's default, it finally pushed me to write some code to backup my mouth. As you can imagine, it cost me lots of spare time. But at the end, I learned so much more.

The reason I dare to challenge webpack is not because I am a rock star dev. I am a decent dev with middle level experience. I had zero experience on static code analysis before this project. Why I jumped in? Because I was taught about a simple fact before.

> "Everything around you that you call life, was made up by people that were no smarter than you."

In my naive view, what webpack can do is not impressive, how it does it is too impressive. The core business of webpack is just two things:

1. trace. Static analysis on code, find dependencies, bring them all in.
2. pack. Concatenate files together.

It also added lots of other features which I don't agree in engineering point of view. I have used webpack for one year when I was with React during 2015-2016. You can see, I left them both.

Webpack constantly reminds me how stupid I am, always having great trouble to do any trivial task. It brings back my terrifying memory on XML, SOAP and J2EE.

Instead of offending majority of the community (I know I am minority) on continue talking about webpack. Let me elaborate what I like with existing cli setup, and what I just made.

1. we don't need transpile plugin because gulp-babel solved it.
2. we don't need scss/less/** plugin because gulp-** solved it.
3. we don't need a kitchen-sink because gulp solved it.
4. HMR is not critical, because gulp-watch + browser-sync is good enough and predicable.
5. Tree-shaking is not critical, no need reasoning.
6. I read lots of code on static code analysis. And I think, man this is too tedious, I can make it way more intuitive with way less code. So I did.
7. I have to replace amodro-trace, not because it's not good. It's because it doesn't match Aurelia's use case. I made new tracing part very efficient, and very permissive.
8. I simplified the bundler code in compromise with maximum backward compatibility.
9. reported some edge cases back to aurelia-webpack-plugin and aurelia-loader-default. Discovered parser bug in amodro-trace (which it borrowed from requirejs).
10. implemented sort of webpack replacement (only for Aurelia right now) in far less code.
11. I don't need to create a new module loader. Requirejs is battle tested, with intuitive and flexible code splitting support.

#### In the end:

1. trace.
  * Static analysis on code, find depdencies.
    - [`lib/build/ast-matcher.js`](https://github.com/huochunpeng/cli/blob/auto-tracing/lib/build/ast-matcher.js)
    - [`lib/build/find-deps.js`](https://github.com/huochunpeng/cli/blob/auto-tracing/lib/build/find-deps.js)
    - [`lib/build/bundled-source.js`](https://github.com/huochunpeng/cli/blob/auto-tracing/lib/build/bundled-source.js)
  * bring them all in.
    - [`lib/build/bundler.js` `build()`](https://github.com/huochunpeng/cli/blob/d68771b308ecd862a26380368c7cff6b7a2c977a/lib/build/bundler.js#L139)
2. pack. Concatenate files together. I didn't change much on `lib/build/bundle.js` `write()`.

I leave rest of the job to gulp.

By doing very little, I am following the teaching I received on engineering. Stay focus, don't deviate. If you want to replace something, only replace it with something better.

There is 40 years Unix tradition (copied from wikipedia):

* Write programs that do one thing and do it well.
* Write programs to work together.
* Write programs to handle text streams, because that is a universal interface.

By doing very little, and do your part really well, you gain enormous flexibility. To me, gulp is the shell, cli bundler is a little program can be used in the shell.

This means you can [replace html template preprocessor with markdown](https://github.com/aurelia-contrib/aurelia-getting-started/blob/master/guides/how-to-support-markdown-preprocessing.md), or any other templating engine with ease. cli bundler doesn't even need to know about it.

That's the beauty of being small and liberal. And that to me, is the tradition of engineering.

The art is to do less, not more. It is the courage to say no, not say yes. But sadly, it seems, I am the minority now.

#### About tree-shaking

Tree-shaking is a neat feature, but not anyway close to critical feature in my understanding. Furthermore, it creates some new issues that should not exist. Now you need to escape tree-shaking's babysit sometimes.

```js
new ModuleDependenciesPlugin({
  'aurelia-testing': [ './compile-spy', './view-spy' ]
})
```

I did nothing about tree-shaking. But it doesn't mean my tracer is stupid. If you do:

```js
import map from 'lodash/map';
```

My tracer will only fetch partial lodash required by `_.map`. I believe tree-shaking can do same thing for you if you write `import _ from 'lodash';` but only use `_.map` in code.

My approach is to respect your conscious decision in code. For lib developers, there are many ways you can make your lib partially acquirable. Mind your craft, instead of rely on tree-shaking.

#### My take away from this project

I learned, as I expected, static code analysis is not that hard.

I learned, as I expected, I can implement webpack in far less code.

My ast-matcher is good enough to be stand-alone npm package to help others to do static code analysis. I will publish it.

I can use my experience to design a generic webpack alternative. I can make it very lean without thinking about compatibility of aurelia-cli. In the dozen of webpack alternatives, I might have a chance, because I do very very little, but do it really well.

But that is a big project. I am quite lazy, have zero commitment on it.

#### To conclude

I didn't write this post to convince you to join my way. I didn't write this to shame/bash majority of the community. I am sincerely sorry if I let you feel that way.

I write this post and all my code to simply follow my teaching I received when I was a teenage long long time ago.

> "To express oneself honestly, not lying to oneself -- that, my friend, is very hard to do." -- Bruce Lee

fkleuver:
Could you maybe put a `fork-aurelia-cli` or something along those lines on npm? This is quite a big change that needs lots of testing, so the easier you make it for people to test the more likely it is you'll get it tested sufficiently :)

huochunpeng:
Thx. Good idea. I will wait for few days, to let core team to decide how to proceed on testing.

BTW, I did not test on Windows for last 2 days, because I have no access to PC at home. And I have few changes. So if you can run some quick test on your PC, it would be great.

zewa666:
Thanks god im not the only one to dislike Webpack. Cool stuff hope I can find a few moments to give it a try. Main:false is heavily used to optimize lazy imports with e.g. RxJS. Essentially there is no need for any Main in that case and Imports happen manually. How would that be covered? LazyMain doesnt sound like its gonna help with this.

zewa666:
Btw an excellent article on rjs + hmr. This might inspire you for more cool stuff “Hot-reloading in 2017: updated and unabridged” @the1mills https://medium.com/@the1mills/hot-reloading-with-react-requirejs-7b2aa6cb06e1

EisenbergEffect:
Wow!!! 🥇  @JeroenVinke Do you have some time to dig into this?

bigopon:
Awesome attempt, much fun had, much learnt, much improvement, like 👍

huochunpeng:
@zewa666, for main: false, just remove the config. The default behavior is exactly what you need, no main packed unless explicitly required by your code.

huochunpeng:
Sorry about broken `at2` tag. I didn't test it after final code cleanup.
New tag `at3` was released, at least I tested on my Mac.

huochunpeng:
New tag `at4` is released for fixing aurelia-i18n problem.

huochunpeng:
With tag `at4`, I have now converted two of my production apps using it. So far, so good.
One app has 7k lines of code, another one has 46k lines of code. Both don't use shim, so my explicit dependencies are bare minimum.

Finally I can get my Easter back. More apps converting after this holiday.

huochunpeng:
@bigopon show me some real love with report on some big apps 😆 
Happy Easter!

huochunpeng:
New tag `at5` is released to fix auto-install endless loop on npm placeholders for popular typo like `boostrap`.

zewa666:
Auto installing typo packages might induce a security risk http://blog.npmjs.org/post/163723642530/crossenv-malware-on-the-npm-registry

huochunpeng:
Exactly. That is what I read from the readme of boostrap npm module.

Thx for the article.

So give a warning when using auto install? Any other option?

But user still can make exactly same typo with `npm install` or `yarn add`.

I don't think I can do anything more than to warn user about auto-install.

CasiOo:
I don't think a package installer has anything to do in a bundler. I would remove the package installer.

huochunpeng:
@CasiOo thx for the feedback. I have been considering the option, it is not an important feature to keep, just a showcase of extra flexibility of auto tracing. Nothing is set yet, but I sided on removing it.

Some additional notes:
* Users need explicitly turn on auto-install on command line if they want to use it, by default it's off.
* `aurelia-cli` already has installer built in, it can help to install dependencies after `au new`. So there is actually only very little code for me to support auto-install for `au run` (and `au build`).
* The auto-install also behaves differently from normal `npm install` or `yarn add`. A normal install only install the package and its dependencies specified in its `package.json`. But auto-install is based on code tracing, so it can auto install `jquery` and `popper.js` (they are peerDependencies in `package.json`) for `bootstrap`.

fkleuver:
Aurelia-cli has always been more than just a bundler anyway. I find it quite likable that there's some happy path of installing&configuring some plugin with a single command. This is unique as far as I know. With a bundler and installer in one (but without being dependent on one another) you can create a silky smooth user experience.

Probably not worth and time and resources to make it work for every single thing (if that's even possible) but surely at least the Aurelia plugins?

Just my 2c.

huochunpeng:
The auto-install should work for all npm packages that support commonjs or amd or umd. Packages only use global name space would not work nicely, like many old jquery plugins.

Auto-install also only installs current default stable version. If you need any special version, don’t use auto-install.

huochunpeng:
We will gather feedback on usability of auto-install. It might create some confusion.

If I am designing a bundler from scratch, I will give user a callback point on dep failure, so user can hook up anything to deal with missing dep, so user can write few lines of code to install any package on failure and then return a request to retry.

huochunpeng:
Two more apps (size 12k and 3k lines of code) of mine converted to use auto-tracing cli.
Plus the 2 I did previously, the 4 are all my production apps.

huochunpeng:
Tag `at10` finally brings back proper shim support. Sorry for taking so long, as I don't use shim in my apps at all.

huochunpeng:
It has been a while. Anyone tried auto-tracing on some non-trivial apps? Any success or failure?

the1mills:
@oresoftware see this

simonfox:
I've just started testing this as well and am having a similar issue with `ms-signalr-client` which is pretty much identical in content to the `signalr` package @fkleuver is using above....I'll report back with anything else I find after getting around that - tracing otherwise seems to work.

huochunpeng:
Awesome! Thx both.

@fkleuver for your local json plugin, you can probably get away with
```json
{
    "name": "json",
    "path": "../scripts/json",
    "packageRoot": "../scripts"
}
```

For `singalr`, whenever you see a error like `'http://.../src/foo/bar.js'`, it is because requirejs could not find module `foo/bar` in bundles. I had a look at the content of `signalr/jquery.signalr.js`, it's very old that doesn't support AMD or cjs, means you need to use shim to make it work.

I am surprised it works with old cli without shim config. Could you try this with latest `at11` release?
```json
{
  "name": "signalr",
  "path": "../node_modules/singalr",
  "main": "jquery.signalr",
  "exports": "jQuery.signalR"
}
```

Another way to make it definitely work is to use prepend instead of shim config, just prepend after jquery. This also means you need to remove `import 'signalr';` from you js code.

I will do some test on `singalr` package tomorrow, to see how it works on old cli.

BTW, for rest in your config without shim (`"deps"` and `"exports"`), just delete them. You only need some minimum config as my second post said.

```js
"aurelia-bootstrapper",
"aurelia-loader-default",
"aurelia-pal-browser",
{
  "name": "aurelia-testing", // note you don't need path and main, auto-trace auto fills them
  "env": "dev"
},
"text",
// or for systemjs
// {
//   "name": "text",
//   "path": "../node_modules/systemjs-plugin-text",
//   "main": "text"
// }
```

fkleuver:
> @fkleuver for your local json plugin, you can probably get away with
> ```
> {
>     "name": "json",
>     "path": "../scripts/json",
>     "packageRoot": "../scripts"
> }
> ```

Yup, got away with.

> Could you try this with latest at11 release?
> 
> {
>   "name": "signalr",
>   "path": "../node_modules/singalr",
>   "main": "jquery.signalr",
>   "exports": "jQuery.signalR"
> }

Yap, all runs now. Beautiful :)

huochunpeng:
@fkleuver @simonfox please try tag `at12`, it now should allow your app to build with original unmodified `aurelia.json`.

bigopon:
Was testing `at12` with `pnpm`, hard links don't seem to work as it couldn't be found, both version.

```
Error: Unable to find package metadata (package.json) of aurelia-binding
    at loadPackageMetadata.then (C:\Users\bigopon\Desktop\test\node_modules\.github.com\huochunpeng\cli\10edaef0ca222ab61efcaf141b28c8f8df268967\node_modules\aurelia-cli\lib\build\package-analyzer.js:19:17)
    at <anonymous>
```

huochunpeng:
Damn, another alternative of npm and yarn?! Hard yo keep up with nodejs landscape.

I will do some test tomorrow. Maybe related to hard link implementation on Windows. I am not Windows user, does the file system support real hard link? Or some kind of simulation?

bigopon:
I'm not sure, but I think so.

Just tested with aurelia ux, seems to be working nicely without any configuration. Very nice 👍

fkleuver:
@huochunpeng Yes, Windows has real hard links. But looking at their [link source](https://github.com/pnpm/supi/blob/master/src/link/index.ts) they're actually using symbolic links instead of hard links. So it's probably in effect similar to `npm link`

huochunpeng:
@bigopon I believe you cloned auto-tracing branch in local, then somehow linked to your app (pnpm link?).

I found out pnpm doesn't work with normal install of aurelia-cli (either my branch or stock version) due to #857. Current cli without that fix relies on flat structure of node_modules folder (npm >v3 or yarn).

pnpm with normal aurelia-cli install,
```
Error: Cannot find module 'babel-core'
```
Now, cli has no access to babel-core package, because it's NOT in cli's package.json dependencies. pnpm deliberately prevents cli to access a package not in its explicit dependencies.

Your error with pnpm could be another error under pnpm link use case, which is different from normal install (from my limited experience on npm link).

fkleuver:
I just tried pnpm - installing is fine (and super fast), but with linking I ran into one problem after the other on Windows.

It tries to be clever by hijacking the environment and putting its links somewhere higher up the PATH hierarchy because it doesn't have its own install. There are countless direct and indirect problems with that and I just wasted 45 minutes figuring that out and repairing my environment.

They should just remove `pnpm link` as a feature and stop claiming to be a stand-in replacement. Because they aren't, and will never be with their current bogus installation and linking technique. I'm sticking with npm and yarn. Speed isn't everything.

/rantmode over

@huochunpeng Everything worked with your latest `at12` release, without any modifications to my bundle config. Awesome job!

huochunpeng:
The problem with pnpm goes beyond just aurelia-cli package, with branch https://github.com/huochunpeng/cli/tree/auto-tracing-with-deps-fix, I duplicated @bigopon 's error on another aurelia package.
```
Error: Unable to find package metadata (package.json) of aurelia-loader-default
```

There is no local `node_modules/aurelia-loader-default` ! Why? because it's not in your app's package.json (dev)dependencies.

Those much needed `aurelia-*` packages are dependencies of `aurelia-bootstrapper` and `aurelia-framework`.

You need to explicitly put all of them in app's package.json, that is quite verbose, but neccessary to work with pnpm...

Too sad, all Aurelia apps rely on flat node_modules structure, pnpm is out of Aurelia world.

fkleuver:
FYI, according to their [FAQ](https://pnpm.js.org/docs/en/faq.html) this can be resolved easily with `pnpm install --shamefully-flatten`

Doesn't change the fact that pnpm is still out of Windows world TT

huochunpeng:
@fkleuver I shamefully used that switch, and pnpm worked on my Mac!

Only one fix (not pnpm problem) I need, to put `"source-map": "^0.5.7"` into my package.json, because aurelia-cli has error with source-map v0.6.x.

I got same issue with npm because the postcss package I use. npm resolve my source-map version to 0.6.1, but yarn resolve it to 0.5.7.

simonfox:
@huochunpeng looking good!

Found an issue with an old cli app around `aurelia-animator-velocity`....we managed to make it work using [the suggestion here](https://github.com/aurelia/animator-velocity/issues/13#issuecomment-245799650). But with auto trace that doesn't seem to be working...tried just removing that config as well....trace output
```
INFO [Bundler] Tracing files ...
INFO [Bundler] Auto tracing package: 1.2.0      aurelia-animator-velocity
INFO [Bundler] Auto tracing package: 6.1.4      highcharts
INFO [Bundler] Auto tracing package: 1.5.2      velocity-animate
ERROR [PackageAnalyzer] Unable to load package metadata (package.json) of velocity:
INFO [PackageAnalyzer] Error: cannot resolve npm package folder for "velocity"
ERROR [Bundler] Unable to analyze velocity
INFO [Bundler] Error: Unable to find package metadata (package.json) of velocity
ERROR [Bundler] Failed to add Nodejs module velocity
INFO [Bundler] Error: Unable to find package metadata (package.json) of velocity
```

huochunpeng:
@simonfox thx!
I checked velocity-animator, the v1 is really messed up. v2 fixed the module deps, but aurelia-animator-velocity doesn't work with v2 right now.

1. short-term fix with auto-tracing: (I will add this to the doc)

* Add this to aurelia.json, note this is all you need in aurelia.json.
```
{
  "name": "velocity-animate",
  "deps": ["jquery"]
}
```
* modify your `aurelia_project/tasks/build.js`, this bypass the messed up dep in velocity-animate/velocity-ui. Similar to what you got from the ref suggestion, but this one is easier.
```
function writeBundles() {
  return buildCLI.dest({
    onRequiringModule: function(moduleName) {
      if (moduleName === 'velocity') return "define(['velocity-animate'],function(m){return m;});";
    }
  });
}
```

2. long-term fix, update aurelia-animator-velocity to velocity-animator v2. Everything will be automatic by then. @EisenbergEffect who is the maintainer?

EisenbergEffect:
@huochunpeng Unfortunately, we haven't had a maintainer of that plugin for a while now. If you want to submit a PR with the fixes, I'd be happy to accept them and put out a release.

huochunpeng:
@EisenbergEffect the velocity v2 api seems changed quite a lot, especially the api to customise effect. I am not experienced enough on the subject to upgrade it. If the customise api requires a different shape of input, this will be a breaking change.

huochunpeng:
@simonfox btw, I just tried exact solution that you had in the ref sugguestion. It works in auto-tracing for me. What's your error with this setup?

scripts/velocity-shim.js:
```
define('velocity', ['velocity-animate'], function(velocity) {
  return velocity;
});
```
aurelia.json:

```
"aurelia-animator-velocity", // this line is not needed, but doesn't hurt
{
  "name": "velocity-animate",
  "path": "../node_modules/velocity-animate",
  "main": "velocity",
  "deps": ["jquery"],
  "resources": ["velocity.ui.js"]
},
{
  "name": "velocity",
  "path": "../scripts/velocity-shim"
},
```

simonfox:
@huochunpeng sorry for lack of detail..I'm getting a runtime error...

I've got the following `define` generated by auto trace which seems to be the cause of the problem:
```
define("velocity-animate/velocity", ["jquery"], function(){});
```

huochunpeng:
@simonfox wait a minute, can you check your velocity-animate version? v2 will cause a runtime error. I guess you have explicit velocity-animate in your app's package.json.

huochunpeng:
Thx @simonfox for catching the regression.

Issue #490, opened 2017-02-16
Consider allowing alternate minifier instead of uglify for better ESNext support


delebash:
**I'm submitting a feature request**

* **Library Version:**
0.24.0

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
7.50

* **NPM Version:**
4.0.5

* **Browser:**
Chrome 56

* **Language:**
ESNext

**Current behavior:**
UglifyJS does not support some ESNext features available in babel and current browsers such as Generators and Async Functions.

Both fail with error  'SyntaxError: Unexpected token operator «*», expected punc «(»',

**Expected/desired behavior:**
ESNext features supported by babel should also be supported by the aurelia-cli minification process.

To reproduce
Add a generator function to your code and run au build --env prod

**Possible solutions:**
If the minification process is extracted to a gulp task then the user can swap out uglify with 
[babili](https://github.com/babel/babili) which supports the same features as babel.

Uglify does have a [harmony version](https://github.com/mishoo/UglifyJS2/tree/harmony) which has better support for ESNext features but there is a long [discussion ](https://github.com/mishoo/UglifyJS2/issues/1181)on which might be better to use moving forward.  By extracting to a gulp task it gives the user the ability to choose.

AStoker:
I could be an extra layer of thick today, so bear with me if I'm overlooking something :) 
You're talking about the process of transpiling the code, right? That process takes place in the transpile task located inside your `aurelia_project/tasks` folder. In there, there's a gulp task that uses Babel to transpile. But that process is completely up to the user to change, convert, or do whatever to. If you don't want to use Babel, are you not simply able to swap out the transpiler there?

delebash:
No :) When you specify --env prod aurelia-cli runs uglify-js process which is baked into the cli, uglify is not a gulp task in aurelia_project/tasks.  If you look at `node_modules/aurelia-cli/lib/build/bundle.js`  you can see uglify is hard coded.

```
const UglifyJS = require('uglify-js'); //added to user project devDependencies
let minificationResult = UglifyJS.minify(String(contents), minificationOptions);
```

Yes you can use any transpiler you want with your own minification process via gulp tasks, but if you want to use the built in --env prod switch then aurelia-cli will use it's own minification process via uglify.

AStoker:
Okay, thanks for getting me straight.  
So the Babel task happens first, and then the uglification happens as the last step. So what babel is outputting isn't compatible with the uglifier built in, correct?  
[...]

PR #864, opened 2018-04-04
refactor(bundler): uses uglify-es in place of uglify-js

rodu:
With better support of ES6+ in modern browsers and the introduction
of babel-preset-env it's possible to serve ES6 code to clients able
to consume it. This change uses uglify-es in place of ugliify-js to
handle minification that supports ES6.

**uglify-es is API/CLI compatible with uglify-js@3.**
**uglify-es is not backwards compatible with uglify-js@2.**

Issue #490

### DIFF
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -291,7 +291,22 @@ exports.Bundle = class {
       }
 
       if (buildOptions.isApplicable('minify')) {
-        const UglifyJS = require('uglify-es');
+        let UglifyJS;
+
+        try {
+          UglifyJS = require('uglify-es');
+        } catch (e) {
+          logger.warn(
+            'uglify-es package not available: Attempting to use uglify-js.'
+          );
+          logger.warn(
+            'Consider adding the uglify-es package to your npm dependencies ' +
+            'if you intend to support ES6 code minification.'
+          );
+
+          UglifyJS = require('uglify-js');
+        }
+
         // fromString is not a supported option in v3
         const isV3 = !!(UglifyJS.minify('', {fromString: true}).error);
 
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -291,7 +291,7 @@ exports.Bundle = class {
       }
 
       if (buildOptions.isApplicable('minify')) {
-        const UglifyJS = require('uglify-js');
+        const UglifyJS = require('uglify-es');
         // fromString is not a supported option in v3
         const isV3 = !!(UglifyJS.minify('', {fromString: true}).error);
 
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -291,21 +291,7 @@ exports.Bundle = class {
       }
 
       if (buildOptions.isApplicable('minify')) {
-        let UglifyJS;
-
-        try {
-          UglifyJS = require('uglify-es');
-        } catch (e) {
-          logger.warn(
-            'uglify-es package not available: Attempting to use uglify-js.'
-          );
-          logger.warn(
-            'Consider adding the uglify-es package to your npm dependencies ' +
-            'if you intend to support ES6 code minification.'
-          );
-
-          UglifyJS = require('uglify-js');
-        }
+        const UglifyJS = require('uglify-es');
 
         // fromString is not a supported option in v3
         const isV3 = !!(UglifyJS.minify('', {fromString: true}).error);
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -8,6 +8,7 @@ const DependencyInclusion = require('./dependency-inclusion').DependencyInclusio
 const Configuration = require('../configuration').Configuration;
 const Utils = require('./utils');
 const logger = require('aurelia-logging').getLogger('Bundle');
+const UglifyJS = require('uglify-es');
 
 exports.Bundle = class {
   constructor(bundler, config) {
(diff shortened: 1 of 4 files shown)


Alexander-Taran:
well.. for libraries that also distribute es6 modules it can open a way..
aurelia webpack plugin allows to select which versions to use for bundle..
if there is a native es with async.. it would improve somehow the bundle size..

huochunpeng:
Just tried this patch on one of my apps. All works, harmless. Not surprising, given uglify-es and uglify-js are from same developer.

huochunpeng:
In long term, I think it's better to push minify and source-map out of cli, to rely more on gulp pipeline, if that's possible.

huochunpeng:
Hi @rodu this PR is a breaking change, existing app that has no 'uglify-es' in their package.json file will fail.

To make it smoother for people to upgrade, I suggest to use try-catch block to try `require('uglify-es')` first, then fallback to `require(`'uglify-js'`)`. That will make this not a breaking change.

Alexander-Taran:
@huochunpeng  it actually adds a dependency to locally installed cli. I think.

huochunpeng:
I am talking about exiting app upgrading to newer cli. That requires user to manually change local package.json.

huochunpeng:
To clarify, that require('uglify-es') is resolved at runtime, in user app's deps, not cli's deps.

Alexander-Taran:
@huochunpeng don't see what you mean.

huochunpeng:
Sorry, I missed that local install change. I did not test directly on this PR' code.

But this probably changes existing behavior, it means user does need to install uglify at all.

huochunpeng:
I guess @rodu added it to devDependencies because he ran into some trouble of local testing with npm link.

To bypass the limitation of npm link, use npm install —save-dev git+ssh://yourbranch.

In normal npm install, devDeps are ignored, which means this would not help end user’s app at all.

huochunpeng:
@rodu just read your option1 and 2 on #490, I think you need to move 'uglify-es' from devDeps to deps to really make option1 to work.

When you use npm install from git repo, npm pulls all deps and devDeps. But for a future normal release of cli, a normal `npm install aurelia-cli` only pulls down deps, not devDeps.

rodu:
@huochunpeng  thank you for your feedback. I should have posted my indications for option 1 and 2 here.

I agree that adding the uglify-es dependency to the CLI package.json changes the existing behaviour slightly. I did consider the possibility of using a try/catch to test the presence of the uglify-es package, and in case fallback to uglify-js, if that's missing. I pushed a [different branch to try that approach](https://github.com/aurelia/cli/compare/master...rodu:feat-uglify-es-fallback).

At first, it all seems to make sense, but then I am actually not able to test it properly. When I run the build in one of my existing projects, the CLI looks for uglify-es in the CLI node_modules, not in the project node_modules, even though I have one there.
It must be the use of `npm link aurelia-cli` that I am using to test things out. But I don't have such a deep knowledge of npm. If I physically copy the CLI folder to my node_modules, then it works, because it finds uglify-es in my project now.

I will push a commit to move the uglify-es from devDeps to deps as you suggested. Apart from that, should we also consider the try/catch approach?

huochunpeng:
You are welcome. `npm link` confused hell out of me before, the behaviour is different from normal install. So I stopped using it. I guess it's due to the symlink. Now if I want to test my local changes, I either install it from my github repo, or simply replacing my `node_modules/aurelia-cli` with a copy (not symlink) of my local `aurelia-cli` repo.

@JeroenVinke will have the final decision. Personally I think, for end users, try-catch is more friendly than adding it to deps.

Alexander-Taran:
If it settles with try catch - there should be a warning message maybe  in case configs target es6 modules..
in ts or babel.. don't know how to figure it out.

rodu:
Using the try/catch seems a better approach, allowing a fall-back configuration without adding dependencies to the CLI packages.

* New projects will use uglify-es by default.
* Existing projects will use what they currently have, but will get a warning by default if using uglify-js, just to inform users that they can switch to uglify-es if they want to support for ES6 minification.

Please advise on this approach and if you are happy with the warning messages.

The way to test the patch changes a little. Instead of using `npm link aurelia-cli` in a project it's better to install the patch directly with: `npm install rodu/cli#feat-uglify-es` and then run the build as usual.

Alexander-Taran:
@JeroenVinke need your opinion here

JeroenVinke:
Hey all :) thanks for working on this @rodu and your feedback @huochunpeng 

I think we might have to add uglify-es as a dependency of the cli due to https://github.com/aurelia/cli/pull/803#issuecomment-362900878. It does work to keep the uglify-es dependency in the package.json of the app but since the cli is the entity that's using uglify-es it should declare it in the package.json of the aurelia-cli

That would also remove the need for the try catch, but we'd need to make sure that switching to uglify-es is not a breaking change

huochunpeng:
I added some notes in #857 about new project template cleanup.

huochunpeng:
Switch to uglify-es might introduce side-effect to new tracer in my #862.

Not because uglify-es itself, which is after tracing. It is the transpiled code that can could be in newer syntax. I need to test more to make sure new tracer can (most likely) handle them.

rodu:
@huochunpeng to run your checks about the tracer making sure to be using ES6 output I would suggest you use this configuration for babel:

```JS
"presets": [
  ["env", { "targets": { "chrome": 62 } }],
  "stage-1"
]
```
This way you are asking to target only Chrome 62+ which largely supports ES6. I think with `env` only it will transpiler in such a way to support browsers like IE11. More [here](https://www.npmjs.com/package/babel-preset-env).

@JeroenVinke thanks for your feedback too. For the CLI, adding the uglify-es dependency seems to work well for me, therefore also removing the try/catch.

huochunpeng:
@rodu, thx, I tested with `"targets": { "browsers": ["Firefox ESR"] }` before. All good.

You can further

* remove `isV3` check and all code branch for dealing `uglify-js` v2.
* move `require('uglify-es')` to top of the file.

JeroenVinke:
Thanks everyone for helping out. Do you think this is ready to go?

huochunpeng:
I installed this branch as my global cli, created new app, modified package.json to use this branch. All good for me.

huochunpeng:
Wait for #884 to be merged first.

rodu:
OK, thanks for keeping an eye on this. I am not sure if the situation is already stable enough to use the new fork. In case let me know at some point. I guess there is still to wait for #884 anyway from your previous comment.

huochunpeng:
@JeroenVinke now we have another reason to switch from uglify-js to uglify-es or probably terser.

Even with babel targeting es2015, the currently uglify-js still could fail on vendor js file (namely CKEditor) who ship js code with using "const".

```
SyntaxError: Unexpected token: keyword (const)
```

JeroenVinke:
https://github.com/aurelia/cli/pull/907 has been merged

PR #884, opened 2018-05-18
fix(bundler): fix compatibility with source-map >= v0.6

### LINKED ISSUE #870: Not compatible with source-map v0.6.1

* **Library Version:**
0.33.1

* **Operating System:**
All

* **Node Version:**
8.10.0

* **NPM Version:**
5.8.0
yarn is not affected

* **Browser:**
NA

* **Language:**
NA

* **Loader/bundler:**
NA

**Current behavior:**

Depends on user app's extra dependencies (like postcss), sometimes npm resolved source-map package to v0.6.1 instead of v0.5.7.

Yarn somehow always gives me v0.5.7.

When v0.6.1 is in used, following error is thrown
```
Error: original.line and original.column are not numbers -- you probably meant to omit the original mapping entirely and only map the generated position. If so, pass null for the original mapping instead of an object with empty or null values.
at SourceMapGenerator_validateMapping [as _validateMapping] (/Users/huocp/bw/bcx-ui/node_modules/source-map/lib/source-map-generator.js:276:15)
```
You can duplicate the error by forcing `"source-map": "^0.6.1"` in local app's package.json.

This can be temporarily fixed by adding `"source-map": "^0.5.7"` to aurelia-cli's package.json dependencies.
But long term fix should be updating cli's code to be compatible with latest source-map version.

huochunpeng:
Upgrade embedded concat-with-source-maps to latest version.

Close #870

### DIFF
--- a/lib/build/concat-with-sourcemaps/index.js
+++ b/lib/build/concat-with-sourcemaps/index.js
@@ -13,9 +13,9 @@ function Concat(generateSourceMap, fileName, separator) {
   this.contentParts = [];
 
   if (separator === undefined) {
-    this.separator = new Buffer(0);
+    this.separator = bufferFrom('');
   } else {
-    this.separator = new Buffer(separator);
+    this.separator = bufferFrom(separator);
   }
 
   if (this.sourceMapping) {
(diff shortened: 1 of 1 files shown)


huochunpeng:
@JeroenVinke we have embedded 3 tools

lib/build/concat-with-sourcemaps
lib/build/convert-source-map
lib/build/map-stream

It looks like it's better to not embedding them, but to use latest npm package version.

Issue #868, opened 2018-04-10
Error: Path variable [contenthash] not implemented in this context: [contenthash].css


constructor:
**I'm submitting a bug report**

* **Library Version:**
0.33.1

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
8.10.0

* **NPM Version:**
5.8.0

* **Browser:**
all 

* **Language:**
TypeScript X.X

* **Loader/bundler:**
Webpack

**Current behavior:**
1) Using Windows 10, I create a new folder and open Powershell.

2) `au new --here` with options:
	3: Webpack
	2: ASP.NET Core
	2: Typescript
	1: Default - No markeup processing
	1: Default - Use standard css
	1: None - Skip testing
	2: No - Skip testing
	1: Visual Studio Code
	1: Yes (Default) - Creates the project structure…
	1: Yes (Default) - Installs all server, client and tooling dependencies…

3) Create app.css with content:
`body { background-color: red; }`
4) Include css in app.ts:
`import './app.css';`
5) `au build --env prod` gives:
Error: Path variable [contenthash] not implemented in this context: [contenthash].css

It seems to be related to:
https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/763

I have followed one suggestion from the above link and added mini-css-extract-plugin for latest webpack. This and a few other small changes make my experience with aurelia-cli now much better. Find 
 my changes here: https://github.com/constructor/aurelia-cli-fixes

* **What is the expected behavior?**
To build without errors using latest libs. As a new user this was causing lots of unnecessary pain. It should just work.

* **What is the motivation / use case for changing the behavior?**
So it just works.

Alexander-Taran:
Would you like to provide a pull request for it @constructor ?

constructor:
@Alexander-Taran yes, I can do that. I have just one question. The changes I made consist of 3 commits that address 3 issues. The main one being the '[contenthash] not implemented' error. This issue was a single commit (af08725fca36d2ab94f408bf9870a8f2e264915e). The other 2 smaller issues were a 404 css error and the addition of a launchSetting for convenient switching of ASPNETCORE_ENVIRONMENT to 'Production' in Visual Studio. One file from the launchSetting change was committed with 404 fixes. Should I resolve this or would they all be taken together?

Alexander-Taran:
let's ask @JeroenVinke  (-:

but there is a chance that plugins will get fixed and this won't be needed right?

[...]

PR #886, opened 2018-05-20
fix(dependencies): lock webpack version to 4.1.1

doktordirk:
Temporarily locked to avoid breaking changed in about 4.4.0 which saw a switch from extract-text-webpack-plugin to mini-css-extract-plugin, preventing production build in the current configuration

see #868

### DIFF
--- a/lib/dependencies.json
+++ b/lib/dependencies.json
@@ -102,7 +102,7 @@
   "url-loader": "^1.0.1",
   "vinyl-fs": "^3.0.2",
   "wait-on": "^2.1.0",
-  "webpack": "^4.1.1",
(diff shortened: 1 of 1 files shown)


Alexander-Taran:
@JeroenVinke 
there is also lazy option not woriking since 4.2 and paths with spaces are broken since 4.5 both issues are with webpack-dev-server

JeroenVinke:
Sorry for the late response. Based on https://github.com/aurelia/cli/issues/900 I've made the change to a different plugin: https://github.com/aurelia/cli/commit/e9641e4c86fe0e08ecd18dbe96ace8d69fe05a7b

PR #897, opened 2018-06-26
fix(jest): update usage of jest.runCli

### LINKED ISSUE #896: jest.runCli returns a promise now, the callback in tasks/jest.js never gets executed

* **Library Version:**
0.33.1

* **Operating System:**
OSX 10.13.5

* **Node Version:**
9.11.1

* **NPM Version:**
5.8.0

* **Browser:**
N/A

* **Language:**
all

* **Loader/bundler:**
all

**Current behavior:**

I wanted to run the jest tests automatically when running `au build --env dev` so I added the following to tasks/build.js:

```
import jest from './jest';

...

function test(done) {
  if (!production) {
    jest(done);
  } else {
    done();
  }
}

const build = gulp.series(
  clearDist,
  test,
  configureEnvironment,
  buildWebpack
);
```

And the build exits every time after the test phase.

The build should continue to `configureEnvironment` and `buildWebpack` after `test`.

As discovered while reading this post: https://github.com/facebook/jest/issues/5048, "commit [200b032](https://github.com/facebook/jest/commit/200b032d170054dc3860f3cd869a94201c54dd19) changed the `runCli()` - it removed the `onComplete` callback and replaced it with a Promise".

Changing 
  ```
jest.runCLI(options, [path.resolve(__dirname, '../../')], (result) => {
    if (result.numFailedTests || result.numFailedTestSuites) {
      cb(new gutil.PluginError('gulp-jest', { message: 'Tests Failed' }));
    } else {
      cb();
    }
  });
```
to
```
 
  jest.runCLI(options, [path.resolve(__dirname, '../../')]).then((result) => {
    if (result.numFailedTests || result.numFailedTestSuites) {
      cb(new gutil.PluginError('gulp-jest', { message: 'Tests Failed' }));
    } else {
      cb();
    }
  });
```

fixes this issue.

1hella:
jest.runCli returns a promise. This change updates the usage of jest.runCli to use the promise rather than pass in a callback which never will get executed.

closes #896

### DIFF
--- a/lib/resources/tasks/jest.js
+++ b/lib/resources/tasks/jest.js
@@ -13,7 +13,7 @@ export default (cb) => {
 
   process.env.BABEL_TARGET = 'node';
 
-  jest.runCLI(options, [path.resolve(__dirname, '../../')], (result) => {
+  jest.runCLI(options, [path.resolve(__dirname, '../../')]).then((result) => {
     if (result.numFailedTests || result.numFailedTestSuites) {
       cb(new gutil.PluginError('gulp-jest', { message: 'Tests Failed' }));
     } else {
--- a/lib/resources/tasks/jest.ts
+++ b/lib/resources/tasks/jest.ts
@@ -12,7 +12,7 @@ export default (cb) => {
     Object.assign(options, { watch: true});
   }
 
-  jest.runCLI(options, [path.resolve(__dirname, '../../')], (result) => {
+  jest.runCLI(options, [path.resolve(__dirname, '../../')]).then((result) => {
     if(result.numFailedTests || result.numFailedTestSuites) {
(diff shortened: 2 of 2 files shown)


Alexander-Taran:
@1hella which version of jest breaks behavior? 
Was it a major bump? Because this kind of change will be backwards incompatible. So the jest dependency should be changed as well.

1hella:
@Alexander-Taran I'll double check that

1hella:
@Alexander-Taran It looks like this change was added in jest-cli version 21.0.0, and the version in the package.json from `au new` looks to be ^23.0.0-alpha.0, so should be fine.

JeroenVinke:
Thank you!

PR #904, opened 2018-07-24
fix(test): fix broken karma setup

huochunpeng:
Currently karma setup is all broken for requirejs/systemjs app.


huochunpeng:
test has been fixed in current head.

PR #906, opened 2018-07-26
fix: enforce single aurelia-binding in webpack bundle

huochunpeng:
aurelia/binding#702 this is to avoid v1/v2 duplication due to out-of-date dependencies on 3rd party aurelia plugins.


EisenbergEffect:
@JeroenVinke This is another good PR to accompany the other I mentioned. This should help webpack users avoid multiple binding library bundles.

PR #908, opened 2018-07-29
feat: upgrade to babel-preset-env

### LINKED ISSUE #754: Deprecated use of babel-preset-es2015

* **Library Version:**
0.31.3

**Current behavior:**
When creating a Babel project, `babel-preset-es2015` is added as a dependency

Should use `babel-preset-env` instead

Warning from `babel-preset-es2015`:

> babel-preset-es2015@6.24.1: We're super 😸  excited that you're trying to use ES2015 syntax, but instead of making more yearly presets 😭 , Babel now has a better preset that we recommend you use instead: npm install babel-preset-env --save-dev. preset-env without options will compile ES2015+ down to ES5 just like using all the presets together and thus is more future proof. It also allows you to target specific browsers so that Babel can do less work and you can ship native ES2015+ to user 😎 ! We are also in the process of releasing v7, so please give http://babeljs.io/blog/2017/09/12/planning-for-7.0 a read and help test it out in beta! Thanks so much for using Babel 🙏, please give us a follow on Twitter @babeljs for news on Babel, join slack.babeljs.io for discussion/development and help support the project at opencollective.com/babel

huochunpeng:
Unify babelrc for webpack and cli bundler.
Cleaned up webpack babelrc to use same plugins as other setup.
Add IE10/11 support back to webpack setup as that is what Aurelia claimed to support.

closes #754, #909


lib/resources/src/main-webpack.js
```diff
@@ -1,6 +1,5 @@
 import environment from './environment';
 import {PLATFORM} from 'aurelia-pal';
-import 'babel-polyfill';
```
huochunpeng: aurelia-bootstrapper does load aurelia-polyfills. This line is unnecessary.

lib/resources/content/babelrc.js
```diff
@@ -2,26 +2,23 @@
...
-            "browsers": [
-              "last 2 versions",
-              "not ie <= 11"
-            ],
-            "uglify": process.env.NODE_ENV === 'production',
+            "browsers": [ "last 2 versions" ]
           },
           "loose": true,
-          "modules": process.env.BABEL_TARGET === 'node' ? 'commonjs' : false,
-          "useBuiltIns": true
```
huochunpeng: See #909

lib/resources/content/babelrc.js
```diff
@@ -2,26 +2,23 @@
...
+          "modules": process.env.BABEL_TARGET === 'node' ? 'commonjs' : false
         }
-      ]
+      ],
+      "stage-1"
```
huochunpeng: For webpack, use "stage-1" to replace handpicked "transform-class-properties".
Added "syntax-flow" for webpack setup too.

lib/resources/content/babelrc.js
```diff
@@ -2,26 +2,23 @@
...
       [
         "env", {
           "targets": process.env.BABEL_TARGET === 'node' ? {
             "node": process.env.IN_PROTRACTOR ? '6' : 'current'
           } : {
-            "browsers": [
-              "last 2 versions",
-              "not ie <= 11"
-            ],
-            "uglify": process.env.NODE_ENV === 'production',
```
huochunpeng: As we are using terser now, we don't need to ask preset-env to generate compatible code for uglify-js v3.

lib/resources/content/babelrc.js
```diff
@@ -2,26 +2,23 @@
...
     ],
     "presets": [
       [
         "env", {
           "targets": process.env.BABEL_TARGET === 'node' ? {
             "node": process.env.IN_PROTRACTOR ? '6' : 'current'
           } : {
-            "browsers": [
-              "last 2 versions",
-              "not ie <= 11"
```
huochunpeng: Aurelia claimed supporting IE10.

Issue #888, opened 2018-06-01
Out-of-the-box high vulnerabilities for Aurelia-CLI generated projects


zewa666:
**I'm submitting a bug report**

* **Library Version:**
latest

**Please tell us about your environment:**
* **Operating System:**
all

**Current behavior:**
The out-of-the box setup has a dependency on npm which is introducing vulnerabilities with `https-proxy-agent`. Run `npm audit` to see the vulnerability marked as `high`

* **What is the expected behavior?**
Do we even need npm as a dependency? Chances are extremely high the user has node installed anyways, so dropping that and defining a min version via the engines section in the package.json file.

* **What is the motivation / use case for changing the behavior?**
Get rid of high vulnerabilities

JeroenVinke:
Yeah we should just run `npm install` from the project directory

zewa666:
@JeroenVinke so something like the following should be enough?

```typescript
const npm = spawn('npm', ['install'], { cwd: projectPath });
npm.on("exit", () => {
  // continue with process
});

npm.on("error", (e) => {
  // handle error
});
```

JeroenVinke:
@zewa666 exactly

JeroenVinke:
@huochunpeng can we really close this, or are there other vulnerabilities left?

huochunpeng:
There are two low vulnerabilities left due to outdated deep dep on loash from browser-sync. Nothing I can do about it.

zewa666:
I think this is good enough. The mission with this story was to eliminate the critical ones.

PR #915, opened 2018-08-22
chore: remove dependency on "npm" pacakge

### LINKED ISSUE #914: Question about dependency on npm
The npm package itself is pretty cumbersome. Is there any particular reason why we have to depend on it?

huochunpeng:
closes #914, #888


JeroenVinke:
Awesome!

PR #919, opened 2018-09-12
fix: fix regression on npm/yarn path when it has white space

huochunpeng:
@JeroenVinke quotes are back. Sorry about that.


JeroenVinke:
No worries 😄 you’re doing a great job

PR #935, opened 2018-10-11
fix(bundler): Revisions are inserted into platform.index for all bundles

### LINKED ISSUE #933: Revisions are not inserted into index.html properly

* **Library Version:**
1.0.0-beta.1

* **Operating System:**
Ubuntu 18.10

* **Node Version:**
10.7.0

* **NPM Version:**
6.1.0

* **Browser:**
all

* **Language:**
ESNext

* **Loader/bundler:**
RequireJS

**Current behavior:**
When enabling the revision option in the bundler, it is generating bundles with revision but not updating references in index.html except for vendor-bundle.
```js
"options": {
  "minify": "prod",
  "sourcemaps": "prod",
  "rev": "prod"
}
```
Result
```html
<script src="dist/vendor-bundle-85d36844fc.js" data-main="aurelia-bootstrapper"></script>
<script src="dist/plugin-bundle.js"></script>
<script src="dist/3rd-party.js"></script>
<script src="dist/app-bundle.js"></script>
```

By enabling the revisioning I would expect all bundles to be updated properly in the index.html

When running in a docker build `au test` and right after `au build --env prod` the live version will use the bundles built for the test runner. In my case it contains code coverage code of istanbul babel plugin^^

fragsalat:
This fixes #933 by first writing all bundles to the disk and after wards update the file name in the platform.index for each bundle. This allows users to load all bundles in parallel via script tags.

PS: There were 18 tests failing even before I started^^


fragsalat:
@huochunpeng @JeroenVinke ping

huochunpeng:
@fragsalat please check the code comments.

fragsalat:
I'm actually not seeing any. Did you may forgot to submit the review?

huochunpeng:
My bad

huochunpeng:
The reason your PR didn't trigger circleci check is that your commit is on very old master (before circleci was turned on), please rebase to latest master.

lib/build/bundler.js
```diff
@@ -232,7 +232,12 @@ exports.Bundler = class {
   }
 
   write() {
-    return Promise.all(this.bundles.map(x => x.write(this.project.build.targets[0])));
+    return Promise.all(this.bundles.map(bundle => bundle.write(this.project.build.targets[0])))
```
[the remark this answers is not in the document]
fragsalat: It was actually like this except that the write was asynchronous. Once the vendor bundle was reached it wrote the revisions to the html but the vendor bundle doesn't know about the revisions of other files. I don't wanted to introduce code to pass revisions of other bundles into the last bundle because it felt a bit dirty and therefore I delayed the write process. 

Regarding the sync write: I started with that solution but it were a bit hard because you use a mkdir function in the async write which is based on callbacks. I don't wanted to create a sync write function without this mkdir stuff as it would mean different behavior for those functions.

fragsalat:
I pulled aurelia/cli@master into my branch but circle ci still don't trigger

lib/build/bundle.js
```diff
@@ -437,26 +431,24 @@ exports.Bundle = class {
     return 'function _aureliaConfigureModuleLoader(){' + config + '}';
   }
 
-  setIndexFileConfigTarget(platform, location) {
-    //Replace the reference to the vendor bundle in the "index.html" file to use the correct build revision file (or remove the build revision hash);
-    const createSrcFileRegex = require('./utils').createSrcFileRegex;
-    let indexLocation = platform.index;
+  async writeRevToIndex(platform) {
```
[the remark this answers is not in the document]
fragsalat: Hmmm mdn tells me that the support in nodejs starts of 7.6.0
https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Operators/await

lib/build/bundler.js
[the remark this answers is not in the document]
huochunpeng: I see.

I have some concern about current bundle writes.
All bundles writes are in parallel (`Promise.all`), they are fine because they deal with different files.
But now they all wants to touch same file, if:
```
bundle A reads index file
bundle B reads index file
bundle A writes modified index file (if A writes after B reads, this modification will be lost)
bundle B writes modified index file
```

I think we need to ensure the operation on index file is in serial (chain of promises), not parallel.
Serialise all bundle writes may be an overkill. But serialise index file modifications is needed, maybe extract it out and do it after all bundles writes (I mean the bundle files themselves) finished?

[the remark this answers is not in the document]
huochunpeng: For efficiency, maybe move the logic to bundler? So you can do just one write for all bundles, instead of rewrite index file multiple times.

lib/build/bundler.js
```diff
@@ -233,7 +233,12 @@ exports.Bundler = class {
...
+      .then(async() => {
+        for (let i = this.bundles.length; i--; ) {
+          await this.bundles[i].writeRevToIndex(this.project.build.targets[0]);
```
huochunpeng: Just realised you did serial write on index file. 👍 
I am not familiar with async/await, took me quite a while to understand what's doing here.

huochunpeng:
👍 looks good, and worked for me.
I didn't use async/await, the serialised `await` spun my head, but it's a cool usage.

lib/build/bundle.js
```diff
@@ -438,26 +432,24 @@ exports.Bundle = class {
...
-      } else {
-        logger.error('Error: Unable to update ' + this.moduleId + ' path to ' + location + ', could not find existing reference to replace');
-      }
-    } catch (err) {
-      if (err.code === 'ENOENT') {
-        logger.error('Error: No index file found at "' + indexLocation + '"');
-      } else {
-        logger.error(err);
+      // Validate rev is enabled and a hash was generated
+      if (platform.index && this.hash) {
```
huochunpeng: Found a bug, you need to write empty hash to index file.
Consider:
```sh
au run --env prod
# terminate it, then
au run
```

The second `au run` need to overwrite index file to empty hash.

lib/build/bundle.js
```diff
@@ -438,26 +432,25 @@ exports.Bundle = class {
...
+      let indexFile = fs.readFileSync(platform.index);
+      const outputDir = platform.baseUrl || platform.output;
+      const configMatcher = Utils.createSrcFileRegex(outputDir, this.moduleId);
+      const bundleFileName = this.hash ? Utils.generateHashedPath(this.config.name, this.hash) : this.config.name;
+      const bundleLocation = path.posix.join(outputDir, bundleFileName);
+      // Replace file name with hashed file name
+      if (configMatcher.test(indexFile)) {
+        indexFile = indexFile.replace(configMatcher, 'src="$2' + bundleLocation + '"');
+        await fs.writeFile(platform.index, indexFile);
+        logger.info(`INFO [Bundle] Updated file name for ${this.moduleId} in ${platform.index}`);
```
huochunpeng: Please remove "INFO [Bundle] " prefix, `logger.info` does that for you.

lib/build/bundle.js
```diff
@@ -438,26 +432,25 @@ exports.Bundle = class {
...
       }
+    } catch (error) {
+      logger.error(`ERROR [Bundle] Couldn't update file name with revision in ${platform.index}`, error);
```
huochunpeng: Please remove "INFO [Bundle] " prefix, `logger.error` does that for you.

lib/build/bundle.js
huochunpeng: You got duplicated prefix.
```
INFO [Bundle] Writing app-bundle.js...
INFO [Bundle] Writing vendor-bundle.js...
INFO [Bundle] INFO [Bundle] Updated file name for vendor-bundle in index.html
INFO [Bundle] INFO [Bundle] Updated file name for app-bundle in index.html
Finished 'writeBundles'
```

fragsalat:
Fixed the logger stuff :)

huochunpeng:
@JeroenVinke could you review this? Looks good to me, not critical feature but does not hurt.

JeroenVinke:
Hey @fragsalat and @huochunpeng  :)

I've had a look, and had to think about the for loop for a while too 😄 

Only thing I could find to complain about is the name of the function writeRevToIndex. I believe this function also updates the bundle names in index.html when rev isn't enabled. If you agree, would you mind changing the name? Then we can merge this :)

Great work @fragsalat

Alexander-Taran:
@JeroenVinke naming is fixed (-:

fragsalat:
Still not merged...?

JeroenVinke:
Sorry about that, thanks a lot for your efforts @fragsalat

PR #939, opened 2018-10-13
chore: readable error for unsupported nodejs version

huochunpeng:


bin/aurelia-cli.js
```diff
@@ -2,6 +2,15 @@
 'use strict';
 const resolve = require('../lib/resolve');
 
+const nodeVersion = process.versions.node;
+const [major, minor] = nodeVersion.split('.');
+if (major < 8 || (major === '8' & minor < 9)) {
```
bigopon: Doesn't look like major and minor can ever be number, no?

huochunpeng: As long as nodejs follows semver, it is safe.

huochunpeng: sh*t, typo '&'

huochunpeng: surprised the typo didn't fail my tests on various version.

bigopon:
```js
[major, minor] = '8.9'.split('.')
(2) ["8", "9"]
major < 8
false
```

bigopon: Ok, I see. I forgot it's valid JS coercion 👍

huochunpeng: `"8" < 8` is false, what's not right?

bin/aurelia-cli.js
```diff
@@ -2,6 +2,15 @@
 'use strict';
 const resolve = require('../lib/resolve');
 
+const nodeVersion = process.versions.node;
+const [major, minor] = nodeVersion.split('.');
+if (major < 8 || (major === '8' && minor < 9)) {
```
zewa666: var semver = require("semver")

if (semver.lt(process.versions.node, "8.9.0")) {
...

zewa666: Why do by ourselves if semver got us here :)

huochunpeng: Thx!
Question: this is a bin file, can it access aurelia-cli dependencies? I will try it.

huochunpeng: Great! It works.

PR #951, opened 2018-10-16
test(bundler): cleanup test code on windows

huochunpeng:
Also fix a missed forceCjsWrap on windows. Normalized package description path format.

### DIFF
--- a/lib/build/bundled-source.js
+++ b/lib/build/bundled-source.js
@@ -214,7 +214,7 @@ exports.BundledSource = class {
         let contents;
         // forceCjsWrap bypasses a r.js parse bug.
         // See lib/amodro-trace/read/cjs.js for more info.
-        let forceCjsWrap = !!modulePath.match(/\/(cjs|commonjs)\//i);
+        let forceCjsWrap = !!modulePath.match(/(\/|\\)(cjs|commonjs)(\/|\\)/i);
 
         try {
           contents = cjsTransform(modulePath, this.contents, forceCjsWrap);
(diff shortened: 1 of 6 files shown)


lib/build/bundled-source.js
```diff
@@ -214,7 +214,7 @@ exports.BundledSource = class {
         let contents;
         // forceCjsWrap bypasses a r.js parse bug.
         // See lib/amodro-trace/read/cjs.js for more info.
-        let forceCjsWrap = !!modulePath.match(/\/(cjs|commonjs)\//i);
+        let forceCjsWrap = !!modulePath.match(/(\/|\\)(cjs|commonjs)(\/|\\)/i);
```
zewa666: I'm not sure what's the best practice here. In theory this would accept a mix of both forward and backward slashes, which might not even be a problem.

Alternatively we could use the OS native slash style.

```javascript
const sep = require("path").sep === "\\"
  ? "\\\\"
  : "/"
const target = "/cjs/";

const res = target.match(new RegExp(`(${sep})(cjs|commonjs)(${sep})`, "i"))

console.log(res);
```

huochunpeng: I thought about that when I was doing the fix. The lazy side of me prevailed:-)

zewa666: Hehe, yeah I think we're fine with your approach

huochunpeng:
FYI, we can switch from circleci to travis-ci for cross-OS testing which is much needed for aurelia-cli.

https://travis-ci.org/huochunpeng/cli/builds/442139818

zewa666:
Haha yeah we actually were on travis previously and moved to circle because its generally much nicer, stable and faster. Guess we'll have to revisit this at least for the CLI. Maybe lets discuss this internally @EisenbergEffect ?

Alexander-Taran:
why not have both?

EisenbergEffect:
@fkleuver Is our resident CircleCI expert. I'll defer to him for strategy on this. He's done amazing work for the vNext repo.

Issue #958, opened 2018-10-25
Duplication in bundle output using CLI generated webpack config (duplicated modules) 


chrisckc:
With the default CLI generated webpack config, adding jQuery and bootstrap causes issues.

When either jQuery, Bluebird or the jQuery bootstrap extensions are used in code, running webpack results in duplication across both the app and vendor bundles. This causes all kinds of issues when using jQuery.

The CLI generates a package.json that references webpack v4, this also occurs in webpack v3 as used in the skeleton webpack repo, refer to the issue i logged here for more details:

https://github.com/aurelia/skeleton-navigation/issues/894

In the skeleton repo which already includes jquery and bootstrap, the issue is caused by the line `vendor: ['bluebird', 'jquery', 'bootstrap'],` in the webpack.config.js under the `entry` property. 

If jquery and bootstrap are added to the vendor list in the CLI generated config, as initial logic would suggest, referencing jquery in code causes duplication of jquery across the app and vendor bundles.

webpack v4 uses the split-chunks-plugin, the `vendor` under `entry` should not be used:

https://webpack.js.org/plugins/split-chunks-plugin/

https://stackoverflow.com/questions/49017682/webpack-4-migration-commonschunkplugin

Note that commenting out the vendor line results in no vendor bundle being generated and everything being in the app bundle.

To create the separate vendor bundle add the following to webpack.config.js

```
optimization: {
    splitChunks: {
      chunks: "initial",
      cacheGroups: {
        vendors: {
          test: /[\\/]node_modules[\\/]/,
          name: "vendor",
          enforce: true,
          priority: 10
        }
      }
    }
  }
```

The above config is just a simple example, the vendor bundle is pretty big and additional config should be added to split it up further, for example separate out the aurelia modules from the rest of the node modules.

The CLI needs to be updated to output a better webpack config to avoid potential developer frustration and the resulting deep dive into webpack that will be required to figure out what's going on.

chrisckc:
This will also be the cause of this issue:
https://github.com/aurelia/cli/issues/918

chrisckc:
Here is a CLI generated project with source files from the skeleton added and various improvements including this fix:
https://github.com/chrisckc/aurelia-cli-skeleton-typescript-webpack

Consider incorporating some of my changes to the webpack config into the CLI code generator to improve the starting experience for newcomers to either Webpack, Aurelia or both.

[...]

PR #965, opened 2018-11-07
fix(webpack.config.template.js): change webpack config to fix bundle duplication issue

chrisckc:
Updating the webpack config to use optimization.splitChunks rather than entry.vendor resolves an issue that appears when jQuery or Bootstrap is added.
When attempting to use jQuery or the jQuery Bootstrap extensions in code, webpack is duplicating jQuery across the app end vendor bundles.
Using optimization.splitChunks allows webpack to workout the dependencies and also reduces the size of the generated bundles by splitting them up.

Fixes this issue: https://github.com/aurelia/cli/issues/958

### DIFF
--- a/lib/resources/content/webpack.config.template.js
+++ b/lib/resources/content/webpack.config.template.js
@@ -44,8 +44,7 @@ module.exports = ({production, server, extractCss, coverage, analyze, karma} = {
     alias: { 'aurelia-binding': path.resolve(__dirname, 'node_modules/aurelia-binding') }
   },
   entry: {
-    app: ['aurelia-bootstrapper'],
-    vendor: ['bluebird'],
+    app: ['aurelia-bootstrapper']
   },
   mode: production ? 'production' : 'development',
   output: {
@@ -54,6 +53,58 @@ module.exports = ({production, server, extractCss, coverage, analyze, karma} = {
     filename: production ? '[name].[chunkhash].bundle.js' : '[name].[hash].bundle.js',
     sourceMapFilename: production ? '[name].[chunkhash].bundle.map' : '[name].[hash].bundle.map',
     chunkFilename: production ? '[name].[chunkhash].chunk.js' : '[name].[hash].chunk.js'
+  },
+  optimization: {
+    // Use splitChunks to breakdown the vendor bundle into smaller files
+    // https://webpack.js.org/plugins/split-chunks-plugin/
+    splitChunks: { 
+      chunks: "initial",
+      cacheGroups: {
+        default: false, // Disable the built-in groups (default and vendors)
+        vendors: false,
+        bluebird: {
+          test: /[\\/]node_modules[\\/]bluebird[\\/]/,
+          name: "vendor.bluebird",
+          enforce: true,
+          priority: 100
+        },
+        // You can insert additional entries here for jQuery and bootstrap etc. if you need them
+        // Break the Aurelia bundle down into smaller chunks, binding and templating are the largest
+        aureliaBinding: {
+          test: /[\\/]node_modules[\\/]aurelia-binding[\\/]/,
+          name: "vendor.aurelia-binding",
+          enforce: true,
+          priority: 28
+        },
+        aureliaTemplating: {
+          test: /[\\/]node_modules[\\/]aurelia-templating[\\/]/,
+          name: "vendor.aurelia-templating",
+          enforce: true,
+          priority: 26
+        },
+        aurelia: {
+          test: /[\\/]node_modules[\\/]aurelia-.*[\\/]/,
+          name: "vendor.aurelia",
+          enforce: true,
+          priority: 20
+        },
+        // This picks up everything else being used from node_modules
+        vendors: {
+          test: /[\\/]node_modules[\\/]/,
+          name: "vendor",
+          enforce: true,
+          priority: 10
+        },
+        common: { // common chunk
+          name: 'common',
+          minChunks: 2,   // Creates a new chunk if a module is shared between different chunks more than twice
+          chunks: 'async',
+          priority: 0,
+          reuseExistingChunk: true,
+          enforce: true
+        }
+      }
+    }
   },
   performance: { hints: false },
   devServer: {


chrisckc:
@EisenbergEffect I have updated the webpack.config.template.js so that it creates bundles that fall below the webpack bundle size warnings that appear in the webpack output during build.

I tested the updated cli locally and these are the results:

Production build:
```
Version: webpack 4.25.1
Time: 7999ms
Built at: 07/11/2018 15:04:13
                                                    Asset       Size  Chunks             Chunk Names
                       app.e33b1ad83d52dc9df9fa.bundle.js   2.24 KiB       0  [emitted]  app
                      app.e33b1ad83d52dc9df9fa.bundle.map   3.81 KiB       0  [emitted]  app
                                              favicon.ico   14.7 KiB          [emitted]  
                                               index.html  773 bytes          [emitted]  
   vendor.aurelia-binding.d1cee084acf628e4b576.bundle.map   93.4 KiB       3  [emitted]  vendor.aurelia-binding
     vendor.aurelia-binding.d1cee084acf628e4b576.chunk.js   89.1 KiB       3  [emitted]  vendor.aurelia-binding
vendor.aurelia-templating.c981a99cf0c3cf7a0c3c.bundle.map   89.8 KiB       4  [emitted]  vendor.aurelia-templating
  vendor.aurelia-templating.c981a99cf0c3cf7a0c3c.chunk.js   69.7 KiB       4  [emitted]  vendor.aurelia-templating
           vendor.aurelia.92158d83fc4085124928.bundle.map    209 KiB       2  [emitted]  vendor.aurelia
             vendor.aurelia.92158d83fc4085124928.chunk.js    160 KiB       2  [emitted]  vendor.aurelia
          vendor.bluebird.40f3c4893b49b2568b86.bundle.map     94 KiB       5  [emitted]  vendor.bluebird
            vendor.bluebird.40f3c4893b49b2568b86.chunk.js     76 KiB       5  [emitted]  vendor.bluebird
                   vendor.e63579ed699af03388a3.bundle.map      7 KiB       1  [emitted]  vendor
                     vendor.e63579ed699af03388a3.chunk.js   4.96 KiB       1  [emitted]  vendor
```
As can be seen, production bundle sizes are nice and small

Development build:

```
Version: webpack 4.25.1
Time: 3189ms
Built at: 07/11/2018 15:02:39
                                                  Asset       Size                     Chunks                    Chunk Names
                     app.e67ef7f317cff8e07eec.bundle.js   15.3 KiB                        app  [emitted]         app
                                            favicon.ico   14.7 KiB                             [emitted]         
                                             index.html  976 bytes                             [emitted]         
   vendor.aurelia-binding.e67ef7f317cff8e07eec.chunk.js    464 KiB     vendor.aurelia-binding  [emitted]  [big]  vendor.aurelia-binding
vendor.aurelia-templating.e67ef7f317cff8e07eec.chunk.js    429 KiB  vendor.aurelia-templating  [emitted]  [big]  vendor.aurelia-templating
           vendor.aurelia.e67ef7f317cff8e07eec.chunk.js   1.06 MiB             vendor.aurelia  [emitted]  [big]  vendor.aurelia
          vendor.bluebird.e67ef7f317cff8e07eec.chunk.js    448 KiB            vendor.bluebird  [emitted]  [big]  vendor.bluebird
                   vendor.e67ef7f317cff8e07eec.chunk.js   42.7 KiB                     vendor  [emitted]         vendor      
```

As can be seen, the dev build bundle sizes still report [big] however the largest file size would be much bigger if the aurelia bundle was not broken down further.

If you would rather not have the aurelia bundle split up i could modify it with another commit?

EisenbergEffect:
@jods4 Can you review the PR. Any thoughts on this?

lib/resources/content/webpack.config.template.js
```diff
@@ -55,6 +54,58 @@ module.exports = ({production, server, extractCss, coverage, analyze, karma} = {
...
+        vendors: false,
+        bluebird: {
+          test: /[\\/]node_modules[\\/]bluebird[\\/]/,
+          name: "vendor.bluebird",
+          enforce: true,
+          priority: 100
+        },
+        // You can insert additional entries here for jQuery and bootstrap etc. if you need them
+        // Break the Aurelia bundle down into smaller chunks, binding and templating are the largest
+        aureliaBinding: {
```
bigopon: Could you elaborate why the separate of aurelia-binding and aurelia-templating into their own bundle? @chrisckc

chrisckc: To provide a better chance of the app downloading successfully on an unreliable 3g connection the file sizes should be kept small.  Templating and binding are the biggest modules so splitting them out reduces the aurelia bundle size.

By splitting out vendor bundles you can take more advantage of the browser caching and webpack's chunk hashing to prevent re-downloading of unchanged modules when the app is updated and published to production.

The issue of making more http requests is outweighed by the advantages in my opinion, especially when using HTTP2.

While on this topic, it would be great if the webpack skeleton samples contained more examples of lazy loading.

bigopon: I see. This is nice & thanks for the explanation @chrisckc . 

Changes lgtm @EisenbergEffect . No one has complained about bulky single bundle so far, so I don't it's not well known optimization, or unnecessary or everyone has their own requirement of optimization. Regardless, I think this is good to promote as default behavior.

EisenbergEffect:
thanks @chrisckc !

jods4:
Sorry I'm late to the party.

First: I don't think there is a one-size-fits-all config. Different use cases might call for different configs.

Most users on desktop are fine with one big package (maybe 2, typically vendors, because the latter is more stable -- cacheable -- and can be downloaded in parallel).

On the other hand, if you are on a really unstable network, I can see why downloading smaller files is better.

That caveat aside, I would make the following comments on this change:

**1. Long term caching**
> By splitting out vendor bundles you can take more advantage of the browser caching and webpack's chunk hashing to prevent re-downloading of unchanged modules when the app is updated and published to production.

Except I don't think it works with this config.
If you want long term caching you should ensure webpack module ids are stable between compilation, which is typically done by using `HashedModuleIdsPlugin`.
Docs: https://webpack.js.org/guides/caching/#module-identifiers

**2. Manual chunking**
There's quite a lot of hard-coded config there. Have you considered letting Webpack do it instead?

The default config doesn't split initial chunk because unless you use `HtmlWebpackPlugin` it doesn't work.

But you can enable it and then Webpack takes care of splitting. You can tweak how many concurrent downloads you want, how big or small each chunk can be, etc.
I can lead to smaller config and more optimized loading.

In you particular use case of a slow network config, you could define a max chunk size so that Webpack creates as many chunks as you need to stay under that value.

**3. That particular split config**
Just saying: that partitioning creates a little more initial chunks than is typically considered good, at least when used in a normal network environment.

I understand that if you have a bad network chunk sizes is more relevant than number of round-trips. 
It's just that if this template is meant for general usage, this is probably not the best "general" setup.

jods4:
@EisenbergEffect No, I am just saying. 😉 

I have no strong opinion on the CLI templates -- as I said I don't believe there's one-size-fits-all.

At least, if the intent is to have long-term cacheable files, you should add `HashedModuleIdsPlugin`.

chrisckc:
@jods4 Thanks for the info on the HashedModuleIdsPlugin, i was able to achieve stable output hashes by just using "runtime: true" inside splitChunks config to split out the runtime.

I have tried adding another import to a source file and using some function inside the imported module to make sure it is not removed during the build and still get stable output with only the app bundle changing.

Maybe this plugin is enabled by default when running webpack with mode: 'production' ? The webpack docs don't seem do indicate this though so further testing will be required.

I largely agree with your comments above, i think we need some testing though to find out what the most suitable config should be.

Since that pull request i have been experimenting with lazy loading and more webpack configs, i found an issue with the lazy loading example used in the webpack-skeleton-navigation source in relation to chunk splitting.

@EisenbergEffect I am going to start another issue with some test results to so this can be discussed further as the original intention of this pull request was to fix this issue: https://github.com/aurelia/cli/issues/958 by providing an improved webpack config which works with webpack v4.

jods4:
> I have tried adding another import to a source file and using some function inside the imported module to make sure it is not removed during the build and still get stable output with only the app bundle changing.

I might be wrong, but I don't think this is very conclusive. Module numbering is inherently a little random (depends on ordering). Also when using `AureliaPlugin` many modules have a fixed string for id instead of the sequential numbering webpack normally uses.

> Maybe this plugin is enabled by default when running webpack with mode: 'production' ?

Not according to official docs, see https://webpack.js.org/concepts/mode/#mode-production

If you want better insight into what happens, `--display-modules` shows all modules in your bundles, including their ids.
If your ids are sequential, it's unlikely they're going to be long-term stable.

chrisckc:
After further testing and examination of the webpack output, it seems that all the source code is under a single module id related to the aurelia-bootstrapper, so adding importing new source files make no difference to the id sequence.

If i add and import a new node module, the hash of both the vendor and the app bundles changes, so it looks like the docs are correct and we need to add the HashedModuleIdsPlugin.

chrisckc:
@EisenbergEffect @jods4 Here is the new issue as a feature request for general improvements to the webpack config: https://github.com/aurelia/cli/issues/969 along with some test results.

PR #976, opened 2018-11-15
feat(bundler): support direct css import in js file

huochunpeng:
Support `import './foo.css';` directly in JavaScript. It calls aurelia-pal DOM.injectStyles to inject css content to html head.

cc @zewa666 @cluka


huochunpeng:
Interesting, github only checks travis-ci now, no circle-ci. @EisenbergEffect you didn't purposely turn off circle-ci, right?

EisenbergEffect:
Nope. Let me go see if I can turn that on...and we'll see if it turns of Travis 🤣

EisenbergEffect:
It's enabled and says it ran on master 3h ago. They also seem to be having a service degradation issue right now.

huochunpeng:
It's not circle-ci themself I think.

It's how github picks ci to check PR, it looks like github only picks one ci setup, and travis-ci is on top of that list.

huochunpeng:
Probably need to use https://github.com/aurelia/templating-resources/blob/master/src/css-resource.js to handle css, because of `fixupCSSUrls` for relative resource path (img/font/...).

https://github.com/aurelia/templating-resources/blob/4118ff87839e71b8a2991943964a5bb891576d8e/src/css-resource.js#L10

webpack's style-loader did similar fix.

https://github.com/webpack-contrib/style-loader/blob/80039665d612248535557c41de03a5aa377484a8/lib/urls.js#L57

bigopon:
@huochunpeng those import only comes from js, so I think it should not go through Aurelia. It aligns with what other bundlers are doing as well, I think it's least surprising

huochunpeng:
If you look at webpack style-loader, it does also rewrite resources url.

I don’t need to go through aurelia, I probably can just copy few lines of the fixupCSSUrls over.

bigopon:
Very cool. How will this work in conjunction with sass / less / stylus etc? I think this PR is a very good foundation for folks who prefer to have more power in resolving their assets.

huochunpeng:
They all work. cli bundler doesn't know about sass/less/stylus, gulp-* knows them, so cli doesn't have to.

huochunpeng:
One difference from webpack. webpack style-loader by default rewrite url resource into base64 data-uri in order to bypass the copy of static assets (images/fonts), our implementation does not rewrite url to base64 data-uri.

PR #978, opened 2018-11-16
fix(bundler): use topological sort to ensure shim order

### LINKED ISSUE #955: Error with auto-traced jQuery and Bootstrap v3

* **Library Version:**
1.0.0-beta.3

* **Operating System:**
OSX 10.x

* **Node Version:**
8.12.0

* **NPM Version:**
yarn 1.10.1

* **Browser:**
Chrome 70 | Firefox 63.0 | Safari 12.0

* **Language:**
ESNext

* **Loader/bundler:**
RequireJS

**Current behavior:**
I've just upgraded an old app of mine from Aurelia-cli `v0.33.0` to `1.0.0-beta.3`. Due to the major changes, I tried to use the recipes described in [the cookbook](https://aurelia.io/docs/cli/cli-bundler/cook-book) for both `jQuery` and `Bootstrap v3`.
The recipe for `jQuery` basically says it can be auto-traced (so I removed it from my `aurelia.json`).
The recipe for `bootstrap v3` says it shall be manually configured in `aurelia.json` (so I updated my conf to remove the CSS resource I have previously added, but that was it).
When using both recipes, and since `bootstrap` depends on `jQuery`, my app is bundled without warning, but when I execute it, I got an error where Bootstrap says it requires jQuery but can't find it, whereas au-cli has the (expected) following trace:
```
INFO [Bundler] Auto tracing package: 3.3.1      jquery
```

When I add the `jQuery` dependency back into my  `aurelia.json`,  everything works as expected.

Since `jQuery` is auto-traced, it is probably added (in the `vendor-bundle.js` file) AFTER the manually-traced `Bootstrap v3` library, even though bootstrap explicitly depends on `jQuery`. I guess that when `jQuery` is manually-traced, dependency resolution succeeds in placing `jQuery` before `Bootstrap v3` and thus it explains why my app works.

I expect my application not to crash when I follow the recipes. :-)

I want the conjunction of the two recipes to work (but I'm not sure it is even possible without a major rewrite), or at least have the legacy bootstrap's recipe corrected to reflect the incompatibility of the two recipes.

huochunpeng:
closes #955 again

### DIFF
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -132,7 +132,45 @@ exports.Bundle = class {
   }
 
   getBundledFiles() {
-    return uniqueBy(this.includes.reduce((a, b) => a.concat(b.getAllFiles()), []), (file) => file.path);
+    // Sort files by moduleId and deps to be sure they will always be
+    // concatenated in the same order, so revision hash won't change.
+    // https://github.com/aurelia/cli/issues/955#issuecomment-439253048
+    // Topological sort for shim packages.
+
+    let bundleFiles = uniqueBy(
+      this.includes.reduce((a, b) => a.concat(b.getAllFiles()), []),
+      file => file.path
+    ).sort((a, b) => {
+      // alphabetical sorting based on moduleId
+      if (a.moduleId > b.moduleId) return 1;
+      if (b.moduleId > a.moduleId) return -1;
+      return 0;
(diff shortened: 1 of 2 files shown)


huochunpeng:
@EisenbergEffect could we have another beta release just for this fix?

PR #970, opened 2018-11-14
fix(webpack.config.template.js): webpack resolving child node_modules

jbockle:
Addressing #961, when a symlinked package (i.e. a lerna monorepo) exists in node_modules, the child node_modules are resolved in webpack, creating duplicate module ids like './node_modules/symlinked-package/node_modules/some-dependency/index.js' in addition to './node_modules/some-dependency/index.js'.

PR #980, opened 2018-11-20
chore(webpack.config.template.js): alias not needed w/ absolute resolve

jbockle:
With #970 merged, the `aurelia-binding` alias is no longer required due to `resolve.modules` resolving an absolute path instead of relative.


huochunpeng:
With #970 reverted, this one is obsoleted.
@EisenbergEffect this can be closed.

PR #981, opened 2018-11-20
refactor(bundler): dry nodejs module resolving

huochunpeng:


huochunpeng:
@taosd a fix to simple-line-icons package is included here.

lib/build/utils.js
```diff
@@ -138,3 +131,89 @@ function modifyFilename(pth, modifier) {
...
+  return nodejsLoadIndex(resourcePath);
+}
+
+exports.nodejsLoad = function(resourcePath) {
+  return nodejsLoadAsFile(resourcePath) || nodejsLoadAsDirectory(resourcePath);
+};
+
+exports.removeJsExtension = function(filePath) {
+  if (path.extname(filePath).toLowerCase() === '.js') {
+    return filePath.substr(0, filePath.length - 3);
```
bigopon: We may want to use `slice(0, ... -3)` instead of this, according to this https://github.com/jsdom/jsdom/pull/2419

Though it's negligible, still feel good.

huochunpeng:
@bigopon interesting read from the jsdom issue. I replaced substr and substring with slice.

Issue #967, opened 2018-11-07
CLI generated webpack config uses a specific host:port for the dev server websocket client


chrisckc:
**I'm submitting a feature request**

**Current behavior:**
When running a CLI generated project based on webpack, the path used by the webpack dev server client to make it's websocket requests is unnecessarily specified to be the same as the host:port specified in the aurelia.json and run.ts files. (port is specified in aurelia.json and host in run.ts)

Specifying the host:port is not required as it is derived from window.location and output.publicPath in the webpack.config.

This does not cause a problem in a simple app, but becomes apparent when you start to do something like run behind a reverse proxy during development, like in the case of using Aurelia with .NET Core using the new SpaProxy. Refer to this issue that i logged and closed: https://github.com/aspnet/JavaScriptServices/issues/1743

This issue did not occur using Angular, because the webpack config behind the Angular CLI did not specify a specific host:port for the dev server client.

The fix is to change the following in 'aurelia_project/tasks/run.ts'

```
config.entry.app.unshift(`webpack-dev-server/client?http://${opts.host}:${opts.port}/`, 'webpack/hot/dev-server');
```

to this:
```
config.entry.app.unshift(`webpack-dev-server/client`, 'webpack/hot/dev-server');
```

* **What is the expected behavior?**
I should have been able to (or was hoping that i could) transplant the Angular app inside a `dotnet new angular` generated project with an Aurelia app with no changes required in the Aurelia app.

* **What is the motivation / use case for changing the behavior?**
Now that Microsoft have retired their Aurelia SPA template which is no longer available in dotnet v2.1 which uses the new SpaProxy based templates, making it easier to switch it over to Aurelia may help stem the resulting inevitable decline in Aurelia adoption rates for .NET users.

Alexander-Taran:
Would you like to provide a pull request for it @chrisckc ?

Issue #966, opened 2018-11-07
Inconsistency in CLI generated webpack project configuration


chrisckc:
**I'm submitting a bug report**

**Current behavior:**
Looking at the structure of a CLI generated project based on webpack, the webpack dev server client is added in both the index.ejs file and also in 'aurelia_project/tasks/run.ts' when HMR is used.

index.ejs contains this:
```
<% if (htmlWebpackPlugin.options.metadata.server) { %>
  <!-- Webpack Dev Server reload -->
  <script src="/webpack-dev-server.js"></script>
  <% } %>
```

run.ts contains this: 
```
if (project.platform.hmr || CLIOptions.hasFlag('hmr')) {
    config.plugins.push(new webpack.HotModuleReplacementPlugin());
    config.entry.app.unshift(`webpack-dev-server/client?http://${opts.host}:${opts.port}/`, 'webpack/hot/dev-server');
  }
```

As a result of this, when HMR mode is used, the script entry in index.ejs become redundant.

The script entry should be removed from index.ejs and run.ts should be changed to this:

```
  if (project.platform.hmr || CLIOptions.hasFlag('hmr')) {
    config.plugins.push(new webpack.HotModuleReplacementPlugin());
    config.entry.app.unshift(`webpack-dev-server/client?http://${opts.host}:${opts.port}/`, 'webpack/hot/dev-server');
  } else {
    config.entry.app.unshift(`webpack-dev-server/client?http://${opts.host}:${opts.port}/`);
  }
```

This does seem to cause a problem out of the box, but when attempting to run Aurelia from a subpath off the root, the redundant script entry produces an unnecessary complication.

* **What is the expected behavior?**
The generated config to be more consistent

* **What is the motivation / use case for changing the behavior?**
Provide a better development experience for more advanced scenarios

chrisckc:
It looks like it can also be simplified because according to the webpack v4 docs it is not necessary to add the "HotModuleReplacementPlugin" as it is done automatically when webpack is launched with the --hot option, which happens in package-scripts.js
https://webpack.js.org/configuration/dev-server/#devserver-hot

Alexander-Taran:
Would you like to provide a pull request for it @chaskim ?

chrisckc:
No problem, ill get on to it next week.
Btw. I tried removing the HotModuleReplacementPlugin as indicated could be possible in the webpack docs, but it broke hmr, so i guess that statement in the webpack docs in incorrect.

PR #982, opened 2018-11-22
fix(run-webpack.js run-webpack.ts index-webpack.ejs): update run scripts, remove redundant entry in index template

chrisckc:
fix(run-webpack.js run-webpack.ts index-webpack.ejs): update run scripts, remove redundant entry in index template

Fixes issue where the webpack dev server client is added in both the index.ejs file
and also in 'aurelia_project/tasks/run.ts' when HMR is used.
Improves config by removing unnecessarily specified host:port on the dev server client
This avoids config issues when running behind a reverse proxy

Fixes the following issues:
https://github.com/aurelia/cli/issues/966
https://github.com/aurelia/cli/issues/967

### DIFF
--- a/lib/resources/content/index-webpack.ejs
+++ b/lib/resources/content/index-webpack.ejs
@@ -8,9 +8,5 @@
     <!-- imported CSS are concatenated and added automatically -->
   </head>
   <body aurelia-app="main">
-    <% if (htmlWebpackPlugin.options.metadata.server) { %>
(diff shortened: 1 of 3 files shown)


EisenbergEffect:
@JeroenVinke I know you're not a Webpack expert, but if you can give this a general review that would be great. @fkleuver Could you give this a quick look over as well?

PR #988, opened 2018-11-28
fix: fresh build of package-lock.json to remove malicious flatmap-stream

huochunpeng:
The dependency of pstree.remy v1.1.0 -> ps-tree -> event-stream -> flatmap-stream has been cut to pstree.remy v1.1.2 with no deps.
Use merge2 to replace event-stream for TypeScript skeleton.
npm v6.4.1 reports "found 0 vulnerabilities".


EisenbergEffect:
ready to go @huochunpeng ?

huochunpeng:
Few mins, testing TS skeleton.

huochunpeng:
All good.

Issue #969, opened 2018-11-13
Improvements in CLI generated Webpack config


chrisckc:
**I'm submitting a feature request**
Improvement of generated webpack config in relation to bundle splitting, long term cachability and other aspects inline with the following document:
https://developers.google.com/web/fundamentals/performance/webpack/

* **Loader/bundler:**
Webpack (v4)

**Current behavior:**
Webpack config is not optimised

* **What is the expected behavior?**
Webpack config should have some basic level of optimisation and documentation should cover more advanced optimisations.

* **What is the motivation / use case for changing the behavior?**
To provide an improved out of the box development experience.

Following on from this: https://github.com/aurelia/cli/issues/958 i have learned more about webpack optimisations and tested some configs against different code both Aurelia and non-Aurelia.

Chunk splitting seems to be one of the most important aspects of optimisation provided that we are already using minification of both js and css.

The CLI generated webpack config does js minification in production mode as it enabled by default in webpack v4 but does not perform CSS minification without an additional plugin.

I will talk about chunk splitting first with some test results.

chrisckc:
Here are the results of tests using HTTP/1/1 and different webpack configs.

The test were conducted using a project which consists of a CLI generated project with the Skeleton source files transplanted. 
The repo for the above project is here: https://github.com/chrisckc/aurelia-cli-skeleton-typescript-webpack
I commented out the bootstrap sass import and replaced it with the bootstrap css import to simplify testing.
I stripped out the optimisation section from the config and used different optimisation configs as described below.

I used `au build --env prod` to build and then served using `npm start -- serve`
I then checked the timings in Chrome dev tool when requesting the home page.

#### Test 1 (default config)
Using the default splitChunks config, the following 4 files were downloaded by the browser:
-index.html: 893 bytes
-single app/vendor chunk: 524KB
-single app/vendor css file: 157KB (containing fontawesome & bootstrap)
-font-awesome woff2 font: 75.8KB

This is the request waterfall on Slow 3G:
![image](https://user-images.githubusercontent.com/2445837/48427553-e6cef680-e760-11e8-89f4-0d7be21801af.png)


Chrome DevTool timings with network throttling enabled:

Custom 4G (4.0Mb/s 50ms latency)
4 requests | 757KB transferred | finish 1.6s | Load 1.75s

Fast 3G
4 requests | 757KB transferred | finish 5.25s | Load 5.40s

Slow 3G
4 requests | 757KB transferred | finish 19.20s | Load 19.35s

[...]

PR #997, opened 2018-12-05
feat(webpack): Improvements in CLI generated Webpack config

chrisckc:
Improvement of generated webpack config in relation to bundle splitting, long term cachability and other aspects.
Added new question for HTTP protocol to be used in webpack optimisation config
New question only accessible when following the 'custom' path
Choices are HTTP/1.1 or HTTP/2
Default is HTTP/1.1 when the question is not presented.

PR for feature: https://github.com/aurelia/cli/issues/969

### DIFF
--- a/lib/commands/new/new-application.json
+++ b/lib/commands/new/new-application.json
@@ -54,6 +54,10 @@
           "id": "webpack",
           "displayName": "Webpack"
         },
+        "httpProtocol": {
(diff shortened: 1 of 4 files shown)


EisenbergEffect:
Thanks @chrisckc !
@huochunpeng Can you review this to ensure that the create project workflow looks correct?

PR #1002, opened 2018-12-19
fix: avoid early exist of "au new"

### LINKED ISSUE #977: au new command exits early when a local CLI is installed in the current folder

* **Library Version:**
1.0.0-beta.5

* **Operating System:**
OSX 10.x|Linux (distro)|Windows [7|8|8.1|10]

* **Node Version:**
6.2.0

* **NPM Version:**
3.8.9

* **Browser:**
all 
* **Language:**
all

* **Loader/bundler:**
all 

**Current behavior:**
If you accidentally install the Aurelia CLI locally rather than globally by running `npm install aurelia-cli` rather than `npm install aurelia-cli -g` then when you run the `au new` command the CLI exits immediately with the message `No Aurelia project found`

The expected behavior is that the CLI should either not exit early and go head to create the project. Or alternatively, show a more informative message like `Could not create Aurelia project, please remove local version of Aurelia CLI before running this command`.

In the past week, I've run into this situation two times when attempting to get new devs started with Aurelia. If we can get this workflow streamlined it makes the on-ramp for new developers much smoother.

huochunpeng:
Don't try to find local aurelia project when running "au new".
Always use global aurelia-cli to execute "au new".
Log cli version.

closes #977

### DIFF
--- a/bin/aurelia-cli.js
+++ b/bin/aurelia-cli.js
@@ -13,6 +13,10 @@ Please upgrade to latest Node.js https://nodejs.org`);
 
 process.title = 'aurelia';
 
+const userArgs = process.argv.slice(2);
+const commandName = userArgs[0];
+const commandArgs = userArgs.slice(1);
+
 let originalBaseDir = process.cwd();
 
 resolve('aurelia-cli', {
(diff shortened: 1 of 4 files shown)


EisenbergEffect:
Thanks @huochunpeng !

Issue #1001, opened 2018-12-11
Add alameda as loader option


fragsalat:
**I'm submitting a feature request**

* **Library Version:**
all

**Please tell us about your environment:**
* **Operating System:**
all

* **Node Version:**
all

* **NPM Version:**
all

* **Browser:**
all

* **Language:**
all

* **Loader/bundler:**
RequireJS / Alameda

**Current behavior:**
Right now the cli tool allows to select requirejs as loader.

* **What is the expected behavior?**
It should be possible for the user to select alameda or requirejs as default loader.

* **What is the motivation / use case for changing the behavior?**
Alameda targets modern browsers by utilizing Promises instead of interval cycles which improves the performance of the application.

**Todo**:
- [ ] Replace requirejs option with AMD Loader
- [ ] Add another step to select requirejs or alameda
- [ ] Distinguish in require templates between requirejs and alameda
- [ ] aurelia-karma.js must adopt more generic approach to prepend /base/ to urls by proxying nameToUrl function instead of requirejs.load and global.define. Alameda doesn't have requirejs.load

fragsalat:
Question to the code owner, why did you overwrote requirejs.load and created a custom normalize path function? It looks like you only wanted to add /base/ to all urls during test setup?
https://github.com/aurelia/cli/blob/master/lib/resources/test/require.aurelia-karma.js#L47

As far as I tested it and had a look it would be absolutely sufficient to do the following instead
```js
    const origToUrl = requirejs.nameToUrl;
    requirejs.nameToUrl = function extendedToUrl(moduleName) {
      let url = origToUrl(moduleName);
      if (url.indexOf('/base') !== 0) {
        url = `/base/${url}`;
      }
      return url;
    };
```

fragsalat:
Not yet. I'll first intregrate alameda sucessfully in my work's project and based on that create a PR here :)

PR #1003, opened 2018-12-20
feat(loader): Added Alameda AMD loader to cli flow

fragsalat:
This fixes: #1001 

As alameda and requirejs have the same public api I've just duplicated require loader and require.aurelia-karma.js and adjusted it a bit.
The option was added after requirejs e.g.
1. Webpack
2. RequireJS
3. Alameda
4. SystemJS

As alameda has no internal function `load` we could override, I've used the nameToUrl function which serves the same use case.
I've tested application run and test execution for all loaders with jasmine and one with jest.
Also I could not find a test to extend for this use case and did not add any.

### DIFF
--- a/lib/commands/new/buildsystems/cli/loaders/alameda.js
+++ b/lib/commands/new/buildsystems/cli/loaders/alameda.js
@@ -1,0 +1,32 @@
+'use strict';
+
+const ProjectItem = require('../../../../../project-item').ProjectItem;
+
+module.exports = function(project, options) {
+  let model = project.model;
+
+  project.loader = model.loader.id;
+  delete model.loader;
+
+  model.build.loader = {
+    type: 'require',
+    configTarget: 'vendor-bundle.js',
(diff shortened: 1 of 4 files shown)


fragsalat:
Interesting. The deprication should be fixed by updating to latest karma + jasmine and should appear in a clean requirejs build as well.

huochunpeng:
I just installed your branch as global cli, then creates a new app with alameda + karma.

fragsalat:
I have no clue why it worked previously but I fixed it. As mentioned the deprecation error can be fixed by updating jasmine / karma-jasmine but I couldn't find where the version numbers are held.

huochunpeng:
All work for me now.

fragsalat:
If everything is ok, whats the plan to proceed here?

EisenbergEffect:
Thanks @fragsalat ! Release coming soon.

fragsalat:
Thank you too :)

PR #1010, opened 2018-12-29
feat(bundler): support module resolution for lerna hoisting

huochunpeng:
When standard resolution failed, try app's local folder.


bin/aurelia-cli.js
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 'use strict';
-const resolve = require('../lib/resolve');
+const resolve = require('resolve');
```
EisenbergEffect: @huochunpeng Before I merge this...are we sure this is the only place where `resolve` was used?

huochunpeng: Yes, I checked.

EisenbergEffect:
Cool. Just wanted to be sure :)

PR #1015, opened 2019-01-11
fix(bundler): fix plugin prefix/subfix regex match

huochunpeng:
Fixed a bug when tracing a dep with name spaced plugin like "@au/plugin!./foo.html", it also traces module "@au/plugin".


EisenbergEffect:
😄

PR #1017, opened 2019-01-15
fix(webpack): correctly exclude spec/test files from coverage report

peitschie:
The string {spec,test} is not a valid regex and thus didn't match filenames containing spec or test, as was clearly intended.
This caused coverage reports to include these files.

The new regex now correctly excludes these test files.


huochunpeng:
I am confused, shouldn't it be `/\.(spec|test)\.[jt]s$/i` ?

peitschie:
@huochunpeng you are 100% correct. I even had the correct one on my other version of this patch at https://github.com/aurelia/skeleton-navigation/pull/896 ... I don't know how I stuffed that up! Good catch 😸

huochunpeng:
@EisenbergEffect please merge it.

Issue #1018, opened 2019-01-16
Running jest fails when adding bootstrap, popper.js  (navigator is undefined)


shahabganji:
Hello guys, 

I have added bootstrap and wanted to start unit testing with jest, then I faced the following issue:

![image](https://user-images.githubusercontent.com/4405518/51235825-b9cbd980-1985-11e9-9462-92b1acd7b129.png)

I think this is due to the fact that popper has been used with bootstrap, so after investigating through this error we find that the issue will be resolved if we add the following lines [here](https://github.com/aurelia/cli/blob/cb78d99a79d721e841fa4ddf4a648314edd1db97/lib/commands/new/buildsystems/general/unit-test-runners/jest.js#L29) and/or [here](https://github.com/aurelia/cli/blob/cb78d99a79d721e841fa4ddf4a648314edd1db97/lib/commands/new/buildsystems/general/unit-test-runners/jest.js#L67).

```json
"jest": {
...
 "globals": {
      "navigator": {
        "userAgent": "node.js"
      }
    },
...
```

[here](https://stackoverflow.com/questions/37078882/mocha-chai-throws-navigator-not-defined-due-to-react-router-component#comment94967911_37084875) is a stackoverflow thread pointing to the solution.

huochunpeng:
jest by default simulate a browser env which provides navigator as one of the global variables.

But Aurelia did something special to turn off the default.
1. it switches the default jest env to node, in your package.json, you can find
```js
"jest": {
  "testEnvironment": "node",
}
```
2. it uses special aurelia-pal-nodejs instead of normal aurelia-pal-browser, you can find in `test/jest-pretest.js`
```js
import {globalize} from 'aurelia-pal-nodejs';
globalize();
```

## The truth is, we don't need to do that.

Try this:
1. remove `"testEnvironment": "node",` from package.json
2. update `test/jest-pretest.js` to this:
```js
import 'aurelia-polyfills';
import {Options} from 'aurelia-loader-nodejs';
import path from 'path'; // this line will be different is you use TS
Options.relativeToDir = path.join(__dirname, 'unit');

// aurelia-pal-nodejs is what you see from some
// reference aurelia jest setup.
// To use aurelia-pal-nodejs, you need
// testEnvironment: "node" in jest.config.js

// import {globalize} from 'aurelia-pal-nodejs';
// globalize();

// jest actually can work with aurelia-pal-browser
// by default.
import {initialize} from 'aurelia-pal-browser';
initialize();
```

[...]

PR #1019, opened 2019-01-16
WIP feat(test): switch from aurelia-pal-nodejs to aurelia-pal-browser for jest setup

3cp:
jest by default does simulate browser environment. It's unnecessary to use pal-nodejs (which brings in jsdom ~~but doesn't pollute global vars~~). Using pal-browser with default jest simulation has extra benefit: be more compatible with 3rd party libs which expects global vars.

cloese #1018

### DIFF
--- a/lib/commands/new/buildsystems/general/unit-test-runners/jest.js
+++ b/lib/commands/new/buildsystems/general/unit-test-runners/jest.js
@@ -17,7 +17,7 @@ module.exports = function(project) {
     'jest-cli',
     'plugin-error',
     'aurelia-loader-nodejs',
-    'aurelia-pal-nodejs'
+    'jest-environment-jsdom-thirteen'
   ).addNPMScript('test', 'au jest');
 
   if (project.model.transpiler.id === 'babel') {
@@ -46,7 +46,7 @@ module.exports = function(project) {
       setupFiles: [
         '<rootDir>/test/jest-pretest.js'
       ],
-      testEnvironment: 'node',
+      testEnvironment: 'jest-environment-jsdom-thirteen',
       collectCoverage: true,
       collectCoverageFrom: [
         'src/**/*.js',
@@ -89,7 +89,7 @@ module.exports = function(project) {
       setupFiles: [
         '<rootDir>/test/jest-pretest.ts'
       ],
-      testEnvironment: 'node',
+      testEnvironment: 'jest-environment-jsdom-thirteen',
       collectCoverage: true,
       collectCoverageFrom: [
         'src/**/*.{js,ts}',
--- a/lib/dependencies.json
+++ b/lib/dependencies.json
@@ -66,6 +66,7 @@
   "jest": "^23.6.0",
   "jest-jasmine2": "^23.6.0",
   "jest-cli": "^23.6.0",
+  "jest-environment-jsdom-thirteen": "^0.0.2",
   "jest-matchers": "^20.0.3",
   "json-loader": "^0.5.7",
   "html-loader": "^0.5.5",
--- a/lib/resources/test/jest-pretest.js
+++ b/lib/resources/test/jest-pretest.js
@@ -1,6 +1,6 @@
 import 'aurelia-polyfills';
 import {Options} from 'aurelia-loader-nodejs';
-import {globalize} from 'aurelia-pal-nodejs';
-import * as path from 'path';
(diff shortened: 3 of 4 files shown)


huochunpeng:
@shahabganji let's know whether it works for you. Thx.

EisenbergEffect:
@huochunpeng What is the easiest way for @shahabganji to test this out and give us feedback on whether it will work better than the current setup?

shahabganji:
@EisenbergEffect , @huochunpeng Sorry I have not seen [this](https://github.com/aurelia/cli/pull/1019#issuecomment-454955770), I'll check it and will let you know within a couple of hours :pray:

EisenbergEffect:
No need to apologize @shahabganji We appreciate you jumping in whenever you can.

huochunpeng:
@shahabganji this is the same fix as the one posted on your issue.

shahabganji:
@huochunpeng 

> @shahabganji this is the same fix as the one posted on your issue.

I did not use them, since both of the approaches I've mentioned there also worked, however, I'll check your changes to make sure it works properly too :+1:

Vheissu:
I love the idea of removing the need for Node.js in the Jest tests, however, I am unable to get this working with TypeScript/Jest and Webpack based app. Any test that requires staging a component (like a modal) fails with some weird errors @huochunpeng check the Discord when you get some time.

shahabganji:
I checked the changes  @huochunpeng had made on the same repository I was working on, and jest just works seamlessly. Worth mentioning that I am using Typescript.

huochunpeng:
Waiting for a fix on aurelia-bootstrapper to remove the runtime dependency of aurelia-pal-nodejs.

huochunpeng:
- [ ] update `aurelia-bootstrapper` to avoid loading `aurelia-pal-nodejs` aurelia/bootstrapper#70
- [ ] need a polyfill for `MutationObserver` in jest(jsdom) env.
  * either move `aurelia-pal-nodejs/src/polyfills/mutation-observer.ts` to `aurelia-polyfills`.
  * or load the polyfill in `test/jest-pretest.js`. It means the app still has `aurelia-pal-nodejs` in dependencies.
- [ ] @Vheissu has an unsolved issue: `The class property of a object ([object HTMLDivElement]) cannot be assigned`.

huochunpeng:
Note: latest jsdom v13.2.0 has added MutationObserver!

huochunpeng:
@Vheissu could you try one more time with jest-environment-jsdom-thirteen?

Alexander-Taran:
I've tried jest with jsdom env 13 and 14
with pal-browser
problem is that object prototypes are not the same  and the setup fails at a simple `class="${something}"` or  `<element id="${something}"...`

@fkleuver can share some insights 

all in all there will be roadblocks with jest

I do get the appeal of jest as a no setup kitchen sink that gives feeling of a great achievement from the get go.
figuring out how to set up karma with jasmine took a lot of time that most devs don't have or are not willing to spend.

having a nice explained setup for karma/jasmine || karma/mocha  that works and provides close value to jest could be more beneficial to community

testing apps in real browsers gave me more value.
for one I had a feature detection of input[type=color] with branching..
can't get coverage for that with jsdom

I'll share my karma setup and we could improve upon it.

ghiscoding:
Hey guys, I'm not sure if that makes any difference but I followed this article [Unit-testing Aurelia with Jest + JSDOM + TypeScript + Pug](https://blog.sunshower.io/2018/12/18/unit-testing-aurelia-with-jest-jsdom-typescript-pug/) and installed `jsdom-global` with my `package.json` 
```json
    "jest-environment-jsdom": "^24.8.0",
    "jsdom": "^15.1.1",
    "jsdom-global": "^3.0.2",
```
and then `import 'jsdom-global/register';` in my `jest-pretest.ts`  before `globalize();`.
So now I have JSDOM... well I think anyway. 

Anyhow, the description of `jsdom-global` is
> jsdom-global will inject document, window and other DOM API into your Node.js environment. Useful for running, in Node.js, tests that are made for browsers.

I'm not sure if that helps anyone, but might be worth mentioning anyway. Doing this allowed me to get  rid of some errors I had before, for example, I had issues with `Blob` did not exist or was unknown, now this error is gone and I can continue with my tests. :)

3cp:
closed in favour of aurelia/pal-nodejs#34

PR #1026, opened 2019-01-26
feat(find-deps): add build tracing for import elements

EisenbergEffect:
This PR extends the build tracing to not only bundle deps from `<require>` elements but also from `<import>` elements in preparation for vNext's support for both options.

Partial work related to https://github.com/aurelia/aurelia/issues/116


EisenbergEffect:
@huochunpeng Here's the PR to add support for `<import>` elements. Let me know if you think it's ok to merge. Thanks!

EisenbergEffect:
Good point. I'll removed the GH close trigger and merge this. Then we can do the other work in various plugins and compilers later.

Issue #1027, opened 2019-01-30
Failing unit tests cli master


michaelw85:
**I'm submitting a bug report**

* **Library Version:**
Latest (1.0.0-beta.12)

**Please tell us about your environment:**
* **Operating System:**
Windows [10]

* **Node Version:**
10.15.0 (also tested 11.8.0)

* **NPM Version:**
6.4.1

* **Browser:**
N/A (Jasmine test run)

* **Language:**
JS

* **Loader/bundler:**
N/A

**Current behavior:**
**Given** you cloned the cli repo
**And** performed `npm install`
**When** you run `npm run test`
**Then**  Executed 318 of 318 specs (_36 FAILED_) in 44 secs.

Failures seem to be related to accessing file system. I've debugged and noticed fs.existsSync throws an error while it should return a boolean. This seems to be a change in behavior in node. Issue can be resolved by wrapping the method in a try/catch block.

* **What is the expected behavior?**
Expected all tests to pass on master.

* **What is the motivation / use case for changing the behavior?**
I would like to contribute to the cli repo. Unless all tests are green for untouched code I cannot modify code and guarantee its functioning as intended.

huochunpeng:
See my comments in PR. The real issue is mock-fs in nodejs v10.

huochunpeng:
For now, please use nodejs v8 to test your intended contribution.

huochunpeng:
Good news. https://github.com/tschaub/mock-fs/pull/260

PR #1028, opened 2019-01-30
fix(fs.existsSync): change existsSync to correctly handle thrown error

michaelw85:
existsSync throwns an error while it should return a boolean.
Wrapped function in a try/catch to surpress the error.

Partial fix for #1027

### DIFF
--- a/lib/file-system.js
+++ b/lib/file-system.js
@@ -25,7 +25,11 @@ exports.stat = function(path) {
 };
 
 exports.existsSync = function(path) {
-  return fs.existsSync(path);
+  try {
+    return fs.existsSync(path);
+  } catch {
+    return false;
+  }
 };
 
 exports.mkdir = function(path) {
--- a/lib/file-system.js
+++ b/lib/file-system.js
@@ -27,7 +27,7 @@ exports.stat = function(path) {
 exports.existsSync = function(path) {
   try {
     return fs.existsSync(path);
-  } catch {
+  } catch (err) {
     return false;
   }
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "aurelia-cli",
-  "version": "1.0.0-beta.10",
+  "version": "1.0.0-beta.12",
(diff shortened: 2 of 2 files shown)


huochunpeng:
Thanks for looking into this.

The reason of failing test is because of mock-fs, we waited for quite long, but it seems not going to be fixed. There is a link in our `.travis.yml`
```yml
os:
  - linux
  - windows
  - osx
language: node_js
node_js:
  - "8"
  # mock-fs doesn't work in nodejs v10 right now
  # https://github.com/tschaub/mock-fs/issues/238
  # - "10"
```

You cannot just silent `fs.existsSync`, because the `lib/file-system.js` is simple wrapper around the real fs.

1. If the real `fs.existsSync` throws exception, we should too.
2. The fact is the real `fs.existsSync` would not throw exception. https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_existssync_path
3. The exception you saw in test is purely caused by mock-fs in nodejs v10. That's why even when you silent the exception, it would not fix the tests in nodejs v10.

What we really need is to refactor our code and get ride of mock-fs in all unit tests. We really hate to do this, as mock-fs is so intuitive, but we don't have any choice.

michaelw85:
Thank you for the info, I also found something pointing to node and mock-fs but as you stated those issues dated back a long time with no clear info stating it will be fixed.

You are right suppressing any exception is a bit to much, would it be an option to check for the error that is thrown? This would only suppress a specific scenario.

huochunpeng:
The problem is mock-fs, not the behavior of `fs.exitsSync`. The "issue" only exists in tests which is not running real fs module, it is running mock-fs patched fs module which doesn't work in nodejs v10. If mock-fs was still there, there is nothing you can do to fix all the tests.

For now, pls use nodejs v8 for the cli work you intended.

For long term, we are going to remove mock-fs. It's not a light task, I just don't have enough time to contribute.

michaelw85:
Ok will use v8 for now.

Issue #1031, opened 2019-02-01
Readonly attribute on environment scipt file


lorond:
* **Library Version:**
aurelia-cli v1.0.0-beta.12

* **Operating System:**
Windows 10

* **Node Version:**
10.15.0

* **NPM Version:**
6.4.1

* **Browser:**
not applicable

* **Language:**
all

* **Loader/bundler:**
Webpack (didn't tested other)

**Current behavior:**
If target `environment.[tj]s` file is readonly, build fails when trying to copy corresponding file from `aurelia_project\environments`.

* **What is the expected behavior?**
As file created by aurelia on each build, readonly attribute should not fail entire build process.

* **What is the motivation / use case for changing the behavior?**
Team Foundation System with Server Repositories places readonly attribute on files that is not checked out. File `environment.[tj]s` is not exist on first build and copied with all attributes of source file, including readonly attribute as environment configuration files are rarely being edited and usually not checked out. Each subsequent build fails until either file or attribute removed. However, in this case source file copied again with all attributes. So it required to do manual cleanup before each build.

avrahamcool:
I also had the same problem back when I used TFS.
I altered the `transpile` and `copy-files` tasks - so they remove  the `readonly` flag on the target file.

keep in mind that the `cli` generated tasks have changed since then - so you need to carefully change yours accordingly.
also - this need to be fixed in the `ts` versions as well.

first: install `gulp-chmod'` (using npm or yarn)

**transpile.js**
``` js
import gulp from 'gulp';
import changedInPlace from 'gulp-changed-in-place';
import plumber from 'gulp-plumber';
import babel from 'gulp-babel';
import sourcemaps from 'gulp-sourcemaps';
import notify from 'gulp-notify';
import rename from 'gulp-rename';
import project from '../aurelia.json';
import {CLIOptions, build} from 'aurelia-cli';
import chmod from 'gulp-chmod'; //<-- added

function configureEnvironment() {
  let env = CLIOptions.getEnvironment();

  return gulp.src(`aurelia_project/environments/${env}.js`)
    .pipe(changedInPlace({firstPass: true}))
    .pipe(rename('environment.js'))
	.pipe(chmod(666)) //<-- added
    .pipe(gulp.dest(project.paths.root));
}

function buildJavaScript() {
  return gulp.src(project.transpiler.source)
    .pipe(plumber({errorHandler: notify.onError('Error: <%= error.message %>')}))
    .pipe(changedInPlace({firstPass: true}))
[...]

PR #1032, opened 2019-02-01
Remove readonly attribute on destination environment file

lorond:
Issue: https://github.com/aurelia/cli/issues/1031
Based on: https://github.com/gulpjs/gulp/issues/1057#issuecomment-99211021


lorond:
need to revise, found bug

avrahamcool:
what bug did you find?

huochunpeng:
If your problem is testing...

Currently tests only runs on nodejs v8, not v10.

Check `.travis.yml` for more details.

lorond:
> keep in mind you need to manually delete the destination file that was created before your fix (because it already have the readonly flag).

yep, that is why I simply want to remove readonly flag on destination file.

PR #1034, opened 2019-02-04
Delete destination environment file before writting new

lorond:
issue: #1031

### DIFF
--- a/lib/resources/tasks/environment.js
+++ b/lib/resources/tasks/environment.js
@@ -12,16 +12,7 @@ function configureEnvironment() {
   return gulp.src(`aurelia_project/environments/${env}${project.transpiler.fileExtension}`)
     .pipe(rename(`environment${project.transpiler.fileExtension}`))
     .pipe(through.obj(function (file, enc, cb) {
-      // remove readonly attribute on destination file
-      const targetFile = `${project.paths.root}/${file.relative}`;
-      fs.stat(targetFile, function (err, stat) {
-        if (err) {
-          cb(null, file); // most likely, no such file yet
-        }
-        else {
-          fs.chmod(targetFile, stat.mode | 0o200, function () { cb(null, file); });
-        }
-      });
+      fs.unlink(`${project.paths.root}/${file.relative}`, function () { cb(null, file); });
     }))
     .pipe(gulp.dest(project.paths.root))
     .pipe(through.obj(function (file, enc,  cb) {
--- a/lib/resources/tasks/environment.js
+++ b/lib/resources/tasks/environment.js
@@ -11,6 +11,18 @@ function configureEnvironment() {
 
   return gulp.src(`aurelia_project/environments/${env}${project.transpiler.fileExtension}`)
     .pipe(rename(`environment${project.transpiler.fileExtension}`))
+    .pipe(through.obj(function (file, enc, cb) {
+      // remove readonly attribute on destination file
+      const targetFile = `${project.paths.root}/${file.relative}`;
+      fs.stat(targetFile, function (err, stat) {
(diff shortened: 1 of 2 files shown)


huochunpeng:
@JeroenVinke I am not sure is this too much for 90% of users who do not use TFS? Is there any other option we have (like hiding this behind a custom menu)?

avrahamcool:
why change the target file `readonly` flag?
if we are about to replace the whole file one line after that.
why don't just delete the file to make sure that a new one can be safely created?

I thought you wanted to change the `readonly` flag on the source file while copying it to the destination - so the target will always be without `readonly`.

on the same note:
is that the best way to handle environment settings?
by copying a file into the `src` right before `build`?
it's annoying to have a file in the `src` folder - who need's to **not** be checked-in into source control.
is there a better to way to handle this?

JeroenVinke:
@lorond thanks for looking into this and creating the pr!

I also think that there are better ways to handle these environment settings. While we look into that, people who use tfs and run into this problem can use the fix that’s in this PR.

huochunpeng:
Probably need to add an entry here https://aurelia.io/docs/cli/recipes
The repo is here https://github.com/aurelia/documentation/blob/master/current/en-us/11.%20cli/7.%20recipes.md

@lorond would you like to help on the doc?

lorond:
> if we are about to replace the whole file one line after that.
> why don't just delete the file to make sure that a new one can be safely created?

Thought, it would fail to delete readonly file, but it isn't. Changed it.

> @lorond would you like to help on the doc?

Sure

huochunpeng:
@lorond LOL, just found out `gulp.dest` support changing mode out of box.
```js
gulp.src('a.js')
.pipe(gulp.dest('out', {mode: parseInt('666', 8)}));
```

Or if you want more strict, `parseInt('644', 8)`.

lorond:
@huochunpeng not sure if it is a good solution as it changes mode after file copied, not before.
Eventually we need to overwrite artifacts of previous build. I guess, cleanup concept would be more appropriate, and as @avrahamcool noted preliminary file removal fits better.

huochunpeng:
I tested gulp.dest thing, it works very well when source file is read only, it ensures target file is not read only.

lorond:
Yep, but consider it already readonly - someone run build without solution, fount fix, integrated it and run build again - build will fail. Removing readonly is kind of hack, removed entire file - normal cleanup (and could be implemented as separate task).

avrahamcool:
is this PR going to be merged?

it's a beautiful solution for a common TFS problem.
it will not affect those who are not using TFS.
its far better than just pointing the user to a recipe in the docs.

huochunpeng:
@EisenbergEffect @JeroenVinke how do you think of bringing in this fix?

@lorond the new files to be touched are
```
skeleton/cli-bundler/aurelia_project/tasks__if_babel/transpile.js
skeleton/cli-bundler/aurelia_project/tasks__if_typescript/transpile.ts
skeleton/webpack/aurelia_project/tasks/environment.ext
```

EisenbergEffect:
I'm fine with merging this if someone can resolve the conflicts.

lorond:
I'll resolve them conflicts.
@huochunpeng  thanks.

huochunpeng:
@lorond guess you forgot to push --force. Also, please rewrite commit message to follow our [Git Commit Guidelines](https://github.com/DurandalProject/about/blob/master/CONTRIBUTING.md#commit). I guess this one is a "fix(all): ..." as the bug was in both cli-bundler and webpack based projects.

Thanks for the contribution!

lorond:
done #1118

PR #1036, opened 2019-02-04
fix(jest): avoid aurelia-bindings v1+v2 conflict

huochunpeng:
This is a similar fix to jest as #906 to webpack. Thank @ariovistus to point this out at https://discourse.aurelia.io/t/jest-cannot-find-module/2217/3


huochunpeng:
I tested it, @EisenbergEffect this can be merged.

PR #1042, opened 2019-02-07
feat(ui): add Enquirer as UI lib

jwx:
Add Enquirer as the UI lib responsible for all information gathering from the user.


huochunpeng:
I am surprised to see `+230 -229`, I was expecting something like `+230 -2290` 😆

jwx:
@JeroenVinke @huochunpeng Can you (and anyone else that wants to) review please?

PR #1043, opened 2019-02-08
chore: print option index from 1 instead of 0

huochunpeng:


huochunpeng:
@jwx told me the reason for 0 based is because he has not figured out how to use 1 based short-cut in enquirer.

jwx:
> @jwx told me the reason for 0 based is because he has not figured out how to use 1 based short-cut in enquirer.

There's currently no support for it in Enquirer.

Issue #959, opened 2018-10-29
add async/await by default to new project with ESNext


jrgcubano:
**I'm submitting a feature request**

* **Library Version:**
1.0.0-beta.3

* **Language:**
ESNext

* **Loader/bundler:**
SystemJS | RequireJS

* **Current behavior:**
Not async/away by default

* **What is the motivation / use case for changing the behavior?**
It would be wonderful to have async/await by default or at least addible with the new project options.

PR #1046, opened 2019-02-10
fix(bundler): enable tracing @babel/polyfill

huochunpeng:
A simple global var "global" is created before load all AMD modules. This is a practical way to support nodejs global var "global". In future we will build on top of dumberjs to properly support nodejs global vars "global", "process", and "Buffer" (the last two are still not supported transparently in aurelia-cli bundler).

core-js uses "var define = ..." everywhere, we need to force cjs wrap before we can switch to dumberjs bundler.

With this fix, `@babel/polyfill` now works as an AMD module.

### DIFF
--- a/lib/resources/src/main-cli.js
+++ b/lib/resources/src/main-cli.js
@@ -1,3 +1,4 @@
+import '@babel/polyfill';
 import environment from './environment';
 
 export function configure(aurelia) {
--- a/lib/build/bundle.js
+++ b/lib/build/bundle.js
@@ -189,7 +189,8 @@ exports.Bundle = class {
 
     if (loaderOptions.configTarget === this.config.name) {
       work = work.then(() => {
-        files.push({ contents: '_aureliaConfigureModuleLoader();'});
+        // create global var "global" for compatibility with nodejs
+        files.push({ contents: 'var global = this; _aureliaConfigureModuleLoader();'});
       });
     }
 
--- a/lib/build/bundled-source.js
+++ b/lib/build/bundled-source.js
@@ -216,7 +216,10 @@ exports.BundledSource = class {
         let contents;
         // forceCjsWrap bypasses a r.js parse bug.
         // See lib/amodro-trace/read/cjs.js for more info.
-        let forceCjsWrap = !!modulePath.match(/(\/|\\)(cjs|commonjs)(\/|\\)/i);
+        let forceCjsWrap = !!modulePath.match(/(\/|\\)(cjs|commonjs)(\/|\\)/i) ||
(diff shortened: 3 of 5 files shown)


lib/resources/src/main-cli.js
```diff
@@ -1,3 +1,4 @@
+import '@babel/polyfill';
```
EisenbergEffect: I'm confused. Why is the CLI dependent on Babel?

huochunpeng: It's was turned on previously to enable #959. Before this fix, it was in prepend list for built-in bundler based app.

huochunpeng: The other fact is that we have babel polyfill in our webpack app skeleton from day1 I think.

EisenbergEffect: Apologies. I misread this code. This is a file that's being scaffolded into a new project, not the CLI source code itself. I was trying to understand why our CLI code directly needed babel, but it doesn't.

huochunpeng: Actually cli code does... :-)
The bundler use babel to support esm format.

EisenbergEffect: Ugh. Is there any way not to have a dependency on Babel?

huochunpeng: Actually it always have, even before auto-tracing.

cli uses babel to support tasks in exnext, uses typescript to support tasks in typescript.

I argued we should rewrite all tasks in pure commonjs, in order to remove babel and typescript from dependencies.

huochunpeng: But that (removing babel/ts) of-course is a breaking change.

EisenbergEffect: I know it uses it when you generate an ESNext project. Is there any way to not use it directly in the CLI source itself?

huochunpeng: Use babel to do just esm transform is an overkill, there definitely is some other smaller lib for just that. I can do some investigation.

For the tasks part, the babel and ts are runtime dependencies. And we previously don't have them directly in aurelia-cli package.json, but that created plenty of trouble for aurelia-cli developers. Remember those special part in README about npm install after npm link?

@JeroenVinke added those runtime deps (babel and ts) to aurelia-cli package.json, to ease the experience on developing aurelia-cli.

EisenbergEffect: Ah yes. I remember that issue now. Let's keep things as is then. Not a high priority to change this for now. I had just forgotten we had that dependency in the CLI itself.

PR #1047, opened 2019-02-14
chore: remove bluebird, use the promise polyfill in babel/polyfill

huochunpeng:
This is not a breaking change to existing apps. Existing apps with bluebird are not affected at all. This only affects new app.
The promise polyfill provided by babel is very good, for example, it provides promise.prototype.finally on browsers have native promise but do not provide method finally. It guarantees to normalise browser.

Manually tested on all skeletons generated by `au generate-skeletons`. **Cleanup some files, and fixed some minor problems with skeletons**. All apps run, pass unit tests, pass e2e tests.

Manually tested few more skeletons without navigation scaffold (`au generate-skeletons` uses navigation scaffold for all setups). All apps run, pass unit tests, pass e2e tests.

This improve Aurelia performance on IE, as Bluebird is known being very slow on IE.


zewa666:
hey @huochunpeng would the same babel polyfill be used for TS projects as well?

huochunpeng:
It is for all ts and js projects.

Previously we had babel polyfills in prepend of all cli bundler based setup (both ts and js). Now it’s more obvious because it is in code.

avrahamcool:
this is great.

I have a question that may not regard this specific PR,
what is the additional benefits of `aurelia-polyfill` over `babel-polyfill`?

huochunpeng:
Aurelia-polyfills is part of Aurelia, because Rob had this view that Aurelia core should has no 3rd party dependencies. Which is good for Aurelia’s stability.

Aurelia-polyfills only fills the parts that Aurelia needed. But it did not cover promise, probably because promise polyfill is not very small (correct me if I am wrong @EisenbergEffect).

babel/polyfill is definitely bigger and more complete. When using babel with babel/polyfill, user can set option for `preset-env` to smartly reduce actual bundled polyfill size based on target browsers.

Currently there is no way to remove aurelia-polyfill from your app, because it's imported internally by aurelia-bootstrapper.

You can, however, remove babl/polyfill from your app, what you lose are:
1. promise polyfill on old browsers (I mean IE), and some other missing parts not covered by aurelia-polyfills.
2. some promise method even on new browsers, like `Promise.prototype.finally`.
3. the needed polyfill for babel to support transpiled `async` `await` syntax.

avrahamcool:
I need IE support in my projects, and I find it hard to configure.
I'm happy with that PR because it'll remove some part of the work (the part regarding Bluebird configuration with `no-long-stack-trace` because of IE)

in that same thought, can we have a step in the `cli` that ask you about older browser support?
and in the case you need it - will configure your project accordingly?
1. fetch polyfill
2. Bluebird (no longer an issue after this PR)

huochunpeng:
> can we have a step in the cli that ask you about older browser support?

Good idea. I am currently doing some refactoring work to simplify cli workflow/scaffolding. Then I will follow up with aurelia plugin scaffolding.

I will investigate new options after those two tasks.

PR #1048, opened 2019-02-14
fix(bundler): isolate cache between systemjs and requirejs apps

huochunpeng:
Previously, rarely, but when user has both systemjs app and requirejs app, the bundler could use a wrong cache.


huochunpeng:
The bug was found while I test various skeletons.
@EisenbergEffect this is safe to be merged.

PR #1053, opened 2019-02-18
fix(jest task): Return error when tests fail

### LINKED ISSUE #1052: au jest task passes when it should fail/throw error/exit 1

* **Library Version:**
Local aurelia-cli v1.0.0-beta.12

* **Operating System:**
[Azure Pipelines Hosted VS2017 image](https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/win/Vs2017-Server2016-Readme.md) (Windows Server 2016)

* **Node Version:**
10.14.2

* **NPM Version:**
6.5.0

* **Browser:**
N/A

* **Language:**
TS (Bug probably lang independent)

* **Loader/bundler:**
Webpack (Bug probably occurs for any)

**Current behavior:**
When running `au jest` with a failing test cli resolves as if it passed.
Problem is located here: [Line 16](https://github.com/aurelia/cli/blob/master/lib/resources/tasks/jest.ts)
```js
if(result.numFailedTests || result.numFailedTestSuites) {
// Never takes this path because result does not contain numFailedTests
// numFailedTests is inside a results key within result
...
} else {
  cb()
}
```

result console.log:
```js
{ globalConfig:
   { bail: false,
     changedFilesWithAncestor: false,
     //.....
  },
  results:
   { numFailedTestSuites: 0,
     numFailedTests: 0,
     numPassedTestSuites: 4,
     numPassedTests: 10,
     // ....
```

Possible solution:
```js
// Use object destructuring to access results
jest.runCLI(options, [path.resolve(__dirname, '../../')]).then(({ results }) => {
// Use results here instead of result.
}
```

If a test fails the task should fail.

Azure build pipeline proceeds while tests are failing, a unit test build step should fail when there are failing tests.

michaelw85:
Callback always resolved successfully even when test failed
Incorrect data was accessed in result data to determine failure

Closes #1052

### DIFF
--- a/lib/resources/tasks/jest.js
+++ b/lib/resources/tasks/jest.js
@@ -1,8 +1,7 @@
 import jest from 'jest-cli';
-import PluginError from 'plugin-error';
 import path from 'path';
 import packageJson from '../../package.json';
-import {CLIOptions} from 'aurelia-cli';
+import { CLIOptions } from 'aurelia-cli';
 
 export default (cb) => {
   let options = packageJson.jest;
@@ -15,7 +14,7 @@ export default (cb) => {
 
   jest.runCLI(options, [path.resolve(__dirname, '../../')]).then(({ results }) => {
     if (results.numFailedTests || results.numFailedTestSuites) {
-      cb(new PluginError('jest-cli', { message: 'Tests Failed' }));
+      cb('Tests Failed');
     } else {
       cb();
     }
--- a/lib/resources/tasks/jest.js
+++ b/lib/resources/tasks/jest.js
@@ -13,9 +13,9 @@ export default (cb) => {
 
   process.env.BABEL_TARGET = 'node';
 
-  jest.runCLI(options, [path.resolve(__dirname, '../../')]).then((result) => {
-    if (result.numFailedTests || result.numFailedTestSuites) {
-      cb(new PluginError('gulp-jest', { message: 'Tests Failed' }));
+  jest.runCLI(options, [path.resolve(__dirname, '../../')]).then(({ results }) => {
+    if (results.numFailedTests || results.numFailedTestSuites) {
+      cb(new PluginError('jest-cli', { message: 'Tests Failed' }));
     } else {
       cb();
     }
--- a/lib/commands/new/buildsystems/general/unit-test-runners/jest.js
+++ b/lib/commands/new/buildsystems/general/unit-test-runners/jest.js
@@ -15,7 +15,6 @@ module.exports = function(project) {
   ).addToDevDependencies(
     'jest',
     'jest-cli',
-    'plugin-error',
     'aurelia-loader-nodejs',
     'aurelia-pal-nodejs'
   ).addNPMScript('test', 'au jest');
--- a/lib/dependencies.json
+++ b/lib/dependencies.json
@@ -81,7 +81,6 @@
   "minimatch": "^3.0.4",
   "node-sass": "^4.9.3",
   "opn": "^5.4.0",
-  "plugin-error": "^1.0.1",
   "protractor": "^5.4.1",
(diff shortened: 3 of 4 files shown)


michaelw85:
Result after change in azure devops pipeline forcefully making a test fail:

image

lib/resources/tasks/jest.js
```diff
@@ -13,9 +13,9 @@ export default (cb) => {
 
   process.env.BABEL_TARGET = 'node';
 
-  jest.runCLI(options, [path.resolve(__dirname, '../../')]).then((result) => {
-    if (result.numFailedTests || result.numFailedTestSuites) {
-      cb(new PluginError('gulp-jest', { message: 'Tests Failed' }));
+  jest.runCLI(options, [path.resolve(__dirname, '../../')]).then(({ results }) => {
+    if(results.numFailedTests || results.numFailedTestSuites) {
```
huochunpeng: Could you leave a space between `if` and `(` for both jest.js and jest.ts? I see it was not there in jest.ts before, `if(` looks weird without space.

lib/resources/tasks/jest.js
```diff
@@ -13,9 +13,9 @@ export default (cb) => {
 
   process.env.BABEL_TARGET = 'node';
 
-  jest.runCLI(options, [path.resolve(__dirname, '../../')]).then((result) => {
-    if (result.numFailedTests || result.numFailedTestSuites) {
-      cb(new PluginError('gulp-jest', { message: 'Tests Failed' }));
+  jest.runCLI(options, [path.resolve(__dirname, '../../')]).then(({ results }) => {
+    if (results.numFailedTests || results.numFailedTestSuites) {
+      cb(new PluginError('jest-cli', { message: 'Tests Failed' }));
```
huochunpeng: It looks like (existing code) `new PluginError` generated excessive output in the console. To my limited understanding of gulp, it seems this is not a place for PluginError because it's not within context of a plugin.

Following gulpjs's example, `cb(new Error('Tests Failed'));` is enough and probably the correct way. @JeroenVinke can you have a look?

lib/resources/tasks/jest.js
michaelw85: @huochunpeng Updated the PR. Seems like it was already present in the TS file, I guess I just copy pasted it and it did not auto format. (We could use a prettier config in the cli repo or even better a prettier pre-commit hook)

lib/resources/tasks/jest.js
michaelw85: I have hardly any experience with Gulp so left it in place, I agree the output seems excessive but it resolved the issue which was more important in my case. I will replace it with an error in my local setup and test to see if it also forces a build failure (I assume it does). If it needs to be replaced please let me know and I will update the PR.

huochunpeng: BTW, it looks jest is only feature in cli that uses plugin-error as a dep. If we decide to get rid of PluginError here, we can cleanup the `dependencies.json` and `unit-test-runners/jest.js` too.

michaelw85: When using `new Error('...')` the build fails as expected:
image

@huochunpeng I think we can replace the PluginError.

michaelw85: Today I also noticed just returning a string in the callback is enough to make it show an error.

huochunpeng: > Today I also noticed just returning a string in the callback is enough to make it show an error.

That's even better! Let's just clean up everything.

michaelw85: @huochunpeng Updated and removed plugin-error package.

huochunpeng:
@EisenbergEffect this can be merged.

PR #1058, opened 2019-02-20
fix(cypress task): Return error when tests fail

### LINKED ISSUE #1057: Cypress task should fail in build when test(s) fail

* **Library Version:**
Local aurelia-cli v1.0.0-beta.12

* **Operating System:**
[Azure Pipelines Hosted VS2017 image](https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/win/Vs2017-Server2016-Readme.md) (Windows Server 2016)

* **Node Version:**
10.14.2

* **NPM Version:**
6.5.0

* **Browser:**
N/A

* **Language:**
TS (lang independent)

* **Loader/bundler:**
Webpack (Bug probably occurs for any)

**Current behavior:**
When running `au cypress --run` with a failing test a build step does not fail.
Similar issue (different root cause): #1053 

**Root cause**
Callback is not resolved. It should resolve without a message for success and with a error message on failure.

If a test fails the task should fail.

Azure build pipeline proceeds while tests are failing.

michaelw85:
Make au cypress --run emit error when a test failed to allow a build process/step to fail.

Closes #1057

### DIFF
--- a/lib/resources/tasks/cypress.js
+++ b/lib/resources/tasks/cypress.js
@@ -2,9 +2,12 @@ import cypress from 'cypress';
 import { CLIOptions } from 'aurelia-cli';
 import config from '../../cypress.config';
 
-export default () => {
+export default (resolve) => {
   if (CLIOptions.hasFlag('run')) {
-    cypress.run(config);
+    cypress
+      .run(config)
+      .then(results => (results.totalFailed === 0 ? resolve() : resolve('Run failed!')))
+      .catch(resolve);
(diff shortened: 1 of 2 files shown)


michaelw85:
@huochunpeng This PR is similar to #1053

michaelw85:
Updated resolve to cb.

Here's the result of a failed run btw:
image

huochunpeng:
@EisenbergEffect this can be merged.

PR #1062, opened 2019-02-25
fix(sourcemaps): strip path from bundle sourcemap

Alexander-Taran:
having bundle name specified with path in aurelia.json
bundler produces a sourcemap comment with path included
which leads to sourcemap not being loaded by browser


huochunpeng:
LGTM @EisenbergEffect

PR #1065, opened 2019-02-27
fix(bundler): improve compatibility with legacy libs that depends on jquery or momentjs

huochunpeng:
Special treatment for jquery and moment, put them in front of everything else, so that jquery and moment can create global vars as early as possible. This improves compatibility with some legacy jquery plugins.
Note as of momentjs version 2.10.0, momentjs no longer exports global object in AMD module environment. There is special code in lib/build/amodro-trace/write/defines.js to bring up global var "moment".

### DIFF
--- a/lib/build/amodro-trace/write/defines.js
+++ b/lib/build/amodro-trace/write/defines.js
@@ -29,6 +29,13 @@ function defines(options) {
         config = context.config,
         packageName = context.pkgsMainMap[moduleName];
 
+    if (packageName === 'moment') {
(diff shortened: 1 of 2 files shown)


huochunpeng:
cc @jrgcubano

@EisenbergEffect this is a dirty fix, but it does remove some pain when using legacy libs. Tested, safe to merge.

EisenbergEffect:
It's unfortunate that we have to do this but these two libraries are popular enough and have caused enough issues that I think it's a price we have to pay. Thanks @huochunpeng !

Issue #179, opened 2016-06-22
generate TypeScript d.ts for aurelia-cli


atsu85:
# reason
Currently when selecting TypeScript as a transpiler the build scripts are also generated in TypeScript, but IDE can't do much without the typescript definitions.

EisenbergEffect:
I agree. I'm not sure how best to handle this yet. There's more to it than the cli d.ts. We can certainly create one of those. But, we need gulp and all the plugins really.

atsu85:
@EisenbergEffect, 
> But, we need gulp and ...

if You referred to TypeScript IDE support for writing gulp files, then see #183

jasoncarreira:
This also affects the generators since without the cli .d.ts the generator code is a mess

EisenbergEffect:
The cli api is a bit unstable currently. We can certainly add a d.ts just remember this is alpha and it is likely to change over the next few months.

jasoncarreira:
I can deal with changes... I just can't build right now ;)

EisenbergEffect:
I don't understand why you can't build. Can you explain?

EisenbergEffect:
Ok, that compilation is done on the fly, automatically by the Aurelia CLI itself, as needed. So, there's no need to have the ide touch that. However, we can work to address this in the future. Is there a way to have your IDE ignore that folder for now?

jasoncarreira:
Yes, I can ignore it and just use the CLI. Seeing the red is just an annoyance, I suppose, but I've been trained by my IDE to go and fix things highlighted in red.

atsu85:
@jasoncarreira, I've managed to fix most of the errors with [that PR](https://github.com/atsu85/aurelia-cli-experiment/pull/1#issuecomment-229428773).

beatsm:
I currently have this problem with Visual Studio. As soon as I press F5 I get the build errors and then the CLI will also no longer build with the below error.

(function (exports, require, module, __filename, __dirname) { define(["require", "exports", 'gulp', 'gulp-changed-in-place', 'gulp-plumber', 
'gulp-sourcemaps', 'gulp-notify', 'gulp-rename', 'gulp-typescript', '../aurelia.json', 'aurelia-cli'], function (require, exports, gulp, 
changedInPlace, plumber, sourcemaps, notify, rename, ts, project, aurelia_cli_1) {
                                                              ^

ReferenceError: define is not defined
[...]

PR #1067, opened 2019-03-03
refactor(Build) Add typescript compiler

michaelw85:
Refactor project to run from dist folder and add typescript transpiler.

Related to #179


huochunpeng:
Please hold TS for few more days :-) There is one big refactoring PR coming soon (in hrs).

michaelw85:
@huochunpeng Ok, there is no rush. I did make the build pass so I it is ready for review. I do not expect allot of merge conflicts.

What this PR does is add a build setup with a Typescript compiler and copies resources + non js/ts files to the dist folder. I tried to touch a minimum amount of existing code, I did change 2 files to TS extension to prove the setup works and it also enables the first bare minimum for generating d.ts files.

After this is merged I/we can proceed with the next step which is getting TS support setup for jasmine and after that files can be converted one by one.

huochunpeng:
I am touching lots of release-check files in #1075. Please hold this one bit longer.

michaelw85:
> I am touching lots of release-check files in #1075. Please hold this one bit longer.

NP I was just trying to resolve merge conflict. I also want to add support for unit tests before getting this merged. I will update this to WIP

michaelw85:
@huochunpeng now that #1075 is merged would it be possible to get this reviewed and possible even merged?

huochunpeng:
It is going to be reviewed by other team members as I am not a TS user.

Also we are preparing final v1 release. This is probably safer to leave TS refactoring to v1.1. cc @EisenbergEffect

veikkoeeva:
@michaelw85 I'm a n00b but would like to change the TS setup, so asking a bit about the appoach. :)

I took the CLI output and modified it a bit, see at https://github.com/veikkoeeva/MultioutputWebpack (it should work if you remove the `webpack-babel-multi-target-plugin` from the pipeline). The idea wasn't to transpile to `es5`, but the highest possible and then use Babel to transpile down. For performanc etc. use the usual methods of forking and that (the demo is so little it doesn't seem to matter so didn't go through it). So it appears the difference in our approach comes to use of Babel at least.

I did first "the usual" of using Babel 7 to just strip away types, but ran into trouble with decorators. The problem basically boils down into decorator metadata and is explained at https://discourse.aurelia.io/t/typescript-babel-7/2381/6 (see towards the end of the short thread).

After that I thought that maybe a better approach would be to use TS as usually, transpile to highest JS and then use Babel to build a `es5` or some other package. Basically to set up [differential serving](https://calendar.perfplanet.com/2018/doing-differential-serving-in-2019/). It looks a bit involved with setting one or more entries, so I searched solutions and spotted an excellent looking plugin [webpack-babel-multi-target-plugin](https://www.npmjs.com/package/webpack-babel-multi-target-plugin), but ran into a problem, see at https://github.com/DanielSchaffer/webpack-babel-multi-target-plugin/issues/28. It's probably related to how Aurelia pipeline works, but I do I have very new [Babel 7.4](https://babeljs.io/blog/2019/03/19/7.4.0) with the brand new [core-js 3.0](https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md) set up too.

As a side note, I removed the direct link to `core-js` in the Aurelia source code and let BabylonJS to load and it dropped the size of the (dev) bundle to 1/3. I also moved Webpack to TS in one branch, but I'm not sure if this is something anyone would want. :D I'm not also sure if source maps will cause problem once Babel transpilation works (differential serving or not) and how would one solve it.

michaelw85:
@veikkoeeva Thanks for your input!
I do not consider myself an expert on TS and builds so any input is appreciated. My goal with this PR was to get something going. A bare minimum with the least amount of impact allowing future conversion to TS and creation of dts files.

I currently do not have time to completely go over your comment and examples but I did want to reply. I hope to be able to have a look later this week.

If you want you can fork my branch, adapt it and create a new improved PR.

veikkoeeva:
@michaelw85 It's OK, I don't want to imply you need to do anything. You did go further in making something useful to others already. If this is useful, all the better. Maybe gives thoughts to someone.

michaelw85:
@3cp This PR is going stale and it would be a waste to delete it. Any update when this could be reviewed & merged?

3cp:
We said we are going to delay TS conversion to post v1 release. Because we are very close to v1 release.

Since I am not a TS user, probably @bigopon will help to review this change. I can ensure you we are not ignoring this contribution:)

michaelw85:
Hearing that it's not ignored is reassuring.

How can I track any progress on CLI releases? I feel like I'm a bit in the dark when it comes to any releases related to Aurelia in general. I have no idea where to go to get a overview.

tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+      "outDir": "./dist",
+      "allowJs": true,
+      "target": "es5"
```
bigopon: I think it's safe to go for 6 and beyond

bigopon:
This conversion looks simple enough for a merge so if we don't have any issues running from dist, I think we are safe to go.
About release process, pinging @EisenbergEffect

tsconfig.json
michaelw85: This was set by default, will update it. I agree 6 should be fine, here's the node documentation confirming the support: https://nodejs.org/en/docs/es6/

3cp: fyi, cli requires at least nodejs v8.9.0, which is es6+ I think.

michaelw85: Updated

michaelw85:
@bigopon @3cp Conflicts resolved again, tests passed and target was updated.

bigopon:
From your comment of going agile style for this, can you give some info about your plan for next steps?

michaelw85:
After introducing the transpiler the natural next step is to start using typescript. This could either be just for new things to get used to TS and start leveraging it's power or could include changing any file you touch slowly migrating as you go.

I personally would like to 
- start changing files to the ts extension as a first step.  
(This should require relatively low effort to make it transpile.)
- then typings can gradually be introduced 
- and ts config could be made stricter over time (like not supporting js as a start)
- tslint with vnext config can be added

What also becomes possible with introducing the transpiler is to get rid of the duplicate task files (js & ts variant). I will make a separate github issue for this to describe my thoughts.

michaelw85:
@EisenbergEffect  @bigopon Anything left that should be done for this PR to be merged?

EisenbergEffect:
@michaelw85 Just wanted to let you know we aren't ignoring you. We're dealing with a TS-related issue in the DI library. Once we get that all sorted out, we'll circle back to this. Just didn't want you wondering what was going on.

michaelw85:
@EisenbergEffect Thank you for the explanation I was actually at a point where I started to think to close this PR and delete te changes.

package.json
```diff
@@ -13,22 +13,28 @@
...
-    "au": "bin/aurelia-cli.js"
+    "aurelia": "dist/bin/aurelia-cli.js",
+    "au": "dist/bin/aurelia-cli.js"
   },
   "engines": {
     "node": ">=8.9.0"
   },
   "scripts": {
+    "build": "npm run build:ts && npm run build:dts && npm run build:copy-files",
+    "build:ts": "./node_modules/.bin/tsc",
```
bigopon: npm auto uses local `node_modules` first, so probably we don't need to be this specific? Besides that, if someone chose to install with any name that is not `node_modules`, this would break 😓

3cp: pretty sure `"build:ts": "tsc"` does exact same thing.

michaelw85: My primary reason for being specific in these paths is that this enforces using the project installed packages. I might be mistaken but I think node looks for globally installed packages first and then starts looking from the local dir up for node_modules folders.

Secondary I have had numerous situations where not specifically pointing to node modules failed to make chained commands work. I just tried to remove all from the commands and even though I would also expect it to work it does not, copy files for instance fails:
```cli
'copyfiles' is not recognized as an internal or external command,
operable program or batch file.
```
Unless I start using globally installed modules I don't think this will work without being this specific. 

@bigopon I just searched a bit on changing the dir, so far I only see the ability to change the prefix for globally installed modules. Local installation using `npm i` and `npm ci` will always end up in `node_modules`. So this setup would be just fine if someone would decide to change the global install.

3cp: I think npm script will enforce local package for you. Same as you run `npx tsc` manually in console.

fkleuver: @michaelw85
Here's how it works:
- Given a script defined in `package.json`: `"my-script": "foo"` if you call `npm run my-script`, npm will try to call `node_modules/.bin/foo` inside the local package. That is the convention - npm does not deviate from that unless you specifically give it the `--global` flag.
- npm itself runs globally, so it uses the current working dir to figure out its scope. If any of the scripts change the current working directory, npm can lose its scope and fall back to the global.
- An absolute path is no guarantee that it will work either. That will depend on whether the resolved path has already been cached or not - not something you want to rely on.

So, in this case. `"build:ts": "./node_modules/.bin/tsc",` doesn't do any harm but it will not do any good either. npm literally translates `"build:ts": "tsc",` to that: https://github.com/npm/cli/blob/latest/lib/npm.js#L381

fkleuver:
Just my 2c, it looks fine to me. The thing with the package.json npm script is just a nitpick as it has no functional impact, it's just a verbose alias like `npm install` instead of `npm i`.

package.json
michaelw85: @fkleuver Could you checkout my branch, delete the `./node_modules` entries from the json and tell me if copy_files works for you when using the build task. With your explanation it should work flawlessly (which I was also expecting) but it does not.

fkleuver: @michaelw85 
Here's what I did:
```
git clone https://github.com/michaelw85/cli.git aurelia-cli-2
cd aurelia-cli-2
git checkout initial_ts_setup
npm ci
```

Then I opened `package.json` and did a find/replace of `./node_modules/.bin/` with empty string. 5 occurrences replaced.

Then I ran `npm run build` and by the looks of it all succeeded:
image

I do not have `copyfiles` installed globally. I am running node `12` if it matters.

michaelw85: That is interesting, it did not work for me. I'm running node 10.15.1
I will give it a try on my own device when I get home if it fails I will upgrade and try again.

michaelw85: @fkleuver I tried again, deleted node_modules, npm ci, replace paths and run build and it works fine.

@bigopon @3cp Removed './node_modules/...'

fkleuver:
@EisenbergEffect @3cp @bigopon Let's get this in?

EisenbergEffect:
@michaelw85 Thanks for working on this! I wanted to share that for Aurelia vNext, we are planning to take a slightly different approach. Please have a read through of the RFC here: https://github.com/aurelia/aurelia/issues/509 This might be a good opportunity for you to get more deeply involved with Aurelia and vNext.

michaelw85:
@EisenbergEffect thx for the info. Any particular area you think I could/should be involved?

EisenbergEffect:
@michaelw85 If you are interested in tooling, there's opportunities to work on the cli helpers, the convention system, AoT, VS Code integration, and debug plugin. Have a read through the issue above and poke around the vnext repo a bit and let me know what seems interesting to you. We'll get you hooked in.

michaelw85:
Any of those sounds interesting, I guess AOT would be the most challenging. I've done some work on vscode plugins in the past, really enjoyed that.

I will have a look at the vnext repo asap

fkleuver:
@michaelw85 I am working on AOT as we speak, so if you want to help out with integrating it into VS Code that would be perfect. What did you work on previously? Autocomplete etc?

3cp:
I am reviewing the build.

@michaelw85 why we need this to make the final npm package full of duplicated skeleton files. Those skeleton files are static resources, I see no reason to duplicate them.

If need to access the skeleton files, might need to change the code that reads those static resources from aurelia-cli pacakge root, instead of from the relative location of the code `dist/lib/..`.

```
"build:copy-files": "copyfiles -e ./skeleton/**/*.{js,ts} -e ./lib/**/*.{js,ts} ./lib/**/* ./skeleton/**/* ./dist",
```

Also I have to add 
```
"pretest": "npm run build && npm run lint",
    "prepare": "npm run build",
```
To ensure build before test and publish.

Another thing I am doing is to switch from Yarn to NPM in release-check to probably handle `aurelia/cli#master` dep prepare scripts for the build. Yarn has bug yet to be fixed in this use case.

3cp:
And somehow the json file is not copied over to dist/.

```
⋊> ~/playground au new some
Global aurelia-cli v1.0.2
Invalid Command: new
                      _ _          ____ _     ___
  __ _ _   _ _ __ ___| (_) __ _   / ___| |   |_ _|
 / _` | | | | '__/ _ \ | |/ _` | | |   | |    | |
| (_| | |_| | | |  __/ | | (_| | | |___| |___ | |
 \__,_|\__,_|_|  \___|_|_|\__,_|  \____|_____|___|

Error: Cannot find module '../new/command.json'
Require stack:
- /Users/huocp/aurelia/cli/dist/lib/commands/help/command.js
- /Users/huocp/aurelia/cli/dist/lib/cli.js
- /Users/huocp/aurelia/cli/dist/lib/index.js
- /Users/huocp/aurelia/cli/dist/bin/aurelia-cli.js
    at Function.Module._resolveFilename (internal/modules/cjs/l
```

3cp:
The reason that unit tests didn't catch this, is probably due to "jasmine-ts" works directly on lib/ files, not the compiled dist/ files.

3cp:
Given the complications around compilation.

I am not TS user, but is this an approachable solution to let typescript to load at runtime for every `lib/**/*.ts` file? Like the way we support ts gulp task in final app project.

Runtime transpiling can remove all the compilation steps and file duplications. But some TS user might have better understand of this topic.

This is how we handled ts gulp task.
```js
function installTypeScript() {
  let ts = require('typescript');

  let json = require.extensions['.json'];
  delete require.extensions['.json'];

  require.extensions['.ts'] = function(module, filename) {
    let source = fs.readFileSync(filename);
    let result = ts.transpile(source, {
      module: ts.ModuleKind.CommonJS,
      declaration: false,
      noImplicitAny: false,
      noResolve: true,
      removeComments: true,
      noLib: false,
      emitDecoratorMetadata: true,
      experimentalDecorators: true
    });

    return module._compile(result, filename);
  };

  require.extensions['.json'] = json;
}
```

michaelw85:
@fkleuver 
> @michaelw85 I am working on AOT as we speak, so if you want to help out with integrating it into VS Code that would be perfect. What did you work on previously? Autocomplete etc?

Yes I've created my own autocomplete plugin at my previous employer to help write e2e tests using the custom layer I created on top of wdio. I also contributes to a plugin that provided jump to code for js code using with require.

michaelw85:
> @michaelw85 why we need this to make the final npm package full of duplicated skeleton files. Those skeleton files are static resources, I see no reason to duplicate them.

Because normally you only package `dist`, there is no need to ship the source with your package.

> And somehow the json file is not copied over to dist/.

I will have to tweak the copy command a bit, should be an easy fix.

> The reason that unit tests didn't catch this, is probably due to "jasmine-ts" works directly on lib/ files, not the compiled dist/ files.

Yes it will compile on the fly, this uses the same configuration as the build process so in terms of transpiling it is identical. Loading files is done relative so yes this would work in a test but fail if the build command does not copy a json file.

> I am not TS user, but is this an approachable solution to let typescript to load at runtime

Using node-ts that could work although it gives unnecessary overhead in my opinion. I made a mistake in the copy files task when introducing a completely new setup. I don't think we should blow this out of proportions. Maybe I could create a post build task that checks the total amount of files, compares src with dist. I think they should be equal in amount, this should be an easy script to verify the copy task is configured properly.

3cp:
Sketfoler is NOT source code to be compiled. Please don’t copy them, there is no point.

Set “files” field in package.json to include bin dist skeleton folders. That should be enough.

3cp:
If you skip compiling bin folder, only compile lib to dist instead of dist/lib, then you wouldn’t have issue on relative path to skeleton folder.

michaelw85:
> If you skip compiling bin folder, only compile lib to dist instead of dist/lib, then you wouldn’t have issue on relative path to skeleton folder.

I could do that, if that's what you prefer. 

I think it's weird for anything inside `dist` to require something outside of the folder. I always build my packages so that dist is what is shipped, anything outside of dist will not be published. This gives you full control over what you want or do not want to publish as a package.

3cp:
The fact is there is always things outside dist folder published, including package.json which can load from the code inside dist.

3cp:
Many libs load their own package.json in js code to print out version, read config like jest.

3cp:
BTW, in the following up PR, please run through release-check locally to identify potential problems. This change touched many files, it might have few more small complications. There are few adjustments you need on the release-check implementation, see my previous comments. Thx!

In release check, you need to target your branch head instead of `aurelia/cli#master`.

3cp:
@michaelw85 just check with you, what's the roughly expected time frame for the fix?

michaelw85:
Sorry for the delay, I tried to start multiple times but other things came up. Bit short on time after work. I might be able to have a look tonight. If a proper fix is not possible I will at least send a pr for to include the missing file.

3cp:
That's fine. Take your time to have a proper fix, we will merge if the whole release-check passes.

PR #1068, opened 2019-03-04
fix(jest): Sass require in html

### LINKED ISSUE #67: Test command-bundle
Create tests for the executable command `bundle`

michaelw85:
Fix error when testing a component that has a require tag with a
sass file.

Closes https://github.com/aurelia/testing/issues/67

### DIFF
--- a/lib/resources/test/jest-pretest.js
+++ b/lib/resources/test/jest-pretest.js
@@ -1,6 +1,7 @@
 import 'aurelia-polyfills';
-import {Options} from 'aurelia-loader-nodejs';
-import {globalize} from 'aurelia-pal-nodejs';
+import { ExtensionHandlers, Options } from 'aurelia-loader-nodejs';
+import { globalize } from 'aurelia-pal-nodejs';
 import * as path from 'path';
 Options.relativeToDir = path.join(__dirname, 'unit');
+ExtensionHandlers['.scss'] = (path) => new Promise((resolve) => resolve(path));
 globalize();
--- a/lib/resources/test/jest-pretest.ts
+++ b/lib/resources/test/jest-pretest.ts
@@ -1,6 +1,7 @@
 import 'aurelia-polyfills';
-import {Options} from 'aurelia-loader-nodejs';
-import {globalize} from 'aurelia-pal-nodejs';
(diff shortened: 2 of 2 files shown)


huochunpeng:
There is the other generic jest recommended way. https://jestjs.io/docs/en/webpack#handling-static-assets

If don't to write the mock module, can use https://github.com/eddyerburgh/jest-transform-stub

michaelw85:
@huochunpeng Thank you for the info, I will try that proxy setup. If it resolves this issue I will change this PR.

michaelw85:
@huochunpeng I've looked at the documentation and this setup only works for static content loaded via import statements in your js/ts files. The issue reported describes a flow where a sass file is loaded using a `<required from="./style.scss">` tag. This results in aurelia-loader, aurelia-templating-resources (css-resource.js `fixupCSSUrls`) to handle the file which fails for sass files.

huochunpeng:
:-)
You should check. It works. https://github.com/huochunpeng/demo-au-jest-scss

Removing `".+\\.scss$": "jest-transform-stub"` from package.json will make it fail.

Even with `<require>` in html, in jest testing, aurelia-loader-nodejs will use nodejs to load the resource anyway.

michaelw85:
Thank you for the example, I will give it another try. I think I missed the transform entry in the config :/

michaelw85:
@huochunpeng I can confirm installing jest-transform-stub and adding it as a custom transform entry there is no need for the change in this PR. Do you think we should/could provide this solution out of the box in aurelia cli?

huochunpeng:
> Do you think we should/could provide this solution out of the box in aurelia cli?

Absolutely. Let’s get it in after the big refactoring PR.

michaelw85:
@huochunpeng ok I will close this PR and create a new one after the big refactor PR is done. Where can I track the status of the PR you are referring to?

huochunpeng:
#1069

michaelw85:
Some additional info on stubbing, for me it still fails when adding jest-tranform-stub to the moduleNameMapper. Only adding it to transform seems to work fine (haven't testing imports from TS yet)

PR #1069, opened 2019-03-04
feat: unattended mode for "au new", descriptive skeleton

huochunpeng:
This is a refactoring on scaffolding logic.
1. Refactored `lib/` folder into two `lib/` + `skeleton/`, reduced about 3.3K LOC through descriptive skeleton logic.
2. Increased `spec/` folder about 1K LOC (even after removing lots of unneeded tests for dropped code). Added `npm run coverage` to check test coverage.

@EisenbergEffect pls don't panic about the dropped code 😄, I assure you we don't need them.
1. This new scaffolding has much better test coverage.
2. release-check has been enhanced to cover all `au generate` and new `au cypress`.
3. `au generate-skeletons` now generates 64 skeletons by default.
4. **all 64 skeletons passed release-check on both macOS and Windows10. It is safe to be released.**
5. documented descriptive skeleton. README now has links to descriptive skeleton and release-check doc.

### What is unattended mode
`au new project-name --unattended` to generate apps with all default choices.
`au new project-name --unattended --select typescript,jest` to generate apps with some override choices. Or in short form `au new project-name -u -s typescript,jest`

We don't need doc for the choices list. As every time user manually creates an app, there will be a project summary like this.
```
📖  Project Summary 
   This project was bootstrapped based on following selected features:
   cli-bundler requirejs web typescript sass karma scaffold-minimum
   Next time, you can generate a new project faster in unattended mode with the same features:
    au new nice-app --unattended --select cli-bundler,typescript,sass,karma 
   Or use the shorter form:
    au new nice-app -u -s cli-bundler,typescript,sass,karma 
   Note the command might not supply all the selected features, as some of them could be from default offering.
```

Note, you can use `--select` without `--unattended` in interactive mode, the choices you supplied in command line will help aurelia-cli to skip related questions.

### What is descriptive skeleton
First, acknowledge @jwx for the idea of "everything is a feature, with a flat skeleton folder structure to match every feature".

For details, please read [DESCRIPTIVE_SKELETON.md](https://github.com/huochunpeng/cli/blob/descriptive-skeleton/DESCRIPTIVE_SKELETON.md)

The implementation is mainly in [`lib/workflow/write-project/`](https://github.com/huochunpeng/cli/tree/descriptive-skeleton/lib/workflow/write-project).
I recommend you to have a look of the source code.
1. it uses gulp to avoid boilerplate on file operations (read and write).
2. it uses gulp to design a pipeline for descriptive skeleton.
3. it's fully tested.


EisenbergEffect:
I'm very tempted to just push the green button and merge, especially because of the nice test work :) However, it would be good to have someone with more recent CLI dev experience take a look at this to provide a second pair of eyes: @JeroenVinke @jwx Do you have time this week?

jwx:
@EisenbergEffect Yeah, I can find the time to take a look within a couple of days.

JeroenVinke:
I can take a look tonight

JeroenVinke:
I wanted to create a pull request for the commit I made but for some reason GitHub is giving me a 404 when doing it through the web editor :/

.eslintrc.json
```diff
@@ -15,10 +15,7 @@
     }]
   },
   "parserOptions": {
-    "sourceType": "script",
-    "ecmaFeatures": {
-      "impliedStrict": false
```
JeroenVinke: I can vaguely remember nodejs complaining about missing use stricts. Is this no longer an issue because we require a higher nodejs version now?

lib/commands/generate-skeletons/command.js
```diff
@@ -1,9 +1,86 @@
...
 const CLIOptions = require('../../cli-options').CLIOptions;
-const ProjectTemplate = require('../new/project-template').ProjectTemplate;
 const logger = require('aurelia-logging').getLogger('generate-skeletons');
+const selectFeatures = require('../../workflow/select-features');
+const writeProject = require('../../workflow/write-project');
+
+// Designed to cover all unit/e2e/transpiler combinations.
+// Cover decent css/postcss/htmlmin combinations.
+const projectDefs = [
+  'cli-bundler sass htmlmin-min jest dotnet-core protractor scaffold-navigation',
```
JeroenVinke: I like the fact that it's so readable :)

lib/commands/generate-skeletons/command.js
```diff
@@ -1,9 +1,86 @@
-'use strict';
 const UI = require('../../ui').UI;
 const fs = require('../../file-system');
+const _ = require('lodash');
```
JeroenVinke: did we already have this dependency? what are we using it for?

lib/commands/generate-skeletons/project-definitions.json
```diff
@@ -1,71 +0,0 @@
-{
```
JeroenVinke: glad to see this replaced by `'webpack typescript jest protractor'` :)

lib/commands/help/command.js
```diff
@@ -33,20 +32,12 @@ module.exports = class {
...
+    const commands = [
       require('../generate/command.json'),
       require('../config/command.json'),
       require('./command.json')
     ];
 
-    let bundler = this.project.model.bundler;
-    if (!bundler || bundler.id === 'cli') {
-      commands = commands.concat([
-        require('../import/command.json'),
```
JeroenVinke: Will this mean that people using the old bundler might miss these commands?

lib/commands/new/command.js
```diff
@@ -1,47 +1,226 @@
...
-      state.defaultOrCustomOverride = 'custom';
+    await this.ui.log('');
+    await this.logTitle('📖', 'Project Summary');
+    await this.logBody(`This project was bootstrapped based on following selected features:
+${c.green(features.join(' '))}
+Next time, you can generate a new project faster in unattended mode with the same features:
+${cmd(`au new ${wrappedName} --unattended${feats.length ? ' --select ' : ''}${feats.join(',')}${this.plugin ? ' --plugin' : ''}${this.here ? ' --here' : ''}`)}
+Or use the shorter form:
+${cmd(`au new ${wrappedName} -u${feats.length ? ' -s ' : ''}${feats.join(',')}${this.plugin ? ' -p' : ''}${this.here ? ' -h' : ''}`)}
+Note the command might not supply all the selected features, as some of them could be from default offering.`);
```
JeroenVinke: Not sure if I understand what this means. Shouldn't you always get the features that are specified in the command?

lib/commands/new/command.js
```diff
@@ -1,47 +1,226 @@
...
+  }
 
-      if (state.name.indexOf('/') !== -1) {
-        let parts = state.name.split('/');
-        state.name = parts[parts.length - 1];
-      } else if (state.name.indexOf('\\') !== -1) {
-        let parts = state.name.split('\\');
-        state.name = parts[parts.length - 1];
+  async optionallyInstallDeps(packageManager, projectFolder) {
+    await this.logTitle('🚚', 'Project Dependencies');
```
JeroenVinke: wow what's this :)

JeroenVinke: is this going to work in every terminal?

JeroenVinke:
Great work @huochunpeng! :) This should make it very easy to add new features by the community. Left some comments here and there for you to consider. Not sure if we should go full lodash, or if we can do without the extra dep

.eslintrc.json
huochunpeng: It's probably due to some removed embedded 3rd-part libs (concat-source-map, ...), not by this PR though.

Some of those libs might not work in strict mode.

lib/commands/generate-skeletons/command.js
huochunpeng: I used lodash in various places to avoid long tedious array modification.

It's not in explicit dependencies, but it was in some deep dep tree of some other deps we have.

lib/commands/help/command.js
huochunpeng: `au import` and `au install` had been removed since v1.0.0-beta.1 by the auto-tracer. Since then, they were only print out depreciation warning without doing anything.

lib/commands/new/command.js
huochunpeng: If you don't specify some features, the default choice will be picked automatically in unattended mode.

So `au new some --unattended` will generate app with all default choices.
`au new some --unattended --select jest,vscode` will generate an app as if you chose "Default ESNEXT app".
`au new some --unattended --select typescript,jest,vscode` will generate an app as if you chose "Default Typescript app".

lib/commands/new/command.js
huochunpeng: Nope, emoji (utf8) does not print out properly on Windows PowerShell, but it did not hurt much. I am open to remove them, but they are cool :-)

huochunpeng:
Two examples why I use lodash, readable and compact (lodash is in my blood now):

https://github.com/huochunpeng/cli/blob/e5680fd33e16df1e14ed93c9d6246e5bddb63781/lib/workflow/applicable.js#L16-L29

```js
  const expression = _(condition.toLowerCase())
    // Super simple tokenizer: split on underscore or space or &&,||,!.
    // Note with capture group in split regex, the &&,||,! are retained
    // in result array.
    .split(/_|\s|(&&|\|\||!)/)
    .map(_.trim)
    .compact()
    .map(c => {
      if (c === 'and' || c === '&&') return '&&';
      if (c === 'or' || c === '||') return '||';
      if (c === 'not' || c === '!') return '!';
      return features.includes(c) ? 'true' : 'false';
    })
    .join(' ');
```

https://github.com/huochunpeng/cli/blob/e5680fd33e16df1e14ed93c9d6246e5bddb63781/lib/workflow/write-project/write-project.js#L113

```js
const text = _(instructions).pick(skipped).values().join('\n');
```

lodash itself doesn't have any additional dependencies. `Array.prototype` does not have enough common methods to enable some part of the chain, also `Array.prototype` method will complain about non-array caller, while lodash silents the null/undefined input (means I can save one line of null check sometimes).

BTW, lodash is in our dep tree anyway even if you don't use it. Join the lodash minions my friend 😄 
```
aurelia-cli@1.0.0-beta.13 /Users/huocp/aurelia/cli
├─┬ @babel/core@7.2.2
│ ├─┬ @babel/generator@7.3.0
│ │ └── lodash@4.17.11  deduped
│ ├─┬ @babel/traverse@7.2.3
│ │ └── lodash@4.17.11  deduped
│ ├─┬ @babel/types@7.3.0
│ │ └── lodash@4.17.11  deduped
│ └── lodash@4.17.11  deduped
├─┬ @babel/plugin-transform-modules-amd@7.2.0
│ └─┬ @babel/helper-module-transforms@7.2.2
│   └── lodash@4.17.11  deduped
├─┬ @babel/register@7.0.0
│ └── lodash@4.17.11  deduped
├─┬ gulp-conventional-changelog@2.0.10
│ └─┬ conventional-changelog@3.0.6
│   └─┬ conventional-changelog-core@3.1.6
│     ├─┬ conventional-changelog-writer@4.0.3
│     │ ├─┬ handlebars@4.1.0
│     │ │ └─┬ async@2.6.2
│     │ │   └── lodash@4.17.11  deduped
│     │ └── lodash@4.17.11  deduped
│     ├─┬ conventional-commits-parser@3.0.1
│     │ └── lodash@4.17.11  deduped
│     └── lodash@4.17.11  deduped
├─┬ gulp-eslint@5.0.0
│ └─┬ eslint@5.13.0
│   ├─┬ inquirer@6.2.2
│   │ └── lodash@4.17.11  deduped
│   ├── lodash@4.17.11  deduped
│   └─┬ table@5.2.2
│     └── lodash@4.17.11  deduped
├── lodash@4.17.11 
└─┬ nyc@13.3.0
  └─┬ istanbul-reports@2.1.1
    └─┬ handlebars@4.1.0
      └─┬ async@2.6.2
        └── lodash@4.17.11 
```

huochunpeng:
> This should make it very easy to add new features by the community.

This is indeed the intend. I am going to follow up with a PR to scaffold Aurelia plugin project.

JeroenVinke:
Using a dependency while not declaring it in package.json has caused issues before: 
https://github.com/aurelia/cli/issues/801. There are tools which dedupes node_modules that uses package.json files to determine which packages are used

lib/commands/new/command.js
JeroenVinke: I like them too, but not if they do not work everywhere

lib/commands/new/command.js
JeroenVinke: Ah ok :) I didn't understand that at first from reading it, but it could very well be just me

JeroenVinke: Maybe something like: Some features get installed by default, so those are omitted from the command,

JeroenVinke:
I'll leave it up to you whether or not to add the dependency. Given that many of our dependencies already use it I don't see it causing problems

huochunpeng:
The lodash is added to package.json in this PR.

huochunpeng:
I will cleanup the others tomorrow. Thx for spending time reviewing this!

huochunpeng:
@JeroenVinke messages were cleaned up now.

lib/commands/new/command.js
huochunpeng: Removed.

lib/commands/new/command.js
huochunpeng: Updated accordingly.

huochunpeng:
@JeroenVinke is it ready to be merged?

JeroenVinke:
I think so. Maybe @jwx is still planning to have a look?

EisenbergEffect:
I'm happy to merge. I don't want to cut out @jwx though in case he wants to look it over first. Just let me know. I'm going to be doing some releases next week, so if we can get this resolved in the next few days, we can get it out then.

PR #1078, opened 2019-03-27
fix(bundler): remove sourceRoot from produced bundled sourcemaps

Alexander-Taran:


huochunpeng:
Just tested a cli-bundler app, the source map still works.

One small down side is the path of the source file is not now exactly correct `http://localhost:9000/scripts/app.js` instead of `http://localhost:9000/src/app.js`. Not a big issue since all source content is inlined in generated source map, so browser didn't have to fetch `http://localhost:9000/scripts/app.js` (which is wrong) to read source code.

Alexander-Taran:
It also impacts the coverage report..
if we could make root to src.. maybe that could be better.
I'll try

PR #1080, opened 2019-03-29
fix(skeleton): fix Promise polyfill on IE

### LINKED ISSUE #1079: 'Promise' is undefined on IE11

* **Library Version:**
v1.0.0-beta.14

* **Operating System:**
Windows

* **Node Version:**
10.15.2

* **NPM Version:**
6.9.0

* **Browser:**
IE11

* **Language:**
all

* **Loader/bundler:**
all

**Current behavior:**

The default apps (both webpack and cli-bundler) don't work on IE11, showing error 'Promise' is undefined.

Expect 'Promise' is poly-filled properly in IE11.

This is a regression introduced in #1047 which I didn't test in IE11 properly.

Root Cause: the first line in `src/main.js` or `src/main.ts`: `import 'any-polyfill';` is not the first code executed in aurelia app, it is BEHIND aurelia-bootstrapper.

aurelia-bootstrapper uses Promise way before user's main.js could do the polyfill.
bootstrapper uses Promise here:
https://github.com/aurelia/bootstrapper/blob/6b8437a8faf73b56baf6b2fdf953e73ef723760c/src/index.js#L7
before loading main module here:
https://github.com/aurelia/bootstrapper/blob/6b8437a8faf73b56baf6b2fdf953e73ef723760c/src/index.js#L162-L163

This timing issue was not a problem before, because we used to using prepend (for built-in bundle), or ProvidePlugin (for webpack) to load bluebird. They happens before aurelia-bootstrapper kicks in.

I am not sure can we fix this timing issue from bootstrapper. Alter the timing? Try load main module before using Promise? The issue is I think aurelia-loader also uses Promise.

If that's not possible, we will have to revert to using prepend/ProvidePlugin to load core-js (current polyfill). But this is less intuitive to end users.

cc @EisenbergEffect

huochunpeng:
Removed core-js as the import is too late for IE11.
Load promise-polyfill before loading aurelia-bootstrapper. The polyfill can be removed if user don't need to support IE.

closes #1079

### DIFF
--- a/lib/dependencies.json
+++ b/lib/dependencies.json
@@ -29,7 +29,6 @@
   "clean-webpack-plugin": "^1.0.1",
   "connect-history-api-fallback": "^1.6.0",
   "copy-webpack-plugin": "^5.0.0",
-  "core-js": "^3.0.0",
   "cross-env": "^5.2.0",
   "cssnano": "^4.1.10",
   "css-loader": "^1.0.0",
--- a/lib/dependencies.json
+++ b/lib/dependencies.json
@@ -82,6 +82,7 @@
   "opn": "^5.4.0",
   "postcss-loader": "^3.0.0",
   "postcss-url": "^8.0.0",
+  "promise-polyfill": "^8.1.0",
   "protractor": "^5.4.2",
   "regenerator-runtime": "0.13.2",
   "requirejs": "^2.3.6",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "aurelia-cli",
-  "version": "1.0.0-beta.14",
+  "version": "1.0.0-beta.15",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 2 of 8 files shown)


huochunpeng:
@EisenbergEffect passed release-check on selected skeletons to cover cli-bundler/webpack/babel/typescript/web/dotnet/plugin.

The polyfill.io is pretty smart, on ever-green browser, it responds a empty polyfill so there is no overhead for good browsers. Tested, IE11 is now working.

@bigopon please help to review.

huochunpeng:
Added Promise.prototype.finally for browsers have Promise but has no finally() method.

huochunpeng:
Put this one on hold. I am evaluating https://github.com/taylorhakes/promise-polyfill which is better for aurelia app deployed in closed environment.

huochunpeng:
Changed to use promise-polyfill, it polyfills both missing `Promise` and/or missing `Promise.prototype.finally`.

huochunpeng:
Passed my tests on few selected skeletons (webpack/cli-bundler/babel/typescript) on both Windows IE11 and Mac Safari.

huochunpeng:
@EisenbergEffect can we get this in?

PR #1084, opened 2019-04-06
fix(bundler): avoid appending alias after comment (sourceMap line).

huochunpeng:
This fixes tracing npm package "noty".


huochunpeng:
cc @arnederuwe

@EisenbergEffect safe to merged.

PR #1085, opened 2019-04-08
fix(skeleton): for TS plugin, generate d.ts files, allow import {...} from "resources"

huochunpeng:
Thank @kennyfowler and @Stefan for the improvement.


huochunpeng:
@EisenbergEffect tested on affected skeletons. safe to merge.

PR #1087, opened 2019-04-14
fix(bundler): fix sourceMap source file path and sourceRoot

huochunpeng:
The source path should be relative path, not the absolute disk file path. sourceRoot is removed in final bundle source map.


huochunpeng:
@Alexander-Taran can you try whether this one can replace #1078?

huochunpeng:
Tested the generated source map on both windows and mac.

Alexander-Taran:
@huochunpeng works for me and my weird setup.

huochunpeng:
@EisenbergEffect safe to merge. This cleans up sourceMap source map in browser display.

PR #1088, opened 2019-04-18
Keep webpack config for dev server working

PisaJan:
I have spent some time before I realized that run script is "overwriting" my dev server configuration :-)

### DIFF
--- a/skeleton/webpack/aurelia_project/tasks/run.ext
+++ b/skeleton/webpack/aurelia_project/tasks/run.ext
@@ -29,7 +29,7 @@ function runWebpack(done) {
     stats: {
       colors: require('supports-color')
     },
-    https: config.devServer.https
+    ...config.devServer
   }/* @if feat.typescript ** as any/* @endif */;
 
   // Add the webpack-dev-server client to the webpack entry point


huochunpeng:
LGTM.

@fkleuver can you review this? You were the one brought in https from webpack config. #865

3cp:
@Sayan751 could you help to review this one?

Sayan751:
@3cp 

First thing is that this change should work. Therefore, if you want to merge this change, please go ahead.

Here come the "but" 😛 I think that the application of the devserver options can be cleaned up a little bit. Though I am not a heavy user of the CLI, it seems that the options can come from 3 places: webpack config, aurelia config, and CLI options (please correct me if there are more or lesser options). I think the precedence of the these options should be as follows.

0. Recommended defaults by Aurelia team + options from aurelia.json
1. Overridden by user defined webpack config
2. Overridden by CLI options.

I am suggesting something like below. 

```typescript
// omitted "host: 'localhost'" as it is a redundant reassignment of default value "localhost" (https://webpack.js.org/configuration/dev-server/#devserverhost)
const defaults = {
  publicPath: config.output.publicPath,
  filename: config.output.filename,
  hot: project.platform.hmr,
  port: project.platform.port,
  contentBase: config.output.path,
  historyApiFallback: true,
  open: project.platform.open,
  stats: { colors: require('supports-color') },
};

const cliOptions = (() => {
  let tempValue;
  const opts: any = {
    ...CLIOptions.hasFlag('hmr') && { hot: true },
    ...CLIOptions.hasFlag('open') && { open: true },
    ...(tempValue = CLIOptions.getFlagValue('port')) && { port: tempValue },
    ...(tempValue = CLIOptions.getFlagValue('some_other_value')) && { some_other_prop: tempValue },
  };
  return opts;
})();

opts = {
  ...defaults,
  ...config.devServer,
  ...cliOptions,
};

if (opts.hot) {
  // you get the idea
}
```

What do you think about this proposal? It is your call if you want to push the changes in this PR or not, in case it is a "aye".

3cp:
@EisenbergEffect let's merge this fix.

@Sayan751 love your proposal. The community has been talking about our overly complex webpack setup for a while. We are very open for simplification/enhancement. If you want to turn your idea into a PR, we are happy to either merge it before v1 or after v1 release. I myself is not a webpack user, so I rarely touch webpack part of aurelia-cli.

We knew advanced webpack users didn't use cli for their aurelia+webpack app. Our webpack setup forces users to use aurelia-cli (+gulp) to run tasks which is not common to webpack users. For long term, we definitely want to push a clean (aurelia-cli-less) webpack setup for vNext.

Sayan751:
@3cp I would love to work on such change. Just let me look around a little. I will get back to you.

EisenbergEffect:
Ok, I'll merge this for now. @Sayan751 We'd absolutely love to have you help us on making Webpack better. Please feel free to submit a PR based on your ideas above and @3cp will work with you to review and get it in. Thanks everyone!

Sayan751:
@EisenbergEffect Working on a basic demo. Will open an issue with the proposal shortly  :)

Issue #1083, opened 2019-04-05
SASS does not get extracted to CSS for production


arjendeblok:
**I'm submitting a bug report**

* **Library Version:**
1.0.0-beta.15

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
10.13..0

* **NPM Version:**
6.4.1

* **Browser:**
all

* **Language:**
TypeScript

* **Loader/bundler:**
Webpack 

* **Current behavior:**
When running a build for production, CSS is extracted to .CSS files, but SASS files do not get extracted.

* **What is the expected behavior?**
That SASS and CSS are both extracted to .CSS files.

* **What is the motivation / use case for changing the behavior?**
For production all CSS should be in seperate files.

CollinHerber:
This is probably the biggest annoyance with the current beta build of the CLI. Please fix this!

huochunpeng:
If someone can clean up the webpack config, we would love to merge it.

For requirejs build, I am not sure whether we need to seperate css files. Currently css files are bundled in as a wrapped text module. We could use seperate css files for requirejs to do dynamic loading at runtime, requirejs/systemjs supports that. But it creates additional burden for app deployer to maintain correct requirejs baseUrl setting.

Another thing devs can do, is to totally avoid `import "some.css"` and `<require from="some.css"></require>`. Use traditional `<style>` tag to load css, it means you compile all css asserts outside of webpack config (or cli-bundler). The traditional way actually simplifies the toolchain setup, what you lose is per component css file and style on demand.

arjendeblok:
I will have a look and make a pull request.

The last thing that you mention is not a solution for SASS files. Or if you use a third party component library that uses SASS.

arjendeblok:
Problem solved with merge.

PR #1089, opened 2019-04-23
Implemented SASS extraction into CSS files.

arjendeblok:
Solved issue #1083.

### DIFF
--- a/skeleton/webpack/webpack.config.js
+++ b/skeleton/webpack/webpack.config.js
@@ -39,6 +39,20 @@ const cssRules = [
   // @endif
 ];
 
+// @if feat.sass
+const sassRules = [
+  {
+    loader: "css-loader"
+  },
+  {
+     loader: "sass-loader", 
+     options: {
+       includePaths: ["node_modules"]
+     }
+  }
+];
+// @endif
+
 module.exports = ({ production, server, extractCss, coverage, analyze, karma } = {}) => ({
   resolve: {
     // @if feat.typescript
@@ -245,7 +259,10 @@ module.exports = ({ production, server, extractCss, coverage, analyze, karma } =
       // @if feat.sass
       {
         test: /\.scss$/,
-        use: ['style-loader', 'css-loader', 'sass-loader'],
+        use: extractCss ? [{
+          loader: MiniCssExtractPlugin.loader
+        }, ...sassRules
+        ]: ['style-loader', ...sassRules],
         issuer: /\.[tj]s$/i
       },
       {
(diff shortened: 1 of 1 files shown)


skeleton/webpack/webpack.config.js
```diff
@@ -245,7 +259,10 @@ module.exports = ({ production, server, extractCss, coverage, analyze, karma } =
       // @if feat.sass
       {
         test: /\.scss$/,
-        use: ['style-loader', 'css-loader', 'sass-loader'],
+        use: extractCss ? [{
+          loader: MiniCssExtractPlugin.loader
+        }, ...sassRules
+        ]: ['style-loader', ...sassRules],
```
huochunpeng: Your newly created sassRules bypassed postcss config.

It looks like it could be
```
}, 'css-loader', 'sass-loader'
]: ['style-loader', ...cssRules, 'sass-loader']
```


Or
```
}, 'css-loader', ...sassRules
]: ['style-loader', ...cssRules, ...sassRules]
```
With
```
const sassRules = [
	  {
	     loader: "sass-loader", 
	     options: {
	       includePaths: ["node_modules"]
	     }
	  }
	];
```

Also, to be complete, need to do the same thing for less and stylus config.
I am no master of webpack, I am not sure why postcss is skipped when extractCss is on, I mean in our existing config.

huochunpeng:
@jods4 @fkleuver can one of you review this?

skeleton/webpack/webpack.config.js
arjendeblok: I did not foresee that someone wanted to use SASS and PostCSS together.
Normally someone uses one or the other. But I might be wrong of course.

huochunpeng: postcss is not a preprocessor, it's a post-processor, it can be used with css or the css result compiled from less/sass/stylus.

For example the autoprefixer (plugin for postcss) is a useful step for any css setup. Although I am not sure if webpack does some browser prefix automatically.

arjendeblok: It is useful for minification (via nanocss). So I updated the files.

huochunpeng:
Thx, could you update less and stylus config too?

huochunpeng:
I tested with sass/less/stylus, all worked for me.
@EisenbergEffect LGTM, but better to help some webpack guru to review.

EisenbergEffect:
@huochunpeng do we have someone from the community who reached out that could also review?

huochunpeng:
Yes I asked @Sayan751 for a review too.

Sayan751:
The rules are looking pretty straightforward to me.

huochunpeng:
@Sayan751 thanks.
@EisenbergEffect let's get this fix in.

PR #1094, opened 2019-04-26
fix(CLI-Bundler, Aliases): improve alias support

### LINKED ISSUE #1093: CLI Bundler alias/paths matched incorrectly (Plugin development)

* **Library Version:**
"aurelia-cli:  `^1.0.0-beta.15`
* **Operating System:**
Windows [10]

* **Node Version:**
v10.15.0

* **NPM Version:**
6.9.0

* **Browser:**
Not browser related

* **Language:**
TS but assume it's an issue for all

* **Loader/bundler:**
CLI Bundler

**Current behavior:**
When the bundler tries to add aliases it looks at the module id, for example `../src/elements/hello-world` and tries to match it with a configured path.
One of the following in a default configuration:
```json
{
  "paths": {
    "root": "dev-app",
    "resources": "../src",
    "elements": "../src/elements",
    "attributes": "../src/attributes",
    "valueConverters": "../src/value-converters",
    "bindingBehaviors": "../src/binding-behaviors"
  }
}
```
It does this by going over this list top down and will only match 1. For the previously mentioned module id it will match with resources and this will result in failure when running the app.

Logic responsible for this behavior:

```js
// lib/build/bundled-source.js:319
function possibleShortcut(moduleId, paths) {
  const _moduleId = fromDotDot(moduleId);
  for (let i = 0, keys = Object.keys(paths); i < keys.length; i++) {
    let key = keys[i];
    let target = paths[key];
    if (key === 'root') continue;
    if (key === target) continue;

    if (_moduleId.startsWith(target + '/')) {
      return {
        fromId: toDotDot(key + _moduleId.slice(target.length)),
        toId: toDotDot(moduleId)
      };
    }
  }
}
```

I would not expect order to influence alias path matching, neither would I expect duplicate entries with different paths to be an issue.
There is probably a good reason for current approach but I would expect imports to be converted to the actual paths instead of creating a creating a new module that returns an existing one. So search for key and replace by target (path).

~~A low impact solution might be to search for the best match based on amount of character matched (I think this could be accomplished by sorting by length before looping)~~
Edit: Bad idea this could break existing setups

* **Steps to reproduce**

**Given**: You have the aurelia cli installed
**And**: run `au new --plugin`
**And**: follow instructions, choose all examples at the final step
**And**: add `import { HelloWorld } from 'elements/hello-world';` to `dev-app/app.ts` 
**And** add `console.log(HelloWorld)` to a constructor
**And**: run `au run`
**When**: open `localhost:9000` in browser
**Then**: console will show error

>GET http://localhost:9000/src/elements/hello-world.js net::ERR_ABORTED 404 (Not Found)

**Expected**: 
Hello world class should load and be logged in the console.

I would like to utilize paths to keep my code clean and maintainable.

*Work around*
Reorder paths, deepest first, shortest last. This has a drawback, it will force you to use the correct (deepest) alias.

michaelw85:
Define multiple aliases when they can be applied for a path.

Closes #1093

### DIFF
--- a/lib/build/bundled-source.js
+++ b/lib/build/bundled-source.js
@@ -5,6 +5,7 @@ const esTransform = require('./amodro-trace/read/es');
 const allWriteTransforms = require('./amodro-trace/write/all');
 const Utils = require('./utils');
 const logger = require('aurelia-logging').getLogger('BundledSource');
+const { getAliases, toDotDot } = require('./module-id-processor');
 
 exports.BundledSource = class {
   constructor(bundler, file) {
@@ -108,10 +109,9 @@ exports.BundledSource = class {
     let moduleId = this.moduleId;
     let modulePath = this.path;
 
-    let shortcut = possibleShortcut(moduleId, loaderConfig.paths);
-    if (shortcut) {
-      this.bundler.configTargetBundle.addAlias(shortcut.fromId, shortcut.toId);
-    }
+    getAliases(moduleId, loaderConfig.paths).forEach(alias => {
+      this.bundler.configTargetBundle.addAlias(alias.fromId, alias.toId);
+    });
 
     logger.debug(`Tracing ${moduleId}`);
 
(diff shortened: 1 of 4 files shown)


michaelw85:
@huochunpeng Here you go. I moved the function into a separate file and changed it to return multiple matches.

huochunpeng:
Awesome, LGTM.

michaelw85:
@EisenbergEffect Is it possible to get this merged?

EisenbergEffect:
Yes sir! Thanks for the contribution!

PR #1095, opened 2019-04-30
test: turn on travis tests on nodejs v12

huochunpeng:
Also cleaned up deps to avoid npm vulnerability warning


huochunpeng:
@EisenbergEffect safe to merge.

PR #1091, opened 2019-04-24
fix(webpack): fix various TypeScript typing error

huochunpeng:

PR #1100, opened 2019-05-16
fix(ts + jest + cypress): Add missing type

### LINKED ISSUE #1099: [BUG] Typing issues Jest + Cypress

* **Language:**
TypeScript

* **Loader/bundler:**
Webpack

**Current behavior:**
Newly created project with Aurelia CLI shows errors when running the app. Due to compile errors HMR fails to work and Cypress is not working properly.

**Steps to reproduce**
Create new custom app with the following settings (jest & cypress combination is critical):
`au new sample-app`

1. webpack 
2. http2 
3. web 
4. ts 
5. maximum minification 
6. sass 
7. postcss typical 
8. **jest** 
9. **cypress** 
10. vscode 
11. navigation app examples

NPM and Yarn yield the same result.

`au run`

No errors

**How to fix**
- add @types/mocha
- ~~add type reference in jest-pretest~~ seems to work without

michaelw85:
Add @types/mocha to fix compilation when using ts + jest + cypress.

Closes #1099

### DIFF
--- a/skeleton/cypress/package.json
+++ b/skeleton/cypress/package.json
@@ -1,6 +1,7 @@
 {
   "devDependencies": {
     // @if feat.typescript
+    "@types/mocha": "",
     "@cypress/webpack-preprocessor": "",
     "webpack": "",
     "ts-loader": "",


michaelw85:
@3cp Very small fix to get a new and clean project running properly.
In my own setup I also had a reference type set in jest pretest but it seems to be working fine without, maybe you can give it a try on your machine.

@EisenbergEffect

3cp:
This fix maybe not needed. The issue does not appear with latest aurelia-cli master, it's probably fixed by #1091 when we changed ts-loader options.

Also the mocha thing doesn't feel right, because we didn't use mocha. It's a pity that there always are some annoying typing def issues when using TS with any lib that pollutes global namespace.

Try latest master branch with `npx aurelia/cli new jc2 -u -s http2,typescript,htmlmin-max,sass,postcss-typical,jest,cypress,vscode,scaffold-navigation`.

nym:
I hit the typing def issues too. Definitely made me scratch my head, as I had already gotten cypress working in another project.

michaelw85:
@3cp I will double check but I was using the latest cli as far as I am aware and was still having the issue.

Mocha is required for Cypress.

Sidenote the change you refer to sets typeroots property for ts to @types, this is the default value and should not have to be set explicitly.

3cp:
That fix was not released.

3cp:
typeRoots was not for this issue (it's a limited fix only for condition with protractor), as a fact, if you create your app (with cypress) again with latest cli master, it would not set typeRoots.

michaelw85:
Ok not sure what would be responsible for the fix than. Currently it's broken, adding mocha types for Cypress is required and seem to be missing so independent of any other change I think this one is needed.

3cp:
I was talking about fixed by this change `{ test: /\.ts$/, loader: "ts-loader", options: { reportFiles: [ srcDir+'/**/*.ts'] }, include: srcDir }`.

michaelw85:
Ok, not completely sure why this resolves the issue. I will give it a try to verify.

michaelw85:
@3cp @nym I just tested and it seems like my fix is not required after all. When reverting the ts loader rule it fails again so:
✔ #1091 Resolves this issue

PR #1101, opened 2019-05-17
feat(cli): Warn the user for providing multiple arguments with aurelia new command

jamesgeorge007:
As per now `aurelia new <project-name> arg` goes ahead without any sort of warnings being thrown. 
Made it such that the user is warned if he does so.

This is how it looks like:-
Screenshot from 2019-05-17 08-30-26

### DIFF
--- a/bin/aurelia-cli.js
+++ b/bin/aurelia-cli.js
@@ -25,7 +25,7 @@ resolve('aurelia-cli', {
   let cli;
 
   if (commandName === 'new' || error) {
-    const args = commandArgs.filter(item => !item.includes('-'));
+    const args = commandArgs.filter(item => !item.charAt(0).includes('-'));
     if (args.length > 1) {
     	console.log();
     	console.log(' Kindly provide only one argument as the <project-name>');
--- a/bin/aurelia-cli.js
+++ b/bin/aurelia-cli.js
@@ -25,12 +25,6 @@ resolve('aurelia-cli', {
   let cli;
 
   if (commandName === 'new' || error) {
-    const args = commandArgs.filter(item => !item.charAt(0).includes('-'));
-    if (args.length > 1) {
-    	console.log();
-    	console.log(' Kindly provide only one argument as the <project-name>');
-    	process.exit(1);
-    }	
     cli = new (require('../lib/index').CLI);
     cli.options.runningGlobally = true;
   } else {
--- a/bin/aurelia-cli.js
+++ b/bin/aurelia-cli.js
@@ -25,6 +25,12 @@ resolve('aurelia-cli', {
   let cli;
 
   if (commandName === 'new' || error) {
+    const args = commandArgs.filter(item => !item.includes('-'));
+    if (args.length > 1) {
+    	console.log();
+    	console.log(' Kindly provide only one argument as the <project-name>');
+    	process.exit(1);
+    }	
     cli = new (require('../lib/index').CLI);
     cli.options.runningGlobally = true;
   } else {
--- a/lib/commands/new/command.js
+++ b/lib/commands/new/command.js
@@ -29,6 +29,14 @@ module.exports = class {
     // -u
     this.unattended = this.options.hasFlag('unattended');
 
+    // Validate arguments provided as the <project-name>
+    const userArgs = this.options.args.filter(item => !item.startsWith('-'));
+    if (userArgs.length > 1) {
+    	console.log();
+    	console.log(' Kindly provide only one argument as the <project-name>');
+    	process.exit(1);
+    }
+
     // Preselect features
     // In unattended mode, these overwrite default-esnext selections.
     // In interactive mode, some questions will be skipped because they are
@@ -228,4 +236,3 @@ async function minFeatures(features, plugin) {
   }
   return mustHave;
 }
-
--- a/lib/commands/new/command.js
+++ b/lib/commands/new/command.js
@@ -29,14 +29,6 @@ module.exports = class {
     // -u
     this.unattended = this.options.hasFlag('unattended');
 
-    // Validate arguments provided as the <project-name>
-    const userArgs = this.options.args.filter(item => !item.startsWith('-'));
-    if (userArgs.length > 1) {
-    	console.log();
-    	console.log(' Kindly provide only one argument as the <project-name>');
-    	process.exit(1);
-    }
-
     // Preselect features
     // In unattended mode, these overwrite default-esnext selections.
     // In interactive mode, some questions will be skipped because they are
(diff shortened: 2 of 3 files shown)


jamesgeorge007:
@EisenbergEffect @3cp Thoughts :thinking:

3cp:
I don't understand what's the possible intend of the user if she/he tries to run `au new foo bar`?

jamesgeorge007:
@3cp What intended is only the `<project-name>` to be provided as an argument right?

3cp:
`au new` can run without any args, the <project-name> is optional. When it's absent, the UI to prompt to ask for project name.

jamesgeorge007:
Not validating the the user input may make the user think that it accepts all sorts of crazy inputs which doesn't make sense at all :thinking:

3cp:
But your message is also miss-leading, it's better to print out the help of au new command (you can see the full help for `au new` with `au --help`).

3cp:
For fully validating user input, any unknown option (`--unknown`) should lead to print out help. Not just validate one arg.

3cp:
I appreciate the thoughts on guiding user. I know I have been picky, but your current implementation didn't draw the full picture.

3cp:
> I don't understand what's the possible intend of the user if she/he tries to run au new foo bar?

I was asking what's the end user's intention trying to run `au new foo bar`. Not asking Aurelia team's intention.
If end user has no intention to run `au new foo bar`, why you bother to guard end users from a mistake they would likely never make?

fkleuver:
@jamesgeorge007 @3cp I think it's fair game to perform input validation, but then it needs to be feature-complete (covering all arguments instead of just this one) and properly tested. Especially tests are important - we can't afford to have new features or logic introduced at this point without being backed by tests.

Also, tests would make it easier to discuss what the PR accomplishes in a more meaningful manner.

For example:
- What happens if you do `aurelia new webapp foo-bar`?
- Or `aurelia new web-app foo`?
- Or `aurelia new foo-bar1 foo-bar2 foo-bar3`?
- Or `aurelia new "foo bar"`?

There are all kinds of interesting edge cases here. There should probably be a generic validator that can be reused for all commands.

jamesgeorge007:
@fkleuver Currently all these edge cases are been taken care of:-

`aurelia new webapp foo-bar`
`aurelia new web-app foo`
`aurelia new foo-bar1 foo-bar2 foo-bar3`

All these results in the respective message as shown above.

jamesgeorge007:
Still `aurelia new "foo bar"` doesn't seem to be covered. What would you suggest in-order to include the same?

jamesgeorge007:
@3cp @fkleuver Any sort of guidance for writing unit tests will be appreciated :clap:

fkleuver:
@jamesgeorge007 The file you've modified is probably not the ideal place to be doing this sort of validation in the first place.

Interaction related logic, help messages, etc, reside mostly in the commands. For `new` that would be here: https://github.com/aurelia/cli/blob/master/lib/commands/new/command.js

There are already existing unit tests in place for some of those files, so moving it there would not only keep things in the right place but also make testing easier, as you can base new test off of existing ones.

jamesgeorge007:
@fkleuver Thoughts :thinking:

fkleuver:
Around here could be a good place: https://github.com/aurelia/cli/blob/master/lib/commands/new/command.js#L59

Then I'd add a `commands/new.js` under the `specs` folder here: https://github.com/aurelia/cli/tree/master/spec/lib/commands

The general gist could be something among the lines of:

```ts
describe('The new command', () => {
  const Command = require('../../../../lib/commands/new/command');

  const specs = [
    {
      input: ['foo', 'bar'],
      isValid: false,
    },
    {
      input: ['foo'],
      isValid: true,
    },
    {
      input: ['foo-bar'],
      isValid: true,
    },
    {
      input: ['foo', 'foo-bar'],
      isValid: false,
    },
    {
      input: ['foo', '-foo-bar'],
      isValid: false,
    },
  ];

  for (const spec of specs) {
    it(`validates ${spec.input.join(' ')} as isValid=${spec.isValid}`, () => {
      const stubUI = {
        displayLogo() { },
      };
      const stubOptions = {
        hasFlag() { return false; },
        getFlagValue() { return ''; },
      };

      let message;
      const originalLog = console.log;
      const logSpy = (msg) => {
        message = msg;
        console.log(msg);
      };
      console.log = logSpy;

      const sut = new Command(stubUI, stubOptions);

      sut.execute(['foo', 'bar']);

      console.log = originalLog;
    
      if (spec.isValid) {
        expect(message).toBe('...');
      } else {
        expect(message).toBe('Kindly provide..');
      }
    });
  }
```

And a couple more edge cases perhaps. Just to make sure it's all proper and prevent regressions.

jamesgeorge007:
@fkleuver How are we gonna go ahead with the unit tests :thinking: Just included a `new.spec.js` within the `spec` directory.

fkleuver:
@jamesgeorge007 Obviously they need to pass, which they currently don't :) You can see this at the status checks.

And they need to cover the various scenarios mentioned. Rule out false positives as well.

I should add that I have merely been making sure a certain quality bar (put things in the right place, add tests etc) is met.

Like @3cp, I am not entirely sure what problem you are attempting to solve. You took the initiative which is of course appreciated, but I cannot answer the question of where to go with this. That is yours to answer :)

Perhaps we can go back a step to @3cp's question:

> I was asking what's the end user's intention trying to run au new foo bar. Not asking Aurelia team's intention.
If end user has no intention to run au new foo bar, why you bother to guard end users from a mistake they would likely never make?

What would the user's intention be / what would they be expecting, that warrants guarding them from this situation with a message? I don't believe anyone has ever reported such issue.

jamesgeorge007:
@fkleuver Most of the modern day scaffolding tools comes shipped with such sort of  validations. For instance, `@vue/cli`.

fkleuver:
@jamesgeorge007 


If this was a generic piece of validation logic you were introducing architecturally in the right place, it would be worth considering because it could potentially be reused and such. But such a very specific validation like this is patchy, and patchy is only worth it when it fixes a real (preferably even urgent) problem. I believe this is neither.

I'm sorry, but a hypothetical "may make the user think that it accepts all sorts of crazy inputs" for me does not give it a benefit that outweighs the cost of having additional code to maintain. We prefer to focus on real world concerns and protracting users against mistakes they might *actually* make.

Side note, I see you made a nearly identical commit here: https://github.com/FoalTS/foal/commit/8221b9e9067e587decc8a523c23d496c94a06127
It is commendable that you're going around repos trying to help people, I think it's great that you're doing that and please keep doing so. But as for this particular commit for this particular repo, I think it's better not to push through in its current state.

@3cp is there anything you wish to add or address?

jamesgeorge007:
@fkleuver I agree with you :+1:  What about validating the following input:- 

`aurelia new "foo bar"`

How about using packages like [validate-npm-package-name](https://www.npmjs.com/package/validate-npm-package-name)  to validate the respective case :thinking:

PR #1103, opened 2019-05-18
fix(skeleton): fix missing .gitignore file in aurelia-cli npm package

### LINKED ISSUE #1102: .gitignore missing in generated app?
Just generated all-defaults ESNext app, with aurelia-cli v1.0.0-beta.15 on node v10.15.3 and see there is no .gitignore in the new generated app root folder.

Is it a bug because the default .gitignore would be convenient?

3cp:
When building npm package, `npm pack` ignored skeleton/common/.gitignore file because it by default ignores any .gitignore file. Rename .gitignore to .gitignore__append-if-exists is the cheapest way to bypass npm's default behaviour. The alternative fix is to use "files" field in package.json, but that's too verbose.

closes #1102

### DIFF
--- a/skeleton/common/.gitignore__append-if-exists
+++ b/skeleton/common/.gitignore__append-if-exists
@@ -1,3 +1,4 @@
+
 # You may want to customise this file depending on your Operating System
 # and the editor that you use.
 #


3cp:
@EisenbergEffect safe to merge.

Greatly thank @bsrdjan for noticing this weird bug!

PR #1104, opened 2019-05-20
feat(plugin, build-plugin): Add option to process aliases in templates

michaelw85:

* **Library Version:**
Latest CLI

**Current behavior:**
When running build-plugin require paths with aliases in html files will end up in dist. When consuming the plugin this will break since the consuming app does not know how to process the aliases.

Aliases in html files should be replaced with relative paths to make the plugin usable.

Using aliases in the CLI is supported for the dev-app and this leads to the assumption that it would also work in the plugin code itself. Using aliases makes it easier to use/read and maintain the code in my opinion.

* **Implemented change**
Added conditional flag for au build-plugin to replace aliases in template require tags with relative paths.
This feature will not be used by default so it will not affect any existing plugins.

* **How it works**
1.) Looks at the paths configured in aurelia.json to create aliases relative to `src` since the current setup uses the paths relative to the `dev-app`.
2.) Uses a regex to find `require from="..."` tags in html
3.) If the require path starts with an alias it will replace it with a relative path

* **Example**
_src/my-component/comp.js_
`<require from="elements/file-in-root.html">` --> `<require from="../file-in-root.html">`

* **Limitations**
This only add support for parsing the template. For typescript a different implementation is needed. A new feature request could be created for this.

### DIFF
--- a/skeleton/cli-bundler/aurelia_project/tasks/process-markup.ext
+++ b/skeleton/cli-bundler/aurelia_project/tasks/process-markup.ext
@@ -17,6 +17,9 @@ import * as plumber from 'gulp-plumber';
 import * as notify from 'gulp-notify';
 // @endif
 // @endif
+// @if feat.plugin
+import { processMarkupAliases } from './process-markup-aliases';
+// @endif
 
 export default function processMarkup() {
   return gulp.src(project.markupProcessor.source, {sourcemaps: true, since: gulp.lastRun(processMarkup)})
@@ -41,26 +44,34 @@ export default function processMarkup() {
 }
 
 // @if feat.plugin
-export function pluginMarkup(dest) {
+export function pluginMarkup(dest, runProcessMarkupAliases = false, verbose = false) {
   return function processPluginMarkup() {
-    return gulp.src(project.plugin.source.html)
-      // @if feat['htmlmin-min'] || feat['htmlmin-max']
-      .pipe(htmlmin({
-          // @if feat['htmlmin-max']
-          collapseInlineTagWhitespace: true,
-          collapseBooleanAttributes: true,
-          removeAttributeQuotes: true,
-          removeScriptTypeAttributes: true,
-          removeStyleLinkTypeAttributes: true,
-          // @endif
-          removeComments: true,
-          collapseWhitespace: true,
-          minifyCSS: true,
-          minifyJS: true,
-          ignoreCustomFragments: [/\${.*?}/g] // ignore interpolation expressions
-      }))
-      // @endif
-      .pipe(gulp.dest(dest));
+    let stream = gulp.src(project.plugin.source.html);
+
+        if (runProcessMarkupAliases) {
+            stream = stream.pipe(processMarkupAliases(verbose));
+        }
+
+        stream = stream
+            // @if feat['htmlmin-min'] || feat['htmlmin-max']
+            .pipe(
+                htmlmin({
+                    collapseInlineTagWhitespace: true,
+                    collapseBooleanAttributes: true,
+                    removeAttributeQuotes: true,
+                    removeScriptTypeAttributes: true,
+                    removeStyleLinkTypeAttributes: true,
+                    removeComments: true,
+                    collapseWhitespace: true,
+                    minifyCSS: true,
+                    minifyJS: true,
+                    ignoreCustomFragments: [/\${.*?}/g], // ignore interpolation expressions
+                })
+            )
+            // @endif
+            .pipe(gulp.dest(dest));
+
+        return stream;
   };
 }
 // @endif
--- a/skeleton/cli-bundler/package.json
+++ b/skeleton/cli-bundler/package.json
@@ -67,5 +67,9 @@
     "gulp-rename": "",
     "gulp-notify": "",
     "gulp-watch": "",
+    
+    // @if feat.plugin
+    "gulp-replace": "",
+    // @endif
   }
 }
--- a/skeleton/plugin/aurelia_project/tasks/build-plugin.ext
+++ b/skeleton/plugin/aurelia_project/tasks/build-plugin.ext
@@ -10,6 +10,7 @@ import {pluginMarkup} from './process-markup';
 import {pluginCSS} from './process-css';
 import {pluginJson} from './process-json';
 import {buildPluginJavaScript} from './transpile';
+import { CLIOptions } from 'aurelia-cli';
 
 function clean() {
   return del('dist');
@@ -19,13 +20,13 @@ export default gulp.series(
   clean,
   gulp.parallel(
     // package.json "module" field pointing to dist/native-modules/index.js
-    pluginMarkup('dist/native-modules'),
+    pluginMarkup('dist/native-modules', CLIOptions.hasFlag('process-markup-aliases'), CLIOptions.hasFlag('verbose')),
     pluginCSS('dist/native-modules'),
     pluginJson('dist/native-modules'),
     buildPluginJavaScript('dist/native-modules', 'es2015'),
 
     // package.json "main" field pointing to dist/native-modules/index.js
-    pluginMarkup('dist/commonjs'),
+    pluginMarkup('dist/commonjs', CLIOptions.hasFlag('process-markup-aliases'), CLIOptions.hasFlag('verbose')),
     pluginCSS('dist/commonjs'),
     pluginJson('dist/commonjs'),
     buildPluginJavaScript('dist/commonjs', 'commonjs'),
--- a/skeleton/plugin/aurelia_project/tasks/build-plugin.json
+++ b/skeleton/plugin/aurelia_project/tasks/build-plugin.json
@@ -1,0 +1,16 @@
+{
+  "name": "build-plugin",
+  "description": "Build plugin",
+  "flags": [
+    {
+      "name": "process-markup-aliases",
+      "description": "Replace aliases in template require tags to relative paths for distribution",
+      "type": "boolean"
+    },
+    {
+      "name": "verbose",
+      "description": "Increase log output during build",
+      "type": "boolean"
+    }
+  ]
+}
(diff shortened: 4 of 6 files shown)


michaelw85:
I also have a unit test which I created in my own project with jest, to be able to run this test in the CLI a transpiler is required. This could be achieved by completing: https://github.com/aurelia/cli/pull/1067

michaelw85:
@EisenbergEffect  @3cp 
Not sure if anyone is waiting for this feature but since I require it in my own project I wanted to share.
It's added in a opt-in manner and therefore non destructive on any existing projects.

3cp:
Sorry, I cannot speak for other aurelia team members, but technically I don’t understand why go all these trouble to support a feature not really needed. Since everything inside src folder can load each other with relative path which is natural and intuitive.

Just enhance the doc to say refer anything inside src folder with relative path, isn’t that better?

After all, a plugin project is a npm package after all. I didn’t see a npm package not loading inner files with relative path.

michaelw85:
I do not think Aurelia should dictate how to write your code, it should help and offer tooling. Aurelia can offer best practices but restricting in choice does not feel right to me.  Saying you don't see the added benefit, does not mean relative paths is the way to go or preferred by others. In your line of thought I could also dismiss TS as an option since "just write js" is a technical viable option.

In my project I quickly end up with relative paths looking like `../../../../value-converters/converter` which is a pain to read and maintain. My primary reason to never use relative paths is moving files around with dependencies will give you headaches. This is largely resolved by using aliases.

Apart from my motivations if anyone else has his or her own motivation to use aliases with this feature you could decide to use them in templates. Because I assume most will not this feature was implemented in such a way it will not affect anyone unless you explicitly start using it.

fkleuver:
> In my project I quickly end up with relative paths looking like ../../../../value-converters/converter which is a pain to read and maintain. My primary reason to never use relative paths is moving files around with dependencies will give you headaches. This is largely resolved by using aliases.

I ran into the same problem in a large project some time ago. In fact I've also been running into this in the latest big refactor of vNext. Even a relatively small structural reorganization can easily take an hour where it could be a 5 minute chore with absolute paths.

Aliases don't play nicely with the vast majority of tooling out there. Typings often don't get resolved properly, sometimes TypeScript understands it but webpack's ts-loader doesn't, other times it's the other way around, sometimes lerna decides to hide a nice buggy easter-egg. There is always something misbehaving. I gave up on absolute paths after several serious attempts and wasting serious time. I accepted that this pain of working with relative paths is just how it's going to be.

If we can try to make an effort to actually make absolute paths / aliases work, I'm game. It can be a huge time saver in very large projects.

michaelw85:
@fkleuver Thanks for your reply. In our project I now completely resolved the issue (at least I think I did. This PR was part of getting it working). For typescript I'm using https://github.com/joonhocho/tscpaths to replace the aliases to relative paths.

3cp:
Let's put my unpleasant opinion aside. Let's talk about implementation.

You implementation only deals with `<require from="shortcut/foo"></require>` in html file, which's enough for your use case, that's good.

But you need to realise to fully support alias in html as well as js, you need to at least adding support of `import {Foo} from 'shortcut/foo';` and `PLATFORM.module('shortcut/foo');`.

Even your html implementation is not enough, you need to add support of `<compose view-model="shortcut/foo" (optional view) view="shortcut/foo.html"></compose>`.

Let's say you want to enhance your implementation to properly support alias in js/ts and html files. The simple fact is, any code modification without proper AST analysis just wouldn't work. @fkleuver wouldn't argue about this assessment.

Take a simple example which a `<require>` is totally missed even you only want to deal with view template.

```js
// you missed this one
const view = "<template><require from=\"shortcut/foo\"></require></template>"

@inlineView(view)
export class VM {
  tutorial = `
Don't touch this one
<require from=\"shortcut/foo\"></require>
`;
}
```

You see, to properly support the above, you need AST analysis, plus [variable scope analysis](https://github.com/estools/escope). It can be done, but will make your LOC sky rocking. **Regex or split aren't enough for proper code modification.**

Furthermore, although we recommend all plugin authors use `PLATFORM.moduleName` in order for their plugin to support webpack, some authors wouldn't use the helper when their private plugins are only used in apps using cli-bundler (yes, cli-bundler supports all aurelia convention without the help of `PLATFORM.moduleName`), now you have to deal with dependency modification in wide range of Aurelia convention. Few examples:
```js
c.globalResources(['shortcut/foo', 'shortcut/bar']);
@useView('shortcut/foo.html', ['shortcut/foo.css']);
@noView(['shortcut/foo.css']);
```

One more trouble, to be a good citizen in gulp pipeline, any tool can mutate js/ts code should also update sourceMap, which itself is a tricky task.

You didn't realise how troublesome to implement this feature.

fkleuver:
I must admit I didn't even look at the implementation when I wrote my comment - I was merely defending the use case as being a real, important need that could be worth the trouble.

After taking a look at the implementation, I do fully agree with @3cp's assessment. `gulp-replace` is not going to cut it for this to be a core features. A proper AST walk is a must for this kind of thing, as is at the very least several dozens of tests.

However, I don't think we have to account for non-`PLATFORM.moduleName` scenarios in js/ts. The analysis necessary to overcome this would require accurately detecting what might be a reference or an import, which is not possible to do 100% reliably. vCurrent is not static analysis friendly in this regard -  this is why `PLATFORM.moduleName` exists in the first place.

3cp:
> The analysis necessary to overcome this would require accurately detecting what might be a reference or an import, which is not possible to do 100% reliably.

That's so true, cli-bundler is doing best effort to support it. Because forcing user to use `PLATFORM.moduleName` is an unacceptable breaking change from old amodro cli-bundler to the new auto-tracing cli-bundler.
The Aurelia vCurrent's loose-coupling nature, is very flexible for app to build runtime features, but very difficult for bundler to reason about.

fkleuver:
> The Aurelia vCurrent's loose-coupling nature

It's requirejs / loader architecture that's the culprit here - dependencies are declared as arbitrary strings and you can only tell for sure that something is a dependency when fully resolved at runtime after requirejs performed the `require` call.

v2 is even more loosely coupled and yet doesn't have this analysis problem, because you can tell that something is a dependency by tracing the identifier back to its declaration. All dependencies at some point either resolve to a `DI.createInterface` call or to a concrete class. If it's an interface, DI resolves it at runtime.

Any other kind of import is always a "real" import with native ES syntax that can easily be identified as such, and thus manipulated without having to perform sorcery.

The vCurrent equivalent of what becomes `DI.createInterface` in v2, is `PLATFORM.moduleName`

> forcing user to use PLATFORM.moduleName is an unacceptable

We have no choice but to ask this of users if they want to have this very specific feature of using absolute paths instead of relative, just like we have been forced to ask this of users who want to use webpack.

michaelw85:
I feel like I poked the bear with this one 😅

Ok so my implementation is overly simplistic and does not cover all cases. I have to admit I underestimated this a bit. To be fair on the js/ts part I already added the following to the description:

> Limitations
This only add support for parsing the template. For typescript a different implementation is needed. A new feature request could be created for this.

To continue along these lines, seeing this is a big feature as a whole you could probably categorize this as an epic in agile terms. I would like to keep things small, no one likes a massive PR. They go stale very quickly and end up having multiple conflicts. No offence but things aren't going very fast in the Aurelia repositories in terms of review, approval and merges so "smaller is better" definitely applies here. 

To stay in the agile mindset, adding value is important. Although this does not cover all use cases and it's probably not the perfect solution it does add direct value, it is a workable solution for the most common use case of aliases in templates in my opinion. Having this in place might inspire others to collaborate and help improve.

I would like to propose to make a high level (epic) feature request describing all use cases to be able to fully support converting aliases to relative paths. The request would be purely there to indicate and track progress of all sub tasks that have be done to call this feature complete. This PR could be the first small task and could be followed by an improvement on the implementation after properly discussing and describing what such an implementation should look like. The plugin-build task should have a description stating the current limitation and could contain a link to this high level feature request to show progress.

Lastly I would like to mention that I use Aurelia at work for a new project. I try to contribute by adding code I deem generic enough from our own project. I have limit time to be able to do this. I also spend some of my spare time adding things that are not required for my work but could benefit the Aurelia community. I'm unable to do a single big feature, some smaller ones that can also benefit our own code base I can probably sell that to a certain level.

3cp:
I have to reinstate my previous opinion, this is an engineering decision between benefits and efforts for me not supporting pursuit this feature.

Personally I also suggest to separate plugin projects based on relative features, don't merge too many features into one giant plugin, so you wouldn't have too much `"../../../"` in the first place. You can [create wrapper plugin to group many plugins together](https://aurelia.io/docs/plugins/write-new-plugin#advanced-usage), you can also wrap a list of wrapper plugins.

That's only my personal taste, you can understand why I "dismissed TS" (which you mentioned :-)) because I saw too much friends spent lots of time to get around some typing quirks which I don't want to think about. The sad truth is a big portion of popular npm packages didn't provide typing def.

Even after all the features I implemented for aurelia-cli, my contribution status speaks for my taste:
Screen Shot 2019-05-21 at 4 40 26 pm

The overall -2648 didn't tell the full story because I actually increased `spec/` folder by few k LOC.

michaelw85:
I completely understand from your point of view effort vs benefit is not worth it and other features probably have more priority (I assume), but this is the beauty of a community project I would say. If others do see the value they can/will contribute like I'm doing right now. 

The amount of contributions you did to the CLI is amazing no doubt about it, but your high involvement should not result in an opinionated product imho. That you don't see the benefit of TS is fine, although this may never result in rejecting PR's related to TS just because you don't see the benefit and are the primary contributor and reviewer. 

I was under the impression Aurelia vnext is completely (re)build using TS, so I would assume this change will follow in the CLI to maintain the same coding standards and uphold the same quality output.

I think this conversation is going a bit off topic now, I will close this PR for now and leave the code hanging for anyone who needs it. My contributions on this topic will stop as a result of course, any code related I will keep private and not invest time in making it generic enough to be able to share.

3cp:
That was why I repeatedly said it was my personal choice, not representing the whole aurelia team. You are absolutely right on everyone should have a saying on OOS projects.

That is the definition of diversity. People have different ideas work together.

That is always the way how I express myself, sometimes seems bit tough. But that’s not because I want to dictate, I am just trying to be honest to myself.

PR #1114, opened 2019-06-07
fix(postcss): avoid postcss deprecation warning by adding browserslist...

### LINKED ISSUE #1112: New plugin with autoprefixer warnings

* **Library Version:**
1.0.0

* **Operating System:**
Windows [10]

* **Node Version:**
v10.15.3

* **NPM Version:**
6.2.0

* **Browser:**
all

* **Language:**
TypeScript

* **Loader/bundler:**
Webpack

**Current behavior:**
when run npm build we get error/warning
  Replace Autoprefixer browsers option to Browserslist config.
  Use browserslist key in package.json or .browserslistrc file.
 
in process-css.ts autoprefixer is called with {browsers: ['last 2 version']} in options, which is obsolete. There should be a .broserslistrc file or a   
"browserslist": [
    "defaults"
  ] entry in package.json

npm build should run without error/warnings

to work as expected with autoprefixer

MichaelPetrinolis:
remove options when calling autoprefixer

closes #1112

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "aurelia-cli",
-  "version": "1.0.0-beta.15",
+  "version": "1.0.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -3300,8 +3300,7 @@
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "bundled": true,
-          "optional": true
+          "bundled": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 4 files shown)


3cp:
@EisenbergEffect LGTM.

@MichaelPetrinolis do you want to update your commit message to follow [our guideline](https://github.com/DurandalProject/about/blob/master/CONTRIBUTING.md#commit)? So it will be automatically written to the change log.

Something like:

fix(postcss): avoid postcss deprecation warning by adding browserslist...

EisenbergEffect:
I fixed it up :)

PR #1122, opened 2019-06-18
Run aurelia app with webpack inside a docker container

shahabganji:
Changed the webpack config so that an aurelia app running with webpack will be exposed from within a container to the host.

Changed "localhost" to "0.0.0.0" : based on what's described [here](https://github.com/webpack/webpack-dev-server/issues/547#issuecomment-284737321)

Also updated the `packge.json` file for the deprecated `opn` [package](https://www.npmjs.com/package/opn) to its new version `open` [package](https://www.npmjs.com/package/open)

### DIFF
--- a/skeleton/webpack/aurelia_project/tasks/run.ext
+++ b/skeleton/webpack/aurelia_project/tasks/run.ext
@@ -18,7 +18,7 @@ import {CLIOptions, reportWebpackReadiness} from 'aurelia-cli';
 function runWebpack(done) {
   // https://webpack.github.io/docs/webpack-dev-server.html
   let opts = {
-    host: 'localhost',
+    host: '0.0.0.0',
     publicPath: config.output.publicPath,
     filename: config.output.filename,
     hot: project.platform.hmr || CLIOptions.hasFlag('hmr'),


3cp:
Can you leave out the "opn" fix? The opn can be removed from cli itself, I fixed it in another PR. The "open" will be installed for webpack app as a dep of the app, not cli.

Basically you can leave out package.json and package-lock.json, just commit the one line change for webpack.

Also, if you could rewrite the commit message to something like: fix(webpack): allow webpack app to run inside docker container. That will help our changelog automation.

Thx!

shahabganji:
@3cp , I made some changes, does it look fine?

3cp:
LGTM, @EisenbergEffect pls merge.

3cp:
@shahabganji and @EisenbergEffect there is an issue on Windows with `http://0.0.0.0:8080`,

The browser says "This site can't be reached", you have to visit `http://localhost:8080` instead, anyone knows what's the edge case with Windows?

BTW, this is the reason I fixed some release-check task, not knowing it's actually Windows issue, not puppeteer issue.

3cp:
You might need to revert this.

PR #1123, opened 2019-06-18
few fixes

3cp:
fix(skeleton): don't overwrite user's editorconfig
fix(webpack): use npm package "open" to replace deprecated "opn"
fix(cli-bundler): fix source map path for local source file

@EisenbergEffect safe to merge, all tested.


3cp:
@EisenbergEffect pls hold the release.

I just realized the opn fix will break existing webpack apps. I will do another Pr to move “open” from app dep to cli dep.

PR #1124, opened 2019-06-19
feat(plugin, build-plugin): debug/develop plugin from consuming app with npm link

MichaelPetrinolis:

* **Library Version:**
1.0.1

* **Operating System:**
Windows [10]

* **Node Version:**
10.15.3

* **NPM Version:**
6.2.0

* **Browser:**
Chrome 75

* **Language:**
TypeScript 3.3

* **Loader/bundler:**
Webpack

**Current behavior:**
You create aurelia plugin/module for reuse in multiple enterprise apps. When you consume the plugin from the app, you cannot add breakpoints to the plugin neither you can modify the source code while debuging the app. 

I would like to develop the app/plugins from an integrated environment, set breakpoints and debug all the code from a VSCode workspace

In my current projects i use jspm link and in order to debug/build my app and plugin, i have two editors open and cannot live change the code in the plugins.
I want to use the CLI for new projects and npm link my plugins to my developing project, add the plugins from node_modules to my VSCode workspace, and set break points to .ts source files. Also by using au build-plugin --watch i am able to live reload the app when i modify the source code in the plugin. 
The hmr works for the app's files not for plugin where it reloads

### DIFF
--- a/skeleton/cli-bundler/aurelia_project/tasks/watch.ext
+++ b/skeleton/cli-bundler/aurelia_project/tasks/watch.ext
@@ -12,17 +12,62 @@ import * as gulpWatch from 'gulp-watch';
 import * as debounce from 'debounce';
 import * as project from '../aurelia.json';
 // @endif
+// @if feat.plugin
+import { buildPluginJavaScript } from './transpile';
+import { pluginMarkup } from './process-markup';
+import { pluginCSS } from './process-css';
+import { pluginJson } from './process-json';
+// @endif
+// @if !feat.plugin
 import transpile from './transpile';
 import processMarkup from './process-markup';
 import processCSS from './process-css';
 import processJson from './process-json';
 import copyFiles from './copy-files';
+// @endif
 import { build } from 'aurelia-cli';
 
 const debounceWaitTime = 100;
 let isBuilding = false;
 let pendingRefreshPaths = [];
 let watchCallback = () => { };
+// @if feat.plugin
+function processMarkupPlugin() {
+  return gulp.parallel(
+    pluginMarkup('dist/native-modules'),
+    pluginMarkup('dist/commonjs')
+  )
+}
+
+function processCSSPlugin() {
+  return gulp.parallel(
+    pluginCSS('dist/native-modules'),
+    pluginCSS('dist/commonjs')
+  )
+}
+
+function processJsonPlugin() {
+  return gulp.parallel(
+    pluginJson('dist/native-modules'),
+    pluginJson('dist/commonjs')
+  )
+}
+
+function transpilePlugin() {
+  return gulp.parallel(
+    buildPluginJavaScript('dist/native-modules', 'es2015'),
+    buildPluginJavaScript('dist/commonjs', 'commonjs')
+  );
+}
+
+let watches = [
+  { name: 'transpile', callback: transpilePlugin(), source: project.transpiler.source },
+  { name: 'markup', callback: processMarkupPlugin(), source: project.markupProcessor.source },
+  { name: 'CSS', callback: processCSSPlugin(), source: project.cssProcessor.source },
+  { name: 'JSON', callback: processJsonPlugin(), source: project.jsonProcessor.source }
+];
+// @endif
+// @if !feat.plugin
 let watches = [
   { name: 'transpile', callback: transpile, source: project.transpiler.source },
   { name: 'markup', callback: processMarkup, source: project.markupProcessor.source },
@@ -35,6 +80,7 @@ if (typeof project.build.copyFiles === 'object') {
     watches.push({ name: 'file copy', callback: copyFiles, source: src });
   }
 }
+// @endif
 
 let watch = (callback) => {
   watchCallback = callback || watchCallback;
@@ -106,6 +152,24 @@ let refresh = debounce(() => {
 
   log(`Watcher: Running ${refreshTasks.map(x => x.name).join(', ')} tasks on next build...`);
 
+// @if feat.plugin
+  let toExecute = gulp.series(
+    gulp.parallel(refreshTasks.map(x => x.callback)),
+    (done) => {
+      isBuilding = false;
+      watchCallback();
+      done();
+      if (pendingRefreshPaths.length > 0) {
+        log('Watcher: Found more pending changes after finishing build, triggering next one...');
+        refresh();
+      }
+    }
+  );
+  
+  toExecute();
+}, debounceWaitTime);
+// @endif
+// @if !feat.plugin
   let toExecute = gulp.series(
     readProjectConfiguration,
     gulp.parallel(refreshTasks.map(x => x.callback)),
@@ -124,16 +188,17 @@ let refresh = debounce(() => {
   toExecute();
 }, debounceWaitTime);
 
-function log(message/* @if feat.typescript **: string/* @endif */) {
-  console.log(message);
-}
-
 function readProjectConfiguration() {
   return build.src(project);
 }
 
 function writeBundles() {
   return build.dest();
+}
+// @endif
+
+function log(message/* @if feat.typescript **: string/* @endif */) {
+  console.log(message);
 }
 
 export default watch;
--- a/skeleton/cli-bundler/aurelia_project/tasks__if_typescript/transpile.ts
+++ b/skeleton/cli-bundler/aurelia_project/tasks__if_typescript/transpile.ts
@@ -5,13 +5,15 @@ import * as rename from 'gulp-rename';
 import * as ts from 'gulp-typescript';
 import * as project from '../aurelia.json';
 import * as fs from 'fs';
+import { CLIOptions, build, Configuration } from 'aurelia-cli';
 import * as through from 'through2';
-import {CLIOptions, build} from 'aurelia-cli';
+import * as gulpSourcemaps from 'gulp-sourcemaps';
+import * as gulpIf from 'gulp-if';
 
 function configureEnvironment() {
   let env = CLIOptions.getEnvironment();
 
-  return gulp.src(`aurelia_project/environments/${env}.ts`, {since: gulp.lastRun(configureEnvironment)})
+  return gulp.src(`aurelia_project/environments/${env}.ts`, { since: gulp.lastRun(configureEnvironment) })
     .pipe(rename('environment.ts'))
     .pipe(through.obj(function (file, _, cb) {
       // https://github.com/aurelia/cli/issues/1031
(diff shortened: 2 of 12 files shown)


skeleton/cli-bundler/aurelia_project/tasks__if_typescript/transpile.ts
```diff
@@ -5,13 +5,15 @@ import * as rename from 'gulp-rename';
 import * as ts from 'gulp-typescript';
 import * as project from '../aurelia.json';
 import * as fs from 'fs';
+import { CLIOptions, build, Configuration } from 'aurelia-cli';
 import * as through from 'through2';
-import {CLIOptions, build} from 'aurelia-cli';
+import * as gulpSourcemaps from 'gulp-sourcemaps';
```
3cp: You don't need gulp-sourcemaps.
gulp v4 has integrated sourcemaps support in gulp.src and gulp.desc, you can have a look of gulp doc.

skeleton/plugin/aurelia_project/aurelia.json
```diff
@@ -2,11 +2,7 @@
   "type": "project:plugin",
   "paths": {
     "root": "dev-app",
-    "resources": "../src",
-    "elements": "../src/elements",
-    "attributes": "../src/attributes",
-    "valueConverters": "../src/value-converters",
-    "bindingBehaviors": "../src/binding-behaviors"
```
3cp: The rest lines are actually needed for `au generate ...` command.

3cp:
Thx, I just had a glance, will test your branch when I got more time.

MichaelPetrinolis:
i tested gulp 4 support for sourcemaps but did not find a way to modify the sourceRoot for external ts files. So i left gulp-sourceMap dependency.
In order to test it, i use the following procedure
1) create an app
au new aurelia-app --unattended --select http2,typescript,htmlmin-max,sass,postcss-typical,jest,protractor,vscode,scaffold-navigation
2) create a plugin
au new aurelia-lib --unattended --select typescript,htmlmin-max,sass,postcss-typical,jest,vscode,plugin-scaffold-basic --plugin
3) cd aurelia-lib and npm link
4) cd aurelia-app and npm link aurelia-lib
5) open aurelia-app with vscode
6) add aurelia-lib folder from aurelia-app/node_modules to VSCode workspace
7) use aurelia-lib in main.ts
8) copy in the welcome.html the contents of dev-app/app.html
9) from a terminal window i run au build-plugin --watch in aurelia-lib
10)open aurelia-lib/src/binding-behaviours/primary-click in vscode and set a breakpoint
11)run project in watch mode and start debug, the breakpoint will hit in vscode when i click the button.
12) modify the plugin code, the browser will refresh, modify the app's code the browser will show the change

3cp:
I guess you don't need to play sourceRoot. Just use standard `gulp.dest(dest, {sourcemaps: true})` to write sourcemap with embedded content, so you don't need sourceRoot to find the source files.

skeleton/cli-bundler/aurelia_project/tasks/watch.ext
```diff
@@ -12,17 +12,62 @@ import * as gulpWatch from 'gulp-watch';
 import * as debounce from 'debounce';
 import * as project from '../aurelia.json';
 // @endif
+// @if feat.plugin
```
3cp: Please don't reuse this watch task.
This watch task is to watch both dev-app/ and src/ folders, and reload browser window for `au run`.

It's better to implement build-plugin using standard gulp.watch, it's quite simple.

In your build-plugin task
```js
if (CLIOptions.hasFlag('watch')) {
  main = function() {
    console.log('Watching and building plugin sources ...');
// need to adjust extnames based on app feature selection.
    return gulp.watch('src/**/*.{js,html,scss}', {ignoreInitial: false}, build);
  };
} else {
  main = gulp.series(
    build,
    () => console.log(doneMessage)
  );
}
```

skeleton/cli-bundler/aurelia_project/tasks__if_typescript/transpile.ts
```diff
@@ -50,9 +52,21 @@ export function buildPluginJavaScript(dest, format) {
...
+    };
+
+    let buildOptions = new Configuration(project.build.options, defaultBuildOptions);
+    let sourcemaps : boolean = buildOptions.isApplicable('sourcemaps');
+
     return gulp.src(project.transpiler.dtsSource)
       .pipe(gulp.src(project.plugin.source.js))
+      .pipe(gulpIf(sourcemaps, gulpSourcemaps.init()))
       .pipe(typescriptCompiler())
+      .pipe(gulpIf(sourcemaps, gulpSourcemaps.write('.', { includeContent: false, sourceRoot: '../../src' })))
```
3cp: Following code worked for me. BTW you only touched transpile.ts for TS, not transpile.js for babel. You need both for the PR to be complete.

```js
      .pipe(gulp.src(project.plugin.source.js, {sourcemaps: true})
      .pipe(transpile) // or typescriptCompiler()
      .pipe(gulp.dest(dest, {sourcemaps: true}));
```

3cp:
There are other TS typing, vscode, webpack changes I am not familiar with, (I use neither of these three things :-)), I will ask someone else to have a look.

3cp:
@Sayan751 do you use vscode debugger? If so, can you help to review this one?

@MichaelPetrinolis if we want to progress faster, I would suggest you to make another smaller PR to just turn on sourcemap, plus watch mode for `au build-plugin`. They are two features missing right now, and they are not only for `npm link` debug, they are generic features. We could merge and release the two features before playing around typing/vscode/webpack.

FYI, I thought about turning on sourcemaps few months before. But currently cli-bundler (not webpack) has a bug to support some npm package (aurelia plugins) sourcemaps in app that consumes plugins. So I didn't turn sourcemaps on for build-plugin task.

Sayan751:
@3cp I have used vscode debugger for sometime, but as the Chrome devtools offers more, I no longer use the vscode debugger. 
I also don't use npm link to include the dependency in my Aurelia project, as I had difficulties to load the modules correctly during runtime. This  problem is worse with multiple transitive dependencies.
Instead, I follow a different workflow, where a watch script copies the 'dist' of the smaller packages to the 'app' package's node_modules. Therefore, I can simplify the launch task for the vscode debugger to the following:

```json
{
  "type": "chrome",
  "request": "launch",
  "name": "Debug APP",
  "url": "https://local.host:9000/",
  "sourceMaps": true,
  "smartStep": true,
  "sourceMapPathOverrides": {
    "webpack:///./src/*": "${workspaceRoot}/APP_ROOT/src/*",
    "webpack:///./node_modules/*": "${workspaceRoot}/APP_ROOT/node_modules/*"
  }
}
```

Note the `sourceMapPathOverrides` which points to app src and node_modules. This loads the files correctly. The catch is that I need to set the break points in the files in node_modules, if I want to debug the files from the non-app projects. 

Therefore, I may not be the best person to review this PR. But I will take a look anyway, as this looks interesting 😃

skeleton/plugin/aurelia_project/tasks/build-plugin.ext
```diff
@@ -9,13 +9,20 @@ import * as del from 'del';
...
+import watch from './watch';
+
+let doneMessage = 'Finish building Aurelia plugin to dist/commonjs and dist/native-modules.';
 
 function clean() {
-  return del('dist');
+  if (CLIOptions.getEnvironment()==='prod') {
+    return del(['dist']);
+  }
+  return new Promise((resolve) => setTimeout(resolve, 100));
```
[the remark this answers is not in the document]
Sayan751: Another option is to `return Promise.resolve()`

MichaelPetrinolis:
Thanks for the review guys,

@3cp  i use sourceRoot because i want to change the source code while debugging. That's why i use npm link. The motivation is to have git submodules for the plugins in an app, debug/modify the app's/plugins code from the same editor and commit to their repo's. 
I will follow your suggestion and make a new PR for sourcemaps and watch flag based on this review.

@Sayan751  vscode allows me to modify the source files while debugging, that's why i use it - of course i have open chrome tools too :) Regarding node_modules, with npm link you must also add the node_modules/PROJECT to the VSCode workspace and set the breakpoints, otherwise they won't hit. (Unless maybe if you add a sourceMapPathOverrides to the plugins source code folder). Will test the npm install <folder> (which modifies package.json and runs preinstall/postinstall hooks) if it provides better experience than npm link

Sayan751:
@MichaelPetrinolis I can relate to you as I also have a similar setup, but with a different workflow. The watch task that copies the `dist` from plugin project to `app`'s node_modules works pretty predictably and feels more natural with the `npm i` instead of symlinking. Other difference is that instead of different git repo for different modules, I have all of those in a single repo (like monorepo, but without lerna like tools). 

However, as you also have similar setup, I am very interested about your objective. Will it be possible for you to create a GitHub repo with a small example?

MichaelPetrinolis:
@Sayan751 I will try to create an example, as soon as i get some free time. The reason for me not to have a monorepo is that I create multiple applications for different customers. All apps are based on our back-end and demo app, for which I have created some common libraries with business objects, services and widgets that I reuse and extend based on each customer's need.

MichaelPetrinolis:
@Sayan751 @3cp i created two repos that demonstrate the modular app development in an integrated environment. In order to test it, just run

- git clone --recurse-submodules https://github.com/MichaelPetrinolis/auAppA.git
- cd .\auAppA\
- git submodule foreach 'npm i'
- npm i
- code .\auAppA.code-workspace

then in VSCode ctrl+Shift+B npm: watch auModA , npm: auModB and npm start : auAppA
press F5 to start debugger
navigate to auModB project, src/index.ts and add a breakpoint.

If you press F5 in browser, the breakpoint will hit in the source file, and you will be able to modify it and see the result. The developer's flow will be to open a terminal window to the plugin's folder, run au generate something and modify the index while continuing developing and debugging the app

you can also test another example, https://github.com/MichaelPetrinolis/auAppB.git which consumes a common plugin with auAppA

The aurelia_project is created by using the #1131 PR, just added two lines in the webpack.config in the generated apps

Sayan751:
@MichaelPetrinolis I tried to clone your repo, but it seems that there are some commits that are missing your submodules repo. While cloning I am getting this error "Server does not allow request for unadvertised object 67872aba02db16f7db5f347485ce7401489da04f". The reason for this error is explained [here](https://superuser.com/questions/1240216/what-does-the-git-error-message-server-does-not-allow-request-for-unadvertised)

MichaelPetrinolis:
@Sayan751 sorry for the inconvenience, the issue was that i did not checkout the submodule's branch and made some commits. So the main git repo (auAppA) was referencing those commits that did not exist in the origin. You can try again, it should be fine. run auModA watch, auModB watch, auAppA start tasks and launch auAppA chrome debug, modify css in submodules, change code, add breakpoints to see how well behaves.

MichaelPetrinolis:
npm link or npm file install from git subproject can now be done with #1131 and adding symlinks and  source-map-loader to webpack.config

PR #1130, opened 2019-06-24
Revert hostname change in webpack

3cp:
Browser in windows has issue to deal with `http://0.0.0.0:8080` ("This site can't be reached"), have to use `http://localhost:8080` in Windows browsers.

@shahabganji this reverts #1122. This is needed because `au run --open` opens browser with the hostname in run task config.

Anyone knows some other way around this Windows edge case?

### DIFF
--- a/build/tasks/release-checks/tasks/check-javascript-errors.js
+++ b/build/tasks/release-checks/tasks/check-javascript-errors.js
@@ -7,8 +7,7 @@ module.exports = class CheckJavascriptErrors extends Task {
   constructor(url) {
     super('Check javascript errors');
 
-    // somehow puppeteer/win32 thinks http://0.0.0.0:5000 is an invalid address.
-    this.url = url.replace('//0.0.0.0', '//localhost');
+    this.url = url;
   }
 
   execute() {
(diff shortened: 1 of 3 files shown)


shahabganji:
@3cp The only workaround I know is to have `--host` and `--port` so that in the `run.ts` file we could override it with the default values, something like this:

```
au run --open --host 0.0.0.0 --port 8081
```
this way we give more flexibility to the developers and we will not also change the default behavior with `localhost`. 

I think I can handle this and make a new PR, you okay with this approach?

3cp:
I don’t quite understand what’s the change in your proposed new PR.

Is this one needed?

shahabganji:
To be able to override the default `port` and `hostName`, however, no we might not actually need it.

3cp:
I see, that’s a good addition. If I understand you correctly, we still need this PR to make default setup working on Windows.

shahabganji:
Yes, this PR is required, if you think what I've said above could be helpful, I can work on it. just let me know.

3cp:
Yes, pls make a PR. If you can add test, even better.

@EisenbergEffect let’s merge this one and release the patch version.

shahabganji:
@3cp, I don't have the problem you mentioned, on my Mac, I tested with `Chrome` and `Safari`, however, I made a new pull request to enable the user to override the `host` as well.

Screen Shot 2019-07-01 at 23 12 34

PR #1131, opened 2019-06-25
feat(transpile,build-plugin) add sourcemaps and watch support for plugin development

MichaelPetrinolis:
take into account --watch when running build-plugin.
include sourcemaps based on environment configuration

### DIFF
--- a/skeleton/cli-bundler/aurelia_project/tasks__if_babel/transpile.js
+++ b/skeleton/cli-bundler/aurelia_project/tasks__if_babel/transpile.js
@@ -9,7 +9,6 @@ import fs from 'fs';
 import through from 'through2';
 import { CLIOptions, build, Configuration } from 'aurelia-cli';
 import gulpSourcemaps from 'gulp-sourcemaps';
-import gulpIf from 'gulp-if';
 
 let env = CLIOptions.getEnvironment();
 const buildOptions = new Configuration(project.build.options);
@@ -70,22 +69,13 @@ export function buildPluginJavaScript(dest, format) {
     }
     const transpile = babel(opts);
 
-    let defaultBuildOptions = {
-      minify: 'stage & prod',
-      sourcemaps: 'dev & stage',
-      rev: false
-    };
-
-    let buildOptions = new Configuration(project.build.options, defaultBuildOptions);
-    let sourcemaps = buildOptions.isApplicable('sourcemaps');
-
     return gulp.src(project.plugin.source.js)
-      .pipe(gulpIf(sourcemaps, gulpSourcemaps.init()))
+      .pipe(gulpSourcemaps.init())
       .pipe(transpile)
-      .pipe(gulpIf(sourcemaps, gulpSourcemaps.write('.', {
+      .pipe(gulpSourcemaps.write('.', {
         includeContent: false,
         sourceRoot: '../../src'
-      })))
+      }))
       .pipe(gulp.dest(dest));
   };
 }
--- a/skeleton/cli-bundler/aurelia_project/tasks__if_babel/transpile.js
+++ b/skeleton/cli-bundler/aurelia_project/tasks__if_babel/transpile.js
@@ -74,7 +74,7 @@ export function buildPluginJavaScript(dest, format) {
       .pipe(transpile)
       .pipe(gulpSourcemaps.write('.', {
         includeContent: false,
-        sourceRoot: '../../src'
+        sourceRoot: '../../src/'
       }))
       .pipe(gulp.dest(dest));
   };
--- a/skeleton/cli-bundler/aurelia_project/tasks__if_babel/transpile.js
+++ b/skeleton/cli-bundler/aurelia_project/tasks__if_babel/transpile.js
@@ -7,18 +7,24 @@ import cache from 'gulp-cache';
 import project from '../aurelia.json';
 import fs from 'fs';
 import through from 'through2';
-import {CLIOptions, build, Configuration} from 'aurelia-cli';
+import { CLIOptions, build, Configuration } from 'aurelia-cli';
+import gulpSourcemaps from 'gulp-sourcemaps';
+import gulpIf from 'gulp-if';
 
 let env = CLIOptions.getEnvironment();
 const buildOptions = new Configuration(project.build.options);
 const useCache = buildOptions.isApplicable('cache');
 
 function configureEnvironment() {
-  return gulp.src(`aurelia_project/environments/${env}.js`, {since: gulp.lastRun(configureEnvironment)})
+  return gulp.src(`aurelia_project/environments/${env}.js`, {
+      since: gulp.lastRun(configureEnvironment)
+    })
     .pipe(rename('environment.js'))
     .pipe(through.obj(function (file, _, cb) {
       // https://github.com/aurelia/cli/issues/1031
-      fs.unlink(`${project.paths.root}/${file.relative}`, function () { cb(null, file); });
+      fs.unlink(`${project.paths.root}/${file.relative}`, function () {
+        cb(null, file);
+      });
     }))
     .pipe(gulp.dest(project.paths.root));
 }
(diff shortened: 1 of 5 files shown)


skeleton/cli-bundler/aurelia_project/tasks__if_babel/transpile.js
```diff
@@ -49,13 +62,30 @@ export function buildPluginJavaScript(dest, format) {
...
+
+    let buildOptions = new Configuration(project.build.options, defaultBuildOptions);
+    let sourcemaps = buildOptions.isApplicable('sourcemaps');
+
     return gulp.src(project.plugin.source.js)
+      .pipe(gulpIf(sourcemaps, gulpSourcemaps.init()))
       .pipe(transpile)
+      .pipe(gulpIf(sourcemaps, gulpSourcemaps.write('.', {
+        includeContent: false,
+        sourceRoot: '../../src'
```
3cp: Is `'../../src'` only for the use case of your linked npm package debug?

That sourceRoot seems irrelevant and also harmful to general use case.

skeleton/cli-bundler/aurelia_project/tasks__if_babel/transpile.js
```diff
@@ -49,13 +62,30 @@ export function buildPluginJavaScript(dest, format) {
...
     const transpile = babel(opts);
+
+    let defaultBuildOptions = {
+      minify: 'stage & prod',
+      sourcemaps: 'dev & stage',
+      rev: false
+    };
+
+    let buildOptions = new Configuration(project.build.options, defaultBuildOptions);
+    let sourcemaps = buildOptions.isApplicable('sourcemaps');
```
3cp: This seems unnecessary. Why not just turn on sourcemap without condition? That's what core aurelia modules are trying to do now, shipping sourcemap in all the builds.

It's app bundler's job to decide whether to use the source map in dev build, or strip the source map in production build. All bundlers are capable of doing that.

skeleton/plugin/aurelia_project/tasks/build-plugin.ext
```diff
@@ -10,12 +10,13 @@ import {pluginMarkup} from './process-markup';
 import {pluginCSS} from './process-css';
 import {pluginJson} from './process-json';
 import {buildPluginJavaScript} from './transpile';
+import { CLIOptions } from 'aurelia-cli';
 
 function clean() {
-  return del('dist');
+  return del(['dist/**', '!dist']);
```
3cp: It seems unnecessary. Is that removing of 'dist/' folder upsetting some of your tools/editors?

skeleton/cli-bundler/aurelia_project/tasks__if_babel/transpile.js
3cp: As I said before, `.pipe(gulp.dest('dist', {sourcemaps: '.'})` seems working fine with default included source contents. Some trouble with your debug session on that simpler setup?

MichaelPetrinolis: if you inline the sources, you are no more able to modify the original source files while debugging

skeleton/plugin/aurelia_project/tasks/build-plugin.ext
MichaelPetrinolis: yes, it creates some problems in rebuild. I will not delete dist if the --watch switch is enabled

skeleton/cli-bundler/aurelia_project/tasks__if_babel/transpile.js
3cp: Oh I see.

skeleton/plugin/aurelia_project/tasks/build-plugin.ext
3cp: You probably can skip clean step in non-first build in watch mode.

3cp:
The build-plugin task can be cleaned up to only do "clean" at beginning.

```js
import gulp from 'gulp';
import del from 'del';
import { pluginMarkup } from './process-markup';
import { pluginCSS } from './process-css';
import { pluginJson } from './process-json';
import { buildPluginJavaScript } from './transpile';
import { CLIOptions } from 'aurelia-cli';

function clean() {
  return del('dist');
}

let build = gulp.series(
  gulp.parallel(
    // package.json "module" field pointing to dist/native-modules/index.js
    pluginMarkup('dist/native-modules'),
    pluginCSS('dist/native-modules'),
    pluginJson('dist/native-modules'),
    buildPluginJavaScript('dist/native-modules', 'es2015'),

    // package.json "main" field pointing to dist/native-modules/index.js
    pluginMarkup('dist/commonjs'),
    pluginCSS('dist/commonjs'),
    pluginJson('dist/commonjs'),
    buildPluginJavaScript('dist/commonjs', 'commonjs'),
  ),
  done => {
    console.log('Finish building Aurelia plugin to dist/commonjs and dist/native-modules.');
    done();
  }
);

let main;
if (CLIOptions.hasFlag('watch')) {
  main = gulp.series(
    clean,
    () => {
      console.log('Watching plugin sources for changes ...');
      return gulp.watch('src/**/*', { ignoreInitial: false }, build);
    }
  );
} else {
  main = gulp.series(
    clean,
    build
  );
}

export { main as default };
```

skeleton/cli-bundler/aurelia_project/tasks__if_babel/transpile.js
3cp: Also forgot.

Please add a tailing `/` to `sourceRoot: '../../src/'`. Based on sourcemap spec, "This value is prepended to the individual entries in the “source” field.".

I had encountered some libs before that strictly follow the spec, it leads to `../../srcsource/file/path` without probably inserting `/`.

The tailing `/` can avoid those subtle bugs.

3cp: There is another small problem.

The package.json has
```
"files": [
    "dist"
  ],
```

It means by default, it only publish dist folder to npm package, not the src folder. We need to update package.json if we don't have embedded source in sourcemaps.

```
"files": [
    "dist",
    "src"
  ],
```

MichaelPetrinolis:
A concern of mine is that if you npm install a plugin as a file, and you already watch the application, then it might have some issues deleting the dist folder (start watching the plugin after watching the app).

3cp:
`au run` watch generates files in scripts/ folder. The dev-app running mode does not use dist/ files at all.

`au build-plugin --watch` generates files in dist/ folder.

They should be fine running together.

MichaelPetrinolis:
I mean in another Aurelia app, where you have installed the Plugin, like the setup mentioned in https://github.com/aurelia/cli/pull/1124#issuecomment-506683999

3cp:
I did some test on gulp sourcemap, it looks like your sourceRoot is much needed for correct source path no matter includeContent or not.

The default gulp v4 offering is incorrect. https://github.com/gulpjs/gulp/issues/2288#issuecomment-506953894

3cp:
LOL, I am looking back to one of my private aurelia plugin to see why my plugin is not affected by this gulp source path issue. I actually did a hack to remove duplicated paths, which I totally forgot:-)

```js
.pipe(through2.obj(function(file, enc, cb) {
    if (file.isBuffer() && file.sourceMap) {
      file.sourceMap.file = file.basename;
      file.sourceMap.sources = file.sourceMap.sources.map(n => path.basename(n));
    }
    cb(null, file);
  }))
  .pipe(gulp.dest('dist', {sourcemaps: '.'}));
```

So `element/hello-world.js` will have `hello-world.js` in both `sources` and `file` field, this avoid duplicated `element/element/` path. But my source path is still not fully correct (because it starts with dist/ not src/), but that's not an issue for me as I don't run debugger.

3cp:
This looks good to merge.
@MichaelPetrinolis has you tested your concern over the dist deleting?

MichaelPetrinolis:
@3cp It seems to be ok.

PR #1132, opened 2019-07-01
feature: support overriding host name on au run, and e2e tests

shahabganji:
This pull request helps you to override the default running host on webpack configuration from `localhost` to whatever you serve the `au run` command. This will help you to satisfy both #1122 and #1130

**e.g.**

```shell
au run --host 127.0.0.1 --port 8090 
```

```shell
au run --host 0.0.0.0 --port 8090 
```

and you can still have the `--open` command line argument as well to satisfy #1130

```shell
au run --host 127.0.0.1 --port 8090 --open
```
or simply `au run --open` to run the application on `localhost:8080`

<hr /> 

There was a minor misbehavior with `--open` on my machine, it opens two tabs, that is also fixed in this PR.


skeleton/webpack/aurelia_project/tasks/run.ext
```diff
@@ -49,7 +49,7 @@ function runWebpack(done) {
   server.listen(opts.port, opts.host, function(err) {
     if (err) throw err;
 
-    reportWebpackReadiness(opts);
+    // reportWebpackReadiness(opts);
```
3cp: Surely an accident :-)

3cp:
Please update `skeleton/webpack/aurelia_project/tasks/run.json` too for the new `"host"` option. The file is used by `au help`.

skeleton/webpack/aurelia_project/tasks/run.ext
3cp: Oh, I missed your commit message.

There are two choices:
1. removing this line like you did, but it does not help existing webpack apps.
2. remove the browser opening in `lib/build/webpack-reporter.js`, this can help all existing apps.

Both ways it can remove some deps (at least npm package "open") from the webapck reporter.

shahabganji: Should I revert this change and make changes elsewhere?

3cp: I think that should work.

shahabganji: I changed the code in this [commit](https://github.com/aurelia/cli/pull/1132/commits/1608d1b1c56ffd426da210b1636865936e7fbbcc), so that it fixes existing applications as well.

3cp:
Looks great. "open" can be removed from cli package.json too.

shahabganji:
@3cp, the CLI bundler part is tested with `Alameda`, if further testing is required let me know, otherwise remove the `WIP` flag at the beginning.


If `webpack` has been selected as the bundler, we can override both `port` and `host`

`au run --port 8081 --host 127.0.0.1`

if CLI bundler has been selected, only `port` can be overridden, I didn't find how to override the running host, if you know the place, let me know.

`au run --port 6060`

For the `dotnet core` template we could override the port via the `--port` flag, however, for the `host` it is kinda useless to override it. If you think it should be overridden as well, let me know of a mechanism that I could read the values within `.ext` files.

skeleton/cypress/aurelia_project/tasks/cypress.json
```diff
@@ -6,6 +6,23 @@
...
+    // @endif
+    {
+      "name": "port",
+      "description": "Set port number of the dev server",
+      "type": "string"
+    },
+    {
+      "name": "start",
+      "description": "Runs the application before starting to perform a testrun",
+      "type": "boolean"
```
3cp: Did you forget push up your changes on cypress.ext?

3cp:
browserSync supports [host option](https://browsersync.io/docs/options#option-host) too, you can update in the run task.

skeleton/cypress/aurelia_project/tasks/cypress.json
shahabganji: That supposed to be in another PR, however, I could merge the [branch](https://github.com/shahabganji/cli/tree/feature/e2e_cypress) into master and send all in this PR, but the PR, should be renamed then for the sake of the history of the repository. Check it out and let me know of your opinion.

shahabganji:
I have added the host option to the browser sync options just next to the port and open props, but it still reports `localhost`

shahabganji:
@3cp 
This is done, only if your answer to the following question is **no** 😄. 

**_Do you think the default behavior of `karma` should be running in the headless mode?_**

This way all unit tests have similar behavior, just like `e2e` tests which both `protractor` and `cypress` can be run with `--start` and `--run` to run the app before running tests, and run in headless mode respectively, both are ideal for automation purposes.

skeleton/cypress/aurelia_project/tasks/cypress.ext
```diff
@@ -7,14 +7,27 @@ import * as cypress from 'cypress';
...
 
-export default (cb) => {
+const runCypress = (cb) => {
   if (CLIOptions.hasFlag('run')) {
     cypress
       .run(config)
-      .then(results => (results.totalFailed === 0 ? cb() : cb('Run failed!')))
+      .then(results => {
+        (results.totalFailed === 0 ? cb() : cb('Run failed!'))
+        process.exit(0);
```
3cp: It is at least incorrect for failed test. The process will exit with non-zero code if tests failed, I think that works before.
If you want to kill the web server, you need to find way to kill it. I think at least for cli-bundler, browsersync has an api to stop the server. webpack-dev-server probably got an api too.

skeleton/protractor/aurelia_project/tasks/protractor.ext
```diff
@@ -15,13 +26,17 @@ gulp.task('protractor', (cb) => {
   // @if feat.typescript
   gulp.src('test/e2e/**/*.e2e.ts')
   // @endif
-    .pipe(protractor({configFile: 'test/protractor.conf.js'}))
+    .pipe(protractor({configFile: 'protractor.conf.js', args: process.argv.slice(3) }))
     .on('error', cb)
-    .on('end', cb);
+    .on('end', () => {
+      cb()
+      process.exit(0);
```
3cp: Same issue as the above cypress exit. The exit code needs to respect failing, especially to support CI automation.

3cp:
Awesome work! I will test all the features in a week or two. Then I will update release-check to cover some of them. I think default headless mode makes sense.

Please update `skeleton/protractor/README.md`, and readme for cypress, webpack, cli-bundler for the new features.

All those README files are merged together to the final app README.

skeleton/cypress/aurelia_project/tasks/cypress.ext
shahabganji: >  I think that works before.

When you run it in headless mode, and all tests are passed the process is still in running state, preventing the rest of the commands to be run, it is described on [discourse](https://discourse.aurelia.io/t/how-to-start-aurelia-site-before-cypress-testing/2375), but when it has failing tests, there is even no need to `process.exit(0)`

What do you think about this workaround, the listening server will be passed to the callback of the imported runApp function: 

In **run.ts** we could pass the server to the callback, done, method:
```ts

  server.listen(opts.port, opts.host, function(err) {
    if (err) throw err;

    // reportWebpackReadiness(opts);
    done(server);
  });
```
Then in the **cypress.ts** have something like this to **close** the server with **_server.close()_**: 
```ts
import { default as runApp } from './run';

const runCypress = (cb) => {
  if (CLIOptions.hasFlag('run')) {
    cypress
      .run(config)
      .then(results => {
        (results.totalFailed === 0 ? cb() : cb('Run failed!'))
      })
      .catch(cb);
  } else {
    cypress.open(config);
  }
}

export default (cb) => {
  if (CLIOptions.hasFlag('start')) {
    runApp((server) => {
      runCypress((msg) => {
        server.close();
        msg ? cb(msg) : cb();
      });
    });
    return;
  }

  runCypress(cb);
};
```

That being said, I think the following approach will be more general and it will respect the failing tests as well: 

```ts
cypress
      .run(config)
      .then(results => {
        if (results.totalFailed === 0) {
          cb();
          process.exit(0);
        } else {
          cb('Run failed!');
          process.exit(-1);
        }
      })
      .catch(cb);
```

I think the latter will have cleaner code, while the former might be more correct, ideas?

skeleton/protractor/aurelia_project/tasks/protractor.ext
shahabganji: In this case, it could also change to the following to respect the failing tests: 

```ts
gulp.task('protractor', (cb) => {
  gulp.src('test/e2e/**/*.e2e.ts')
    .pipe(
      protractor({ configFile: 'protractor.conf.js', args: process.argv.slice(3) }))
    .on('error', () =>{
      cb('Failed Tests');
      console.log("Finished with errors");
      process.exit(-1);
    })
    .on('end', () => {
      cb();
      console.log("Finished Successfully");
      process.exit(0);
    });
});
```

However, as far as I tested, I might be wrong though correct me please, the **on_end** will not get called if we have failed tests, instead, it calls the **on_error**. but I totally agree to make a clear distinction with `process.exit` to respect the passed and failed tests.

skeleton/cypress/aurelia_project/tasks/cypress.ext
shahabganji: I tested it with a `Dockerfile` for automation purposes, and when we have tests that fail, `Cypress` itself returns a non-zero code, and there is no need to interfere with `process.exit(n)`, so I guess the following will suffice: 

```ts
cypress
      .run(config)
      .then(results => {
        if (results.totalFailed === 0) {
          cb();
          process.exit(0);
        } else {
          cb('Run failed!');
        }
      })
      .catch(cb);
```

skeleton/protractor/aurelia_project/tasks/protractor.ext
shahabganji: The same scenarios goes here, I tested this one in a `Dockerfile` and `protractor` itself returns a non-zero code when it faces failed tests, so the following will suffice, IMO: 

```ts
gulp.task('protractor', (cb) => {
  gulp.src('test/e2e/**/*.e2e.ts')
    .pipe(
      protractor({ configFile: 'protractor.conf.js', args: process.argv.slice(3) }))
    .on('error', cb)
    .on('end', () => {
      cb();
      console.log("Finished Successfully");
      process.exit(0);
    });
});
```

skeleton/cypress/aurelia_project/tasks/cypress.ext
3cp: I tested the idea of `done(server)`, that killed `au run`, because gulp think the argument is an error for `done()` callback.

But `done(undefined, server)` worked, probably too hacky?

3cp:
```js
function _runWebpack(done) {
  // ...
  done(server);
  // ...
}

function runWebpack(done) {
  return _runWebpack(() => done());
}

const run = gulp.series(
  configureEnvironment,
  runWebpack
);

export { run as default, _runWebpack };
```
This is probably better. Then e2e task uses _runWebpack directly.

skeleton/protractor/aurelia_project/tasks/protractor.ext
3cp: `process.exit(-1);` it should be a positive number for exit code, at least for POSIX OS.
http://www.tldp.org/LDP/abs/html/exitcodes.html

3cp: The previous `on('error', cb)` will just work as expected, because it passes error to cb callback.

But the whole protractor is outdated using gulp v3 way. In gulp v4, we need to simplify it to:
```js
import gulp from 'gulp';
const {protractor, webdriver_update }  = require('gulp-protractor');

function runProtractor() {
  return gulp.src('test/e2e/**/*.e2e.js')
    .pipe(protractor({configFile: 'test/protractor.conf.js'}));
}

// Setting up the test task
export default gulp.series(
  webdriver_update,
  runProtractor
);
```
I tested it, pass or fail as expected.

shahabganji: Does it automatically finish the whole process when running in headless mode and all tests passed? On my machine it just waits for the user when all tests pass!

```
au protrctor --start --run --port 4040
```

**protractor.ts**
```ts
import * as gulp from 'gulp';
const { protractor, webdriver_update } = require('gulp-protractor');

import { CLIOptions } from 'aurelia-cli';
import { default as run } from './run';

function runApp(cb) {
  if (CLIOptions.hasFlag('start')) {
    run();
  }
  cb();
}

function runProtractor() {
  return gulp.src('test/e2e/**/*.e2e.ts')
    .pipe(protractor({ configFile: 'protractor.conf.js', args: process.argv.slice(3) }))
  ;
}

export default gulp.series(
  runApp,
  webdriver_update,
  runProtractor
);
```


image

skeleton/cypress/aurelia_project/tasks/cypress.ext
shahabganji: I suggested the mechanism of passing the server, but as you said it seems too hacky, and I'm afraid it becomes more complicated when it comes to other server types rather than webpack dev server; besides the sources are now coupled more and that could be harmful in a long run, in terms of maintainability.

 I would like to go with the following approach:

```ts
const runCypress = (cb) => {
  if (CLIOptions.hasFlag('run')) {
    cypress
      .run(config)
      .then(results => {
        if (results.totalFailed === 0) {
          cb();
          console.log( "Finished successfully" );
          process.exit(0);
        } else {
          console.log(`Number of failed tests: ${results.totalFailed}`);
          cb('Run failed!');
        }
      })
      .catch(cb);
  } else {
    cypress.open(config);
  }
}
```

That way, when anything rather than test fails the cypress will exit with the `catch` block and if the tests fail/pass it will go through the then clause, which either terminates the process automatically if all tests have been passed or goes through the running pipeline with failure.

These are the output on my machine in failure and passed states respectively: 

** Failed Tests**

image

** All Tests Passed **

image

skeleton/protractor/aurelia_project/tasks/protractor.ext
3cp: You new protractor task is waiting for the `run` dev server. That's why it didn't stop.

We talked about this on cypress too. We were looking into native API on web-dev-server and browsersync to close the dev server properly.

The "Process exited with error code 1" is the message from a sub-process of protractor testing (remember you cannot get args from `au protractor` inside protractor config js), **not the process of `au protractor`** where you forced exit code 0. You can check the final exit code of `au protractor` with command `echo $?` (or something equivalent on windows).

skeleton/cypress/aurelia_project/tasks/cypress.ext
3cp: When fail, probably need to double check whether `process.exit()` kills off dev server too (just try to visit http://localhost:5050 again after tests finished).

Did I miss something? I didn't see how dev server is terminated when tests passed.

The exit code 0 is not right for failure branch. See the comments for protractor.

shahabganji: > When fail, probably need to double check whether process.exit() kills off dev server too (just try to visit http://localhost:5050 again after tests finished).

I didn't use `process.exit()` in the case of failure, it is the old way by just calling `cb('Run failed!');`

>Did I miss something? I didn't see how dev server is terminated when tests passed.

 Here, I used `process.exit(0);` and also called `cb()`, since it is a success path, `process.exit(0)` has no harm, right?


>The exit code 0 is not right for failure branch. See the comments for protractor.

I did not return any exit code! could you tell me where you've seen that? Is it the `uid` in the output:

```bash
{ uid: 0,
  name: '<anonymous>',
  branch: false,
  error: 'Run failed!',
  duration: [ 17, 755040600 ],
  time: 1563172848002
}
```


**PS:** I haven't pushed any changes yet, I postponed that to the time when this discussion resolved, I could push changes if you think this way is better though.

skeleton/protractor/aurelia_project/tasks/protractor.ext
shahabganji: > We talked about this on cypress too. We were looking into native API on web-dev-server and browsersync to close the dev server properly.

I reverted all changes ( in the discussion, not pushed up to the repo yet ) on the failure case as it stopped and exited everything automatically, the failure case is exactly like the first day, both on `cypress` and `protractor`. 

For the success path I thought having `process.exit(0);` has no harm, right?, and it is way simpler than hacking around with passing the `server` between scripts, but if you think in the success path it will be better that we ask the server to terminate rather than exiting the process with exit code 0, it will change to the following:

**protractor.ts**

```ts
import * as gulp from 'gulp';
const { protractor, webdriver_update } = require('gulp-protractor');

import { CLIOptions } from 'aurelia-cli';
import { runWebpack as runWebpack } from './run';

let _server;

function runApp(cb) {
  if (CLIOptions.hasFlag('start')) {
    runWebpack((server) => {
      _server = server;
    });
  }
  cb();
}

function runProtractor(cb) {
  return gulp.src('test/e2e/**/*.e2e.ts')
    .pipe(protractor({ configFile: 'protractor.conf.js', args: process.argv.slice(3) }))
    .on('error', cb)
    .on('end', () => {
      cb();
      _server.close();
    });
  ;
}

// Setting up the test task
export default gulp.series(
  runApp,
  webdriver_update,
  runProtractor
);

```

**run.ts**
```ts
import * as webpack from 'webpack';
import * as Server from 'webpack-dev-server';
import * as project from '../aurelia.json';
import * as gulp from 'gulp';

import { config } from './build';
import configureEnvironment from './environment';
import { CLIOptions, reportWebpackReadiness } from 'aurelia-cli';

function _runWebpack(done) {
  // https://webpack.github.io/docs/webpack-dev-server.html
  let opts = {
    host: CLIOptions.getFlagValue('host') || project.platform.host,
    port: CLIOptions.getFlagValue('port') || project.platform.port,
    open: CLIOptions.hasFlag('open') || project.platform.open,
    publicPath: config.output.publicPath,
    filename: config.output.filename,
    hot: project.platform.hmr || CLIOptions.hasFlag('hmr'),
    contentBase: config.output.path,
    historyApiFallback: true,
    stats: {
      colors: require('supports-color')
    },
    ...config.devServer
  } as any;

  // Add the webpack-dev-server client to the webpack entry point
  // The path for the client to use such as `webpack-dev-server/client?http://${opts.host}:${opts.port}/` is not required
  // The path used is derived from window.location in the browser and output.publicPath in the webpack.config.
  if (project.platform.hmr || CLIOptions.hasFlag('hmr')) {
    config.plugins.push(new webpack.HotModuleReplacementPlugin());
    config.entry.app.unshift('webpack-dev-server/client', 'webpack/hot/dev-server');
  } else {
    // removed "<script src="/webpack-dev-server.js"></script>" from index.ejs in favour of this method
    config.entry.app.unshift('webpack-dev-server/client');
  }

  const compiler = webpack(config);
  let server = new Server(compiler, opts);

  server.listen(opts.port, opts.host, function (err) {
    if (err) throw err;

    reportWebpackReadiness(opts);
    done(server);
  });
}

function runWebpack(done) {
  return _runWebpack(() => done());
}

const run = gulp.series(
  configureEnvironment,
  runWebpack
);

export { run as default, _runWebpack as runWebpack };

```

3cp: This should work. Please test it :-)

BTW, I am pretty sure you can simplify
```js
function runProtractor(cb) {
  return gulp.src('test/e2e/**/*.e2e.ts')
    .pipe(protractor({ configFile: 'protractor.conf.js', args: process.argv.slice(3) }))
    .on('error', cb)
    .on('end', () => {
      cb();
      _server.close();
    });
  ;
}
```

To
```js
function runProtractor() {
  return gulp.src('test/e2e/**/*.e2e.ts')
    .pipe(protractor({ configFile: 'protractor.conf.js', args: process.argv.slice(3) }))
    .on('end', () => {
      _server.close(); // or _server.exit() for browserSync
    });
}
```

Gulp knows how to it if return value is a gulp stream.

3cp: The server close should help to finally exit the `au protractor` command without the need of `process.exit`. At least I hope so.

skeleton/cypress/aurelia_project/tasks/cypress.ext
3cp: The error is just some kind of print, not exit code.

if `au cypress --run` exit code is 0 (good), in windows powershell, you can use command `echo $?` (have to do right after `au cypress`) to print True.
If exit code is non-zero (means failed), `echo $?` will print False.

For example, CI (like travis-ci or circleci) will just check the exit code to determine whether a command is successful.

shahabganji:
@3cp, 

I chose to go with passing the server to the `e2e` scripts, so we can stop the server by calling either `close` or `exit` respectively when using `webpack` or `browserSync`; while everything works just fine when using webpack, it turns out that `browserSync.exit` has a [bug](https://github.com/BrowserSync/browser-sync/issues/1121), check [this](https://github.com/BrowserSync/browser-sync/issues/1052) with gulp too,  which stems from another [bug](https://github.com/socketio/socket.io/issues/1602#issuecomment-224270022) in socket.io.

To have one clean approach, and since `exit-code` of `0` has no harm when we are on the right track or success path, I'll go with `process.exit(0)` for now, can't check `echo $` on windows within a couple of days.

3cp:
That's an unbelievable bug on socket.io, open since 2014...
You can use `process.exit(1)` on failure, and `process.exit(0)` on pass I guess.

3cp:
BTW, pls don't rebase your work on latest master head (it doesn't work, waiting for fix on the TS conversion PR).

shahabganji:
> You can use process.exit(1) on failure,

We don't need to, the `protractor` or `cypress` both terminate the process when any tests failed.

> BTW, pls don't rebase your work on latest master head

I don't, I leave it as it is by now, and waiting for you to merge the PR when it could be. I move to #1139 for now.

shahabganji:
@3cp, I know I asked this before, but I lost the answer, should I make Karma's default behavior to be in `headless` mode?

3cp:
Yes, I agree it makes sense to run tests in headless mode by default.

shahabganji:
@3cp 
```bash
karma-coverage > istanbul@0.4.5: This module is no longer maintained, try this instead:
  npm i nyc
Visit https://istanbul.js.org/integrations for other alternatives.
```
should we change something for this warning?

3cp:
Not in this PR pls :-)

shahabganji:
@3cp,

Which one not in this PR, [this](https://github.com/aurelia/cli/pull/1132/commits/8550a99a82c6ea5714059f5e07b14b2d447b2c49) commit, or updating `istanbul` to `nyc`?

3cp:
Yes, I mean outdated istanbul. Replacing it looks like not a small change.

3cp:
We have merged and tested the webpack refactoring PR.

The run task has been simplified to a wrapper of `npm start`.

You can still pass `au run --host x.x.x.x --port 8888` to `au run`, the additional parameters are handled by webpack-dev-server directly.

You can need to rebase on top of current master and resolve the conflicts.

shahabganji:
@3cp 
How can I gracefully stop the webpack server? programmatically, like the `shutdownAppServer` we exposed in this PR.

3cp:
Probably can do from outside.

I updated release-check to use [tree-kill](https://github.com/pkrumins/node-tree-kill) to kill `au run` (which wraps webpack-dev-server command). It might be useful to you too.

Probably you can use cross-spawn (for reliable win32 spawn) to spawn `au run` or `webpack-dev-server` directly, hold the proc instance, then use tree-kill to kill `proc.pid`.

3cp:
Or you can skip `au run` and `webpack-dev-server` command, but use `webpack-dev-server` API. But I remember you pointed out a bug that prevent the dev server API for close to work.

shahabganji:
> But I remember you pointed out a bug that prevent the dev server API for close to work.

That was for `browser-sync`.

I am not sure about `tree-kill` I'll give it a try, but what was weird to me is that when I killed the process with either `process.kill(pid)` os `ps-node` 's `ps.kill(pid)`, it looks like they killed the process, but the port is still in use, either unit tests passed or failed.

3cp:
> but the port is still in use, either unit tests passed or failed

This is what tree-kill solved in release-check. It seems normal kill will skip some child process.

shahabganji:
And one more thing, how could we tell `tree-kill` to fail the processes with a proper `exit-code` when tests failed?

3cp:
I don't get your question.
tree-kill is for the running dev server process. If dev server failed to start, you should have `proc.on('error/exit',cb)` to handle it. Only when dev server started properly (by detecting stdout message like release-check task), you start e2e tests, when e2e tests finishes, you tree-kill dev server, and use the result of e2e tests to exit, not the result of tree-kill.

shahabganji:
@3cp 
I think it's done. Check it out and let me know of any ideas or feedback.

3cp:
Seems you got lots of duplicated commits/changes with those merges. Do you mind to start clean with another branch on top of latest master? New PR is if needed.

shahabganji:
@3cp  check out #1143, I close this one in respect of #1143 😄

PR #1105, opened 2019-05-23
CLI-less webpack build

Sayan751:
This PR resolves #1098.

Summary of changes:

- Parameters consolidations and other minor changes in webpack.config.js. Removed the limiting options from ts-loader as it prevents compiling `*.spec.ts` files during test with karma(+webpack).
- Replaced the au build and run scripts with npm scripts. Affected areas: `package.json`, `aurelia_project/tasks`, and release check tests. For pre-build cleanup activity, there is a small gulpfile introduced.
- Changed `tsconfig.json` to have `typeroots` applied for all cases, without that the release checks were failing for couple of suites. Other changes here enable loading a json file with ES6 import syntax.
- Lastly converted the `environment.ts`s to json files, which are loaded using the [app-settings-loader](https://github.com/Sayan751/app-settings-loader).

Issue #1098, opened 2019-05-14
[Proposal] CLI-less webpack build


Sayan751:
**I'm submitting a feature request (proposal)**

* **Library Version:**
aurelia-cli v1.0.0-beta.15

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
11.14.0

* **NPM Version:**
6.9.0

* **Browser:**
NA

* **Language:**
NA

* **Loader/bundler:**
Webpack

**Current behavior:**

The current build and run pipeline for webpack projects looks a bit overkill.

* **What is the expected behavior?**

It can be made a lot easier if the layer offered by aurelia-cli tasks are removed, and the default webpack and webpack-dev-server CLI do their job.

* **What is the motivation / use case for changing the behavior?**

The idea to have a aurelia-cli-less webpack build for a cli-genrated project was introduced to me [here](https://github.com/aurelia/cli/pull/1088#issuecomment-491507523). It makes less clutter, gives more power to the developer, and lesser things to maintain for Aurelia team.

I have prepared the following "thing" to propose the solution (and mostly as a note to me to understand `au build`, and `au run`, as I am not a heavy user of aurelia-cli).

---------------------------

#### TL;DR
No surprise that it can be done easily using npm scripts and webpack CLI parameters.

In order to have a cli-less webpack build, and run for the webpack users, it is first needed to evaluate what au-cli currently offers. At present au-cli offers 2 commands `au build` (utilizes `webpack`) and `au run` (utilizes `webpack-dev-server`). Following sections explore these 2.

#### `au build`

The `webpack.config.js` exports a function that creates the config based on the supplied arguments.

```javascript
module.exports = ({ production, server, extractCss, coverage, analyze, karma } = {}) =>({
    // webpack config for magic
})
```

`au build` via `buid.ts` script collects the values for these arguments from CLI arguments, and `aurelia.json` and uses Webpack API to build the app. Hypothetically considering, if that is the only function of `build.ts`, it can be simply replaced with the following command line (refer https://github.com/webpack/webpack/issues/2254#issuecomment-203528744).

```bash
webpack --env.production --env.server --env.extractCss --env.coverage --env.analyze --env.karma
```

##### Reducing the number of arguments for `webpack.config`

From a very naive perspective, there are 3 different setups: 

- development, 
- production, and 
- tests. 

From this perspective, the `karma`, and `coverage` parameters can be replaced with a single parameter `tests`.

[...]

PR #1137, opened 2019-07-13
feat(webpack):enable CLI-less webpack build + run

Sayan751:
Usage of aurelia-cli is removed from building and running the webpack apps. This gives developers more control over webpack, and a way to easily change the configuration, without aurelia-cli meddling in between.

The commands `au run` or `au build` will not work anymore in the webpack apps, use `npm start`, or `npm run build` instead. This commit closes the webpack part of the the issue #1098.

Summary of changes:

- Parameters consolidations and other minor changes in webpack.config.js. Removed the limiting options from ts-loader as it prevents compiling `*.spec.ts` files during test with karma(+webpack).
- Replaced the au build and run scripts with npm scripts. Affected areas: `package.json`, `aurelia_project/tasks`, and release check tests. For pre-build cleanup activity, there is a small gulpfile introduced.
- Changed `tsconfig.json` to have `typeroots` applied for all cases, without that the release checks were failing for couple of suites. Other changes here enable loading a json file with ES6 import syntax.
- Lastly converted the `environment.ts`s to json files, which are loaded using the [app-settings-loader](https://github.com/Sayan751/app-settings-loader).

Note: This PR is an updated version of PR #1105 ([justification](https://github.com/aurelia/cli/pull/1105#issuecomment-511093770))


jwx:
While removing the dependency to the CLI for building and running the webpack apps is great, I think it should still be possible to use the `au` commands. Not necessary, but possible.

skeleton/webpack/gulpFile.js
```diff
@@ -0,0 +1,10 @@
+const del = require('del');
+const gulp = require('gulp');
+const path = require('path');
+const project = require('./aurelia_project/aurelia.json');
+
+function clearDist() {
+  return del([path.resolve(__dirname, project.platform.output)]);
```
3cp: We can remove gulp and del from webpack app :-)
https://github.com/johnagan/clean-webpack-plugin

skeleton/scaffold-navigation/src/main.ext
```diff
@@ -7,7 +7,7 @@ import 'bootstrap';
 // @if feat.typescript
 import {Aurelia} from 'aurelia-framework';
 // @endif
-import environment from './environment';
+import * as environment from '../config/environment.json';
```
3cp: Both webpack and TypeScript support normal `import environment from './environment.json';`.
And explicit `.json` is only required by TypeScript.

I am not sure why you need `* as`.
Also the original line should be retained in a condition `// @if !feat.webpack` for cli-bundler app to work.

skeleton/webpack/webpack.config.js
```diff
@@ -279,20 +280,23 @@ module.exports = ({ production, server, extractCss, coverage, analyze, karma } =
       // @if feat.babel
       {
         test: /\.js$/i, loader: 'babel-loader', exclude: nodeModulesDir,
-        options: coverage ? { sourceMap: 'inline', plugins: ['istanbul'] } : {}
+        options: tests ? { sourceMap: 'inline', plugins: ['istanbul'] } : {}
       },
       // @endif
       // @if feat.typescript
-      { test: /\.ts$/, loader: "ts-loader", options: { reportFiles: [ srcDir+'/**/*.ts'] }, include: karma ? [srcDir, testDir] : srcDir },
+      { test: /\.ts$/, loader: "ts-loader" },
```
3cp: I forgot the story around karma and webpack because I am no fan of both. Please test karma setup.

skeleton/webpack/gulpFile.js
Sayan751: There might be a downside to it. I have not yet tested it but my hypothesis is that it might create a conflict with the build steps that create compiled output in the `dist` folder as a pre-webpack build step, and then during webpack build, this plugin cleans the `dist` folder. An example of such build step can be to generate the locale files in a gulp task. I think the use cases can be explained in README, and the alternatives, `del` and `clean-webpack-plugin` are presented to devs.

3cp: You are the webpack guy :), so the decision is yours.
Our flex design holds us back, if there is no `project.platform.output`, it would be just `"del scripts && ..."` in npm scripts.

alexdresko: _Please_ remove gulp and del if possible. Can't you use [npm scripts](https://docs.npmjs.com/misc/scripts)?

Sayan751: Exactly. With `npm scripts` it will be lot easier, only thing is that the output directory path comes from the config file. The path can be replicated and hardcoded in `package.json`, but there is a good chance that will create maintenance overhead. I also don't like the gulp file. I think I will go with the webpack plugin and add some info about that in README.

skeleton/scaffold-navigation/src/main.ext
Sayan751: `import * as environment ` is needed unless [`allowSyntheticDefaultImports` is set to `true`](https://www.typescriptlang.org/docs/handbook/compiler-options.html). I think I will stick to `* as` syntax instead.

Sayan751: Retained the `import environment from './environment';` for cli-bundler

skeleton/webpack/webpack.config.js
Sayan751: All release checks worked

3cp:
I played your branch, the new app still has some gulp files in `aurelia_project/tasks`.

`environment` task can be removed because it's not used.
`jest` task can be removed. Plain jest command can replace it.
`karma`/`protractor`/`cypress` task I am not very sure.

`aurelia_project/environments/*` can be removed for webpack app.

Sayan751:
@3cp I have missed the tasks files in `aurelia_project/tasks`. I will add `gulp` back to the dependency of webpack projects. I will also remove the `environment` task. Not sure what this `new-thing` task do. Can you shed some light? For the test tasks I think it is better to handle that in another PR.

3cp:
Are you talking about `au generate value-converter ...`? It's handled with in aurelia-cli (with dep on gulp), plus those template files in app's local `aurelia_project/generators/*`. That part should be fine.

3cp:
It's unfortunate this change could not get rid of all of aurelia-cli and gulp. 😄
Too many breaking changes if we do so.
Moving webpack build out of gulp will still please users.

3cp:
Since gulp is still there, I think we should have a run.ext doing something like this :-)

```js
import NPM from 'aurelia-cli/lib/package-managers/npm';

export default function() {
  // first print out a warning to say `au run` is replaced by `npm start`
  return (new NPM()).install([], process.cwd(), 'start');
  // gulp understand the returned promise
}
```

For build.ext, it is `return (new NPM()).install(['build'], process.cwd(), 'run');`.

3cp:
Oh no, just came into my mind, this PR will complicate #1132 where @shahabganji tries to add `--host`  to `au run`.
Probably need to handle that through `process.env`?

Sayan751:
@3cp @shahabganji I have not went thoroughly with the other PR. However, if devs want to change any the default config, this PR will make it easier to manipulate any configuration option in webpack. This PR generates the webpack skeletons with following npm script.

```json
"start": "webpack-dev-server --env.extractCss --port 8080"
```

Devs can now simply use `npm start -- --host 127.0.0.1 --open`, and that's it.

Sayan751:
> Since gulp is still there, I think we should have a run.ext doing something like this :-)
> 
> ```js
> import NPM from 'aurelia-cli/lib/package-managers/npm';
> 
> export default function() {
>   // first print out a warning to say `au run` is replaced by `npm start`
>   return (new NPM()).install([], process.cwd(), 'start');
>   // gulp understand the returned promise
> }
> ```
> 
> For build.ext, it is `return (new NPM()).install(['build'], process.cwd(), 'run');`.

@3cp I would suggest to have the following deprecation notice.

>`au run` task is deprecated for the webpack projects, and a potential candidate for removal in future. Consider using `npm start` instead.

Otherwise, release checks needs to be written for `au run`, and `au build` which are otherwise removed.

jwx:
I strongly disagree with removing the `au` commands for any CLI build. They might not do anything but run npm scripts, but they should still always be available. @Sayan751 @3cp

Sayan751:
@jwx That can be arranged as @3cp pointed out. In that case, the warning/deprecation notice can also be avoided. However, if those tasks are generated in the skeletons, there needs to be release check tests in place to assert the functionality. 

One possibility is that in those tests it is asserted that `au run` executes `npm start`, and that's all. Every other tests that ran for `npm start` need not to be duplicated for `au run`. However, I am not sure how to make that assertion, maybe @3cp can shed some light. Other possibility is to simply replicate the tests for `au run`.

However, I would like to point out that devs can always write those tasks. The readme can be updated with copy-paste-able snippets.

3cp:
We didn't assert in release-check, it's all predefined based on what features user selected (webpack or cli-bundler).

3cp:
If we got the compatible run/build tasks, We might be able to avoid changing release-check tasks, less code change in this PR.

shahabganji:
> Devs can now simply use `npm start -- --host 127.0.0.1 --open`, and that's it.

@Sayan751, @3cp 

What is more important than overriding `port` and `host` is the ability to run both the application and tests, in particular e2e test, in one terminal rather than separately running each one, I mean before that PR, we had to run the application first and then in another terminal run e2e tests either in interactive mode or headless;  While in the case of CI automation, we need to run the application automatically in the same process(Kinda) as the e2e tests, and tear down everything whenever necessary. Some of the work which has been done in the above-mentioned PR is about that. checkout the `--start` argument, that will run the application, runs the `e2e` tests afterward and then tears down everything. 

**e.g.** `au cypress --run --start`.

check out [this](https://github.com/shahabganji/cli/blob/master/skeleton/cypress/aurelia_project/tasks/cypress.ext#L27) line, in which we run the application before running `cypress`, the same scenario exists for `protractor`. Is there a way to replace that line of code if we use this PR? Or is there any alternative approaches?

Sayan751:
@shahabganji Does this support your use case?
https://github.com/bahmutov/start-server-and-test

Sayan751:
@3cp Does supporting the `au` tasks means supporting the same set of arguments to these tasks as well? Does that not make everything more complex than before?

shahabganji:
@Sayan751 

> @shahabganji Does this support your use case?
> https://github.com/bahmutov/start-server-and-test

I'll check it, but off the top of my head, it only supports `cypress`, correct me if I'm wrong; but `CLI` offers a bunch of combinations that need to be supported and have the same behavior, IMO.

* Protractor + Built-in CLI-Bundler
* Protractor + Webpack
* Cypress + Built-in CLI-Bundler
* Cypress + Webpack

This PR is about when we are using webpack, right? What about built-in cli bundler?

Sayan751:
@shahabganji I think I the package has nothing specific for cypress. In fact, I tested it on one of the generated skeletons, and it worked quite well. I just added the following npm scripts, and used `npm run e2e` to run the tests.

```json
"protractor": "webdriver-manager update && protractor ./test/protractor.conf.js",
"e2e": "start-server-and-test :8080 protractor"
```
>This PR is about when we are using webpack, right? What about built-in cli bundler?

You are right, this PR is about webpack only, and does not affect the cli-bundler. This means that the existing tasks will still continue to work for the cli-bundler.

3cp:
> @3cp Does supporting the `au` tasks means supporting the same set of arguments to these tasks as well? Does that not make everything more complex than before?

Don't have to. We can do very cheaply to send all the args to webpack. So the new webpack config has to support the old args as much as possible.
```js
const additional = process.argv.slice(3); // get `--host xxx` from `au run --host xxx`.

if (additional.length) {
  // calls `npm start -- --host 127.0.0.1` which pass to webpack directly
  return (new NPM()).install(['--', ...additional], process.cwd(), 'start');
}
```

Sayan751:
@3cp That's nice, I can make the change

shahabganji:
@Sayan751 

Sounds great, just one question out of curiosity, for `protractor` We have to add [these](https://github.com/shahabganji/cli/blob/master/skeleton/protractor/protractor.conf.js#L77-L80) lines to run it in `headless` mode which has no conflict with your `PR`, I guess we should still pass `--run` right?, but for the `cypress` to run in `headless` mode, I think it was [these](https://github.com/shahabganji/cli/blob/master/skeleton/cypress/aurelia_project/tasks/cypress.ext#L14-L15) lines, how is it possible to run `cypress` and `protractor` in `headless` mode?

And, one thing more, don't you think it would be great to keep the old version as well as adding this new approach? Kinda have the `deprecated`  flag for the old approach, it is just a suggestion.

cc: @3cp

3cp:
LGTM. Waiting for #1140 fix on master head.

skeleton/webpack/aurelia_project/tasks/run.ext
```diff
@@ -1,62 +1,8 @@
...
-  let server = new Server(compiler, opts);
-
-  server.listen(opts.port, opts.host, function(err) {
-    if (err) throw err;
-
-    reportWebpackReadiness(opts);
-    done();
-  });
+  const args = process.argv.slice(3);
+  return (new NPM()).install(['--', ...args], process.cwd(), 'start');
```
3cp: One thing we almost forgot. We hijacked our `npm.install()` api to do other npm scripts, this looks odd to users.

Suggest to improve our `lib/package-managers/base-package-manager.js`
```js

  run(command, args = [], workingDirectory = process.cwd()) {
    return new Promise((resolve, reject) => {
      spawn(
        this.getExecutablePath(workingDirectory),
        [command, ...args],
        { stdio: 'inherit', cwd: workingDirectory }
      )
        .on('close', resolve)
        .on('error', reject);
    });
  }

  // keep existing API so we don't have to touch many files
  install(packages = [], workingDirectory = process.cwd(), command = 'install') {
    return this.run(command, packages, workingDirectory);
  }
```

So we can do `return (new NPM()).run('start', ['--', ...args]);` here.

Another thing to suggest is to re-export NPM and Yarn from `lib/index.js`. So that we can do
```js
import { NPM } from 'aurelia-cli';
```
This will shield us from details of TS conversion, the dist file will be `dist/package-managers/npm`, not in `lib/` folder anymore.

skeleton/webpack/aurelia_project/tasks/build.ext
```diff
@@ -1,67 +1,8 @@
...
-    process.exit(1);
-  } else {
-    process.stdout.write(stats.toString({ colors: require('supports-color') }) + '\n');
-
-    if (!CLIOptions.hasFlag('watch') && stats.hasErrors()) {
-      process.exit(1);
-    }
-  }
+  const args = process.argv.slice(3);
+  return (new NPM()).install(['--', ...args], process.cwd(), 'run');
```
3cp: Did you test this? I think you missed `'build'`.
```js
return (new NPM()).install(['build', '--', ...args], process.cwd(), 'run');
```

Sayan751: Sorry, I missed it. I did test the changes. Actually, just realized that there are no release-check in place for `build` for webpack skeletons.

3cp: Oh, I see. I will add more check in release-check.

Sayan751: @3cp I have made the changes including the one you pointed above. Also added release-check for `au build` for `webpack` skeleton. I am running the release checks now. Will update this PR once it is successful

skeleton/webpack/aurelia_project/tasks/run.ext
Sayan751: done

skeleton/webpack/aurelia_project/tasks/build.ext
Sayan751: done

3cp:
I will test this one locally first before merging other PRs.

PR #1143, opened 2019-08-16
feature: overriding host and port and run tests in headless mode, e2e & karma

shahabganji:
This is a simplified, mergeable version of #1132, check that PR for discussions.

@3cp, please take a look at this, and let me know of the results. 😃


skeleton/cli-bundler/aurelia_project/tasks/run.ext
```diff
@@ -48,9 +58,17 @@ function reload() {
...
+const run = gulp.series(
   serve,
   done => { watch(reload); done(); }
 );
 
-export default run;
+const shutdownAppServer = (code) =>{
+  return new Promise(resolve => {
+    bs.exit();
+    process.exit(code || 0);
```
3cp: I understand because of the browsersync/socket.io issue you mentioned, you want to force termination. But that issue was fixed in BrowserSync/browser-sync#1121 https://github.com/BrowserSync/browser-sync/commit/4b17d60f82b5de6cb7187511d4554a4311edbb88

Are you sure we still need this patch?

skeleton/webpack/aurelia_project/tasks/run.ext
```diff
@@ -1,8 +1,41 @@
...
   console.log('`au run` is an alias of the `npm start`, you may use either of those; see README for more details.');
 
   const args = process.argv.slice(3);
-  return (new NPM()).run('start',['--', ...args]);
+
+  if (!CLIOptions.hasFlag('port')) {
+    args.push('--port');
+    args.push(port);
+  }
+  if (!CLIOptions.hasFlag('host')) {
```
3cp: Either have to remove this condition check, or remove the default value "localhost" at previous assignment.

skeleton/webpack/aurelia_project/tasks/run.ext
```diff
@@ -1,8 +1,41 @@
...
+    args.push('--host');
+    args.push(host);
+  }
+
+  return (new NPM()).run('start', ['--', ...args]);
 }
+
+const shutdownAppServer = () => {
+  return new Promise(resolve => {
+    find('port', port)
```
3cp: It's probably more reliable to enhance the implementation of `lib/package-managers/base-package-manager.js` to retain pid (or proc instance).

Something like this.

base-package-manager.js
```js
run(...) {
  ...
    this.proc = spawn(...)
}
```

run task
```js
const npm = new NPM();

const run = () {
  return npm.run(...);
}

const shutdownAppServer = () => {
    if (npm.proc) kill(npm.proc.pid...
}
```

3cp:
I am thinking another option for those killing in protractor/cypress task.

Since
1. because of webpack config change, you cannot directly create webpack dev server.
2. browsersync server has some trouble to exit by API.

why not just do `const proc = spawn('au', ['run', args])` in protractor/cypress task, then `treeKill(proc.pid)` after? We probably can skip the detection of server booting, if server failed to start, protractor/cypress will fail timeout anyway.

skeleton/webpack/aurelia_project/tasks/run.ext
shahabganji: We need both of these, because if the developer sets the `port` and `host` from the `aurelia.json` file and not the command line we will run the application on a different port than what is expected.

skeleton/cli-bundler/aurelia_project/tasks/run.ext
shahabganji: `bs.exit()`, stops the `browser-sync`, but there is still something weird the `node` does not stop, and I couldn't know why?! 

Screen Shot 2019-08-21 at 21 40 27

So, I think for now it would be better to have this patch for now, until we find a solution.

shahabganji:
I think we have discussed this approach on the other thread, at first I did the same thing, not by `spawn` but I found the `pid` and was killing the process. As far as I remember you have some points about not doing so, that's why you [suggested](https://github.com/aurelia/cli/pull/1132#discussion_r300888814) going through the `API`s of `webpack-dev-server` and `browser-sync`. I could change it to what you suggested, but I guess it is unnecessary since:

1. Everything works just now 😉 😄 
2. This way we are just creating nested child processes, that are not required somehow, besides adding other dependencies.

The start/booting of servers is not our concern for the same reason you said.

shahabganji:
@3cp, I didn't find out why the build on windows failed on Travis, could you take a look at it?

3cp:
Travis windows image has some issue. Normally a re-run can fix it. I requested re-run.

skeleton/cli-bundler/aurelia_project/tasks/run.ext
3cp: I see, I will test this.

skeleton/webpack/aurelia_project/tasks/run.ext
3cp: Oh, now I understand you try to defer the arg to webpack-dev-server to handle.
I will move `project.platform.port` and `project.platform.host` in future commits, to avoid a breaking change in previous webpack refactor.

skeleton/webpack/aurelia_project/tasks/run.ext
```diff
@@ -22,19 +23,16 @@ const run = () => {
...
 }
 
 const shutdownAppServer = () => {
   return new Promise(resolve => {
-    find('port', port)
-      .then(function (list) {
-        if (list.length) {
-          kill(list[0].pid, 'SIGKILL', function (err) {
+    if(npm.proc){
+      kill(npm.proc.pid, 'SIGKILL', function (err) {
```
3cp: BTW, does the normal gentle kill `kill(npm.proc.pid, function (err) {` work too?

3cp:
I will test this in next few days.

skeleton/webpack/aurelia_project/tasks/run.ext
```diff
@@ -1,8 +1,39 @@
...
   console.log('`au run` is an alias of the `npm start`, you may use either of those; see README for more details.');
 
   const args = process.argv.slice(3);
-  return (new NPM()).run('start',['--', ...args]);
+
+  if (!CLIOptions.hasFlag('port')) {
+    args.push('--port');
+    args.push(port);
+  }
+  if (!CLIOptions.hasFlag('host')) {
```
3cp: This two extra `--port` and `--host` destroyed normal `au run --open`.

Failed at duplicated args in
`webpack-dev-server --env.extractCss  --port 5000 "--open" "--port" "5000" "--host" "localhost"`

We should modify webpack.config.js to get port/host from aurelia.json
```
devServer: {
    ...
    port: project.platform.port || 5000,
    host: project.platform.host || 'localhost'
  },
```

Then we don't need the code here.

3cp: It looks like we also need to remove `--port 5000` from package.json.
```
"start": "webpack-dev-server --env.extractCss  --port 5000",
```

3cp:
Could you rewrite your commit messages to follow https://github.com/DurandalProject/about/blob/master/CONTRIBUTING.md#commit ?

Your `add:` `update:` `fixed:` are not supported types. They weren't be picked up by change log.

skeleton/cli-bundler/aurelia_project/tasks/run.ext
3cp: Got it 😄 

It's not browsersync holds back process, it's our default watch mode in run task!
Modify run task to
```js

const shutdownAppServer = (code) =>{
  return new Promise(resolve => {
    bs.exit();
    resolve();
  });
};

// export serve with the watch
export { run as default , serve, shutdownAppServer };
```

Then do
```js
import { serve, shutdownAppServer } from './run';

export default (cb) => {
  if (CLIOptions.hasFlag('start')) {
    runCypress(cb);
    serve();
    return;
  }

  runCypress(cb);
};
```

All work!

shahabganji: > It's not browsersync holds back process, it's our default watch mode in run task!
Modify run task to

Yeah, I said browser-sync works fine, after they closed the issue, I couldn't find a way for our wtach mode, thanks for the update, I'll change it.

skeleton/webpack/aurelia_project/tasks/run.ext
shahabganji: I have to check it, shouldn't be any problem.

skeleton/webpack/aurelia_project/tasks/run.ext
shahabganji: > This two extra `--port` and `--host` destroyed normal `au run --open`.
> 
> Failed at duplicated args in
> `webpack-dev-server --env.extractCss --port 5000 "--open" "--port" "5000" "--host" "localhost"`
> 
> We should modify webpack.config.js to get port/host from aurelia.json
> 
> ```
> devServer: {
>     ...
>     port: project.platform.port || 5000,
>     host: project.platform.host || 'localhost'
>   },
> ```
> 
> Then we don't need the code here.

`--open` works fine for me, and when I removed the code here and add it to the `devServer` section then overriding the port and host does not work properly, the webpack said it is running on the mentioned port and host but they did not, since `--open` now works and `Failed at duplicated args in` this has been handled, I guess it is good to leave the code here.

I think you were testing with `.net core` setting, since port `5000` is only used there, right? 
If so, I have tested with that configuration again, It ran without error for me for the following commands:
```
au cypress --port 8090 --host 127.0.0.1 --open --start --run 
```
```
au run --open --port 8090 --host 127.0.0.1 
```
```
dotnet run 
```

skeleton/webpack/aurelia_project/tasks/run.ext
shahabganji: It worked, so I will update it to gentle kill

shahabganji:
Sorry, my bad, I'll try my best to update the messages.

shahabganji:
I've also updated the messages, please check them out and let me know of any issues

skeleton/webpack/aurelia_project/tasks/run.ext
3cp: Yes, I was testing donet template.

shahabganji: I've tested that with `cypress`, looks to be working fine :smile:

3cp: You have not updated `webpack.config.js` to read port and host from aurelia.json.

shahabganji: Yes, I said above that it was not working, I'll give it a time tonight again. 

"the webpack said it is running on the mentioned port and host but they did not" <- this happened when I passed `--port` and `--host` as arguments, but normally it works on port `8080`

3cp:
Ok, now I understand why @Sayan751 didn't put `port` and `host` in webpack config devServer.

Once you have port/host in devServer config, `webpack-dev-server --port 8080` would not respect the argument.

The webpack config needs to pass the argument to devServer config manually now.

```js
module.exports = function(env, { mode, port, host }) {
  //...
  devServer: {
    port: port || project.platform.port,
    host: host || project.platform.host
  }
};
```

3cp:
Thanks! I will test and merge this in few days.

PR #1146, opened 2019-09-07
fix(webpack): avoid unnecessary breaking change on argument

3cp:
With this fix, `au run --analyze` `au run --hmr` will continue to work.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "aurelia-cli",
-  "version": "1.0.2",
+  "version": "1.1.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


3cp:
@Sayan751 pls review this fix.

3cp:
@EisenbergEffect I cannot use github tool to request review from @Sayan751. Is there some member setup you need to do?

EisenbergEffect:
For some reason you can't just add anyone as a reviewer. I don't know why GH has that restriction. We can always ping @Sayan751 manually and ask him to take a look.

Sayan751:
@3cp Sorry for the late reply. The changes looks good to me. I have seen this issue https://github.com/aurelia/cli/issues/1147 as well. Let's continue the discussion there.

PR #1148, opened 2019-09-10
fix(webpack): bring back au run --env prod

3cp:
@Sayan751 please help to review this fix.


3cp:
I will just merge this fix for now.

PR #1149, opened 2019-09-21
Change e2e Tests default behavior to headless

shahabganji:
@3cp, As we have talked before, I changed the `--run` flag to `--interactive` and made the default behavior of e2e test to offline/non-interactive/headless mode, just like what has been done for `karma`.

Please check out the changes and let me know of any concerns.

### DIFF
--- a/skeleton/protractor/README.md
+++ b/skeleton/protractor/README.md
@@ -6,7 +6,7 @@ First, run `au run` and keep it running.
 
 Then run `au protractor`.
 
-To perform a test-run and reports the results, do `au protractor --run`.
+To perform a test-run in interactive mode, do `au protractor --interactive`.
 
 To ask the `protractor` to start the application first and then start testing: `au protractor --start`
 
@@ -15,4 +15,4 @@ The two following flags are useful when using `--start` flag:
  * To change dev server host, do `au protractor --start --host 127.0.0.1`
 
 
-**PS:** It is also possible to mix the flags `au protractor --run --start --port 7070 --host 127.0.0.1`
+**PS:** It is also possible to mix the flags `au protractor --interactive --start --port 7070 --host 127.0.0.1`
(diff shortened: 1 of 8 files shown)


3cp:
I like the idea to move `au cypress` to just run all tests (old `au cypress --run`), then use `--interactive` to run the old `au cypress`. But I am not sure should we introduce the breaking change.

BTW, I don't understand the new `--interactive` on `au protractor`. From your implementation, it's "headless" mode, not "interactive" mode.

> To perform a test-run in interactive mode, do `au protractor --interactive`

This line confused me.

shahabganji:
That's is used in [here](https://github.com/aurelia/cli/pull/1149/commits/d807f6aa7a8d206c1733025ce25d3e6b9f399ae7#diff-b9971a5f8fdc87ccdc12c85189c6c8afR11), I tried to have the same flag for `cypress` and `protractor`; maybe I should consider each separate, **e.g.** `au cypress --interactive` and `au protractor --headless`. Ideas?

3cp:
I think we can leave cypress as it is right now to avoid breaking change.

I agree, `au protractor --headless` definitely sounds much better than `au protractor --run`.

skeleton/protractor/README.md
```diff
@@ -6,7 +6,7 @@ First, run `au run` and keep it running.
 
 Then run `au protractor`.
 
-To perform a test-run and reports the results, do `au protractor --run`.
+To perform a test-run in interactive mode, do `au protractor --interactive`.
```
3cp: I guess you forgot updating doc.
```
To perform a test-run in headless mode, do `au protractor --headless`.
```

There is one more preference of `--interactive` in this doc.

shahabganji: My bad, I thought I have changed that, nevertheless I pushed a new commit, I hope I did not miss anything else.

3cp:
LGTM.

PR #1150, opened 2019-09-24
fix(bundle): wrap commonjs module with overridden define function

dima-v:
Wraps commonjs module which has an overridden define function.

### DIFF
--- a/lib/build/amodro-trace/lib/parse.js
+++ b/lib/build/amodro-trace/lib/parse.js
@@ -298,14 +298,12 @@ define(['esprima', './lang'], function (esprima, lang) {
 
     /**
      * Determines if the file defines the require/define module API.
-     * Specifically, it looks for the `define.amd = ` expression.
      * @param {String} fileName
      * @param {String} fileContents
      * @returns {Boolean}
      */
     parse.definesRequire = function (fileName, fileContents) {
-        var foundDefine = false,
-            foundDefineAmd = false;
+        var foundDefine = false;
 
         traverse(esprima.parse(fileContents), function (node) {
             // Look for a top level declaration of a define, like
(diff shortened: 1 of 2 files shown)


dima-v:
There is an issue with commonjs module which has an overridden define function. Aurelia Cli (v1.2.0) will recognize it as an amd or requirejs module and will not add module wrapper.

This issue was found after updating quilljs to the version 1.3.7 in our project. This version adds a new dependency to our solution - "regexp.prototype.flags" package which is using commonjs format for the modules and includes a [local variable named "define"](https://github.com/es-shims/RegExp.prototype.flags/blob/e629ce3d469060bdfff837eb556bfa94bc80c092/index.js#L3).

We have tested these changes with our solution and could not see any changes to the other packages included in the bundle file.

I have created a repo with [small app](https://github.com/dima-v/quill-dep) which could be used for the testing...

3cp:
I suggest to monkey patch here (core-js got same issue), so it gets less side-effects to other npm packages.

https://github.com/aurelia/cli/blob/37f00be6cf4aa7dbea7c66a4ba08b48e1d617a7c/lib/build/bundled-source.js#L217-L223

If those es-shims guys use scoped npm package name, it will be easier for us to monkey patch. But now, I don't mind if you literally check the existance of `define = require('define-properties')` in `this.contents`. 😄

dima-v:
Thanks for the quick reply! 

Dumblerjs looks promising!  I have tested it with provided solution and it works as expected. We are going to try it with our project and hope that we will not need to add "define = require('define-properties')" check :)

3cp:
You are probably the only second person that uses dumber bundler in any production app right now. 🤣 Shoot me any question on gitter or dumber github repo if the gulp task comments were not enough to teach you how to customise it.

dima-v:
ok :) thanks!

PR #1153, opened 2019-10-31
fix(webpack): fix missing postcss in less/scss/styl processing

### LINKED ISSUE #1152: [Webpack] cssRules does not apply to .scss files imported from a .html file.

* **Library Version:**
1.2.1

* **Operating System:**
Windows 8.1

* **Node Version:**
13.0.1

* **NPM Version:**
6.12.1

* **Browser:**
all
* **Language:**
TypeScript 3.6.4

* **Loader/bundler:**
Webpack

**Current behavior:**
If I import SASS styles from the model (.ts), they are processed by the PostCss loader, autorefixer rules are applied, etc. If the same file is imported from the view (.html), the Css loader options are not applied.

This line:
https://github.com/aurelia/cli/blob/17afc51a93e5b2c50a759729e4b6a55fd138e9c9/skeleton/webpack/webpack.config.js#L277

probably should be replaced by something like this:
`use: [...cssRules, ...sassRules],`

I'd like all styles to be handled equally. The appearance of the page should not depend on the import method.

3cp:
closes #1152

@graycrow @Sayan751 please review this fix. At mean while, I am going to run through release check on webpack.

### DIFF
--- a/skeleton/webpack/webpack.config.js
+++ b/skeleton/webpack/webpack.config.js
@@ -222,8 +222,7 @@ module.exports = ({ production } = {}, {extractCss, analyze, tests, hmr, port, h
         issuer: [{ not: [{ test: /\.html$/i }] }],
         use: extractCss ? [{
           loader: MiniCssExtractPlugin.loader
-        },
-        'css-loader'
+        }, ...cssRules
         ] : ['style-loader', ...cssRules]
       },
       {
@@ -244,7 +243,7 @@ module.exports = ({ production } = {}, {extractCss, analyze, tests, hmr, port, h
       },
       {
         test: /\.less$/i,
-        use: ['css-loader', 'less-loader'],
+        use: [...cssRules, 'less-loader'],
         issuer: /\.html?$/i
       },
       // @endif
@@ -259,7 +258,7 @@ module.exports = ({ production } = {}, {extractCss, analyze, tests, hmr, port, h
       },
       {
         test: /\.styl$/i,
-        use: ['css-loader', 'stylus-loader'],
+        use: [...cssRules, 'stylus-loader'],
         issuer: /\.html?$/i
       },
       // @endif
@@ -274,7 +273,7 @@ module.exports = ({ production } = {}, {extractCss, analyze, tests, hmr, port, h
       },
       {
         test: /\.scss$/,
-        use: ['css-loader', 'sass-loader'],
+        use: [...cssRules, ...sassRules],
         issuer: /\.html?$/i
       },
       // @endif


3cp:
release check passed.

skeleton/webpack/webpack.config.js
```diff
@@ -274,7 +273,7 @@ module.exports = ({ production } = {}, {extractCss, analyze, tests, hmr, port, h
       },
       {
         test: /\.scss$/,
-        use: ['css-loader', 'sass-loader'],
+        use: [...cssRules, ...sassRules],
         issuer: /\.html?$/i
```
Sayan751: I have not used much less or stylus, but does it make sense to import these from TS/JS as well? If yes, then additional rules might be needed.

skeleton/webpack/webpack.config.js
```diff
@@ -222,8 +222,7 @@ module.exports = ({ production } = {}, {extractCss, analyze, tests, hmr, port, h
         issuer: [{ not: [{ test: /\.html$/i }] }],
         use: extractCss ? [{
           loader: MiniCssExtractPlugin.loader
-        },
-        'css-loader'
+        }, ...cssRules
```
Sayan751: This is not directly related to the changes. I have noticed that at present CSS modules have not been used. Is it better to add support for that too? If you decide to do so then, it might not make sense to import the CSS (modules) in HTML as the original local name will take no effect, as the names will be morphed (BTW which actually works vNext, with the help of convention I believe :)).

skeleton/webpack/webpack.config.js
3cp: Every file has two config, one for `issuer: /\.html?$/i` which is required css in html view. Another one is for `issuer: /\.(js|ts)?$/i` which is importing from ts/js. This one doesn't need style-loader because aurelia runtime does the work.

skeleton/webpack/webpack.config.js
3cp: The cssRules is 'css-loader' plus postcss. I think the previous line was missed when introducing cssRules into the code. The previous line missed postcss setup which is a bug.

Yes, there is no css-modules offered in Aurelia 1 skeletons.

skeleton/webpack/webpack.config.js
3cp: scss/less/stylus works same in the view of webpack I think, their loader are chained to -> css-loader -> style-loader (for importing from ts/js).

Sayan751: My bad, missed the `[tj]s` issuer.

PR #1161, opened 2020-02-18
refactor(webpack): cleanup promise-polyfill, aurelia-dialog patch for IE11

3cp:
Move promise-polyfill from ProviderPlugin to entry.
Add aurelia-dialog patch for IE11, but user needs to uncomment it.

Closes aurelia/dialog#387

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "aurelia-cli",
-  "version": "1.2.2",
+  "version": "1.2.3",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


3cp:
In local env, I saw protractor failure (similiar to the docker failure)
```
[09:17:01] I/direct - Using ChromeDriver directly...
[09:17:04] E/runner - Unable to start a WebDriver session.
[09:17:04] E/launcher - Error: SessionNotCreatedError: session not created:
  This version of ChromeDriver only supports Chrome version 80
```
But docker failure log did not have the exact log of `This version of ChromeDriver only supports Chrome version 80`. I am not sure is it same problem.

I was able to fix my local protractor issue by upgrading Chrome to latest version.
@shahabganji is there anything we can do for circleci Chrome version?

shahabganji:
@3cp I'll check it tonight, we install chrome in the Docker file and I guess we should also update that to the latest version. But it is weird that only some of them failed, and not all, while all of them use the same configuration for tools installed inside the container.

3cp:
Good, all docker tests passed. Not sure why it did not run on your PR.

PR #1162, opened 2020-02-19
fixed(protractor): fixes chrome-driver running inside container

shahabganji:
When running chrome-driver via protractor it fails with `DevToolsActivePort file doesn't exist`, two arguments should be passed to chrome based on the discussions [here](https://github.com/puppeteer/puppeteer/issues/1834), [here](https://stackoverflow.com/questions/50642308/webdriverexception-unknown-error-devtoolsactiveport-file-doesnt-exist-while-t), and [here](https://github.com/SeleniumHQ/selenium/issues/6049#issuecomment-398867205)


3cp:
This commit is now merged into master.

PR #1164, opened 2020-02-27
chore(all): replace aurelia-tools eslintrc with standard eslint:recommended

3cp:
This fixes npm warning on outdated aurelia-tools deps.


3cp:
Merge for now.
Full release check will be done after upgrading deps for user app, like jest from v24 to v25.

PR #1165, opened 2020-02-27
chore(all): upgrade user app dependencies

3cp:
There are few important major version upgrades, like systemjs, jest, babel-plugin-istanbul, gulp-typescript, karam-jasmine. Pending full release-check.


3cp:
@shahabganji 👍 your docker test covered quite a few skeletons, saved lots of time on the debug.

Maybe we should cover all the 72 skeletons that our release-check overed, if circleci didn't complain abuse 😄

3cp:
jest-cli changed api totally, without clear doc...

shahabganji:
@3cp , I see some of the tests have failed, is it due to ai changes of jest? 

I think we could have 72 tests, at least in theory, in prcatice we should see how many concurrent tasks we could run on CircleCI. I don't know excatly about that.

3cp:
Ok, the jest runCLI is moved from jest-cli to `@jest/core`.

I need to do full local release-check now.

3cp:
release-check passed on both win10 and macos.

PR #1166, opened 2020-02-28
chore(jest): upgrade aurelia-pal-nodejs to latest 2.0.0

3cp:
Pending release-check on all jest skeletons.


3cp:
Passed release-check on win10 and macOS.

PR #1167, opened 2020-03-10
feat(scaffold-navigation): upgrade to bootstrap v4 and fontawesome v5

3cp:
This has been a long requested upgrade.

### DIFF
--- a/skeleton/protractor/protractor.conf.js
+++ b/skeleton/protractor/protractor.conf.js
@@ -56,7 +56,8 @@ const config  = {
         '--disable-renderer-backgrounding',
         '--disable-device-discovery-notifications',
         '--no-sandbox',
-        '--disable-dev-shm-usage'
+        '--disable-dev-shm-usage',
+        '--window-size=1024,768'
       ]
     }
   },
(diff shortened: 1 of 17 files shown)


3cp:
@shahabganji could you help to have a look on test_docer_2? I cannot reproduce that protractor error locally.

```
1) aurelia skeleton app should navigate to users page
  Message:
    Failed: element not interactable
```

shahabganji:
@3cp Sure I will

skeleton/protractor/protractor.conf.js
```diff
@@ -56,7 +56,8 @@ const config  = {
         '--disable-renderer-backgrounding',
         '--disable-device-discovery-notifications',
         '--no-sandbox',
-        '--disable-dev-shm-usage'
+        '--disable-dev-shm-usage',
+        '--window-size=1024,768'
```
3cp: @shahabganji this fix did the job. The reason for `element not interactable` is that the protractor browser window size is too small.

Nobody knows why the default screen size is so small inside docker(or maybe vm). Explicit window size does the trick.

3cp: When element is off screen, protractor says it's `not interactable`.

PR #1168, opened 2020-03-10
fixed(docker): fixed docker tests for second configuration

shahabganji:
Changed some of the CircleCi tests, and moved the second configuration to travis.

@3cp please take a look .

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,40 +53,6 @@ jobs:
       # four scaffolding using webpack with various configurations
       - run: au new first -u -s htmlmin-max,sass,postcss-typical,karma,cypress,docker && cd ./first && npm install && npm link aurelia-cli && npm run docker:build
 
-  test_docker_2:
-    docker:
-      - image: circleci/node:8-browsers
-
-    working_directory: ~/repo
-
-    steps:
-      - checkout
-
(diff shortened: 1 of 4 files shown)


3cp:
The travis config didn't use `npm link`, how does it use local aurelia-cli to generate skeleton?

3cp:
The circleci issue is fixed in #1167, the root cause is protractor default browser window size (in docker or vm) is too small.

shahabganji:
> The travis config didn't use `npm link`, how does it use local aurelia-cli to generate skeleton?

[here](https://github.com/aurelia/cli/pull/1168/commits/94659968118be2730810dfc90dbded23d0fb4a38#diff-354f30a63fb0907d4ad57269548329e3R34) `npm link aurelia-cli` is used.




> The circleci issue is fixed in #1167, the root cause is protractor default browser window size (in docker or vm) is too small.

That is kind of weird, is that differs from docker in Travis and docker in CircleCI?

3cp:
> That is kind of weird, is that differs from docker in Travis and docker in CircleCI?

No idea :-(

3cp:
I see. So your `au new` is before link, that means the skeleton is generated using globally installed aurelia-cli from npm?

PR #1174, opened 2020-04-16
fix(scaffold-navigation): fontawesome class and docker issue

zewa666:
fixed a wrong `far` instead of `fa` class used in the navbar component

### DIFF
--- a/skeleton/docker/Dockerfile
+++ b/skeleton/docker/Dockerfile
@@ -64,11 +64,6 @@ COPY *.* ./
 
 # Copy source files
 COPY src ./src
-// @if feat['scaffold-navigation']
-
-# Copy @fortawesome files
-COPY @fortawesome ./@fortawesome
-// @endif
 
 
 // @if feat.cypress || feat.protractor || feat.jest || feat.karma
--- a/skeleton/docker/Dockerfile
+++ b/skeleton/docker/Dockerfile
@@ -64,6 +64,11 @@ COPY *.* ./
 
 # Copy source files
 COPY src ./src
+// @if feat['scaffold-navigation']
+
+# Copy @fortawesome files
+COPY @fortawesome ./@fortawesome
+// @endif
 
 
 // @if feat.cypress || feat.protractor || feat.jest || feat.karma
@@ -109,6 +114,9 @@ COPY --from=publish-stage /app/wwwroot/ .
 // @if feat["cli-bundler"] && !feat["dotnet-core"]
 COPY --from=publish-stage /app/scripts/ ./scripts/
 COPY --from=publish-stage /app/index.html/ .
+// @endif
+// @if feat['scaffold-navigation'] && feat["cli-bundler"] && !feat["dotnet-core"]
+COPY --from=publish-stage /app/@fortawesome/ ./@fortawesome/
 // @endif
 
 EXPOSE 80
--- a/skeleton/scaffold-navigation/src/nav-bar.html
+++ b/skeleton/scaffold-navigation/src/nav-bar.html
@@ -1,7 +1,7 @@
 <template bindable="router">
   <nav class="navbar navbar-expand-lg fixed-top navbar-dark bg-dark" role="navigation">
     <a class="navbar-brand" href="#">
-      <i class="far fa-home"></i>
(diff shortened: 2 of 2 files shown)


zewa666:
@shahabganji, while I was testing a new docker layout I found this typo but also another one, which might be of interest. I've picked the skeleton navigation project during project scaffold, so it includes fontawesome. I've also picked the Docker scaffold.

Now when built though, the font-awesome icons aren't shown because we're missing

```
# Copy @fontawesome stuff
COPY @fortawesome ./@fortawesome

...

COPY --from=publish-stage /app/@fortawesome/ ./@fortawesome/
```

This brings us to the situation where we'd need to modify the file based on the selection. Any idea how to do that?

3cp:
Should be `fas`, I forgot `far fa-home` is pro only.

In fa 5, `fa` is legacy, should use one of `fab`, `fas`, `far`, `fal` or `fad`.

zewa666:
@3cp, thanks for the hint, haven't used fontawesome for years. Good to know. I've updated the file

3cp:
```
// @if feat['scaffold-navigation']
...
// @endif
```

zewa666:
> ```
> // @if feat['scaffold-navigation']
> ...
> // @endif
> ```

oh and that also works in the Dockerfile? that would be awesome

3cp:
Any file in skeleton. They are all preprocessed.

3cp:
I mean you have to update cli repo, not your local file.

3cp:
Add the commit here if you want.

zewa666:
yep of course, I've just looked it up and saw. Great. I'll update the PR in a min

3cp:
@shahabganji pls help to review the docker file change.

3cp:
Circleci failed to install chrome again. Have been seeing this since latest release.

shahabganji:
I’ll do as far as I get home.

skeleton/docker/Dockerfile
```diff
@@ -64,6 +64,11 @@ COPY *.* ./
 
 # Copy source files
 COPY src ./src
+// @if feat['scaffold-navigation']
+
+# Copy @fortawesome files
+COPY @fortawesome ./@fortawesome
+// @endif
```
shahabganji, line 71: @zewa666 

I am afraid we don't need [these lines](https://github.com/aurelia/cli/pull/1174/commits/bc595e4974c3621350e9a2dc303e8b75f69ffce7#diff-e37943b4c897a421a4c47b666eaff496R67-R71) here, only the one you've added below on the production-stage will be sufficient. 
The reason is that the `@fortawesome` folder will be added when in the `publish-stage` the `au build --env prod` command is running, so it will be available from then on. 

At this point, the specified folder does not exist resulting in failure of docker commands. check the failed CircleCI tests.

zewa666:
Yes I think that's the right one. :+1:

skeleton/docker/Dockerfile
zewa666: I'm not next to the PC at the moment but will change once back. Thx for the review

zewa666:
@shahabganji sorry that it took me so long, but I've got rid of those extra lines now. Thanks for the review

shahabganji:
@zewa666 Don't be! :+1: Thanks for the time and effort. :pray:

@3cp I don't know about installing applications in Linux, is there anything wrong with [these](https://github.com/aurelia/cli/blob/master/skeleton/docker/Dockerfile#L30-L36) lines? we have two failed CI on this line, indicating that `google-chrome --version` does not exist! this line is common in all CI builds and that is weird behavior, would mind to take a look at that or re-run the failed CIs?

3cp:
I guess it's either circleci issue or protractor issue. Let me re-run them.

3cp:
I have no permission to rerun, but it consistently failed for 2/3 commits.
I am going to ignore them.

Let me know if this is ready to merge.

shahabganji:
@3cp Except for that failure, what @zewa666 does LGTM.

3cp:
It seems circleci is recovering. The latest test on master head only has one failure.

shahabganji:
I guess there were sth internal about that since that part is common among all tests.

PR #1176, opened 2020-04-30
Upgrade deps to nodejs v10

3cp:
Many deps dropped nodejs v8 support. It's time for us to move on.


3cp:
Again, circleci has trouble to install chrome in docker test.

PR #1178, opened 2020-05-02
Migrate from tslint to eslint

3cp:


3cp:
Radom failure on downloading chrome image again.

PR #1180, opened 2020-05-06
Update webpack.netcore.config.js__skip-if-exists__if_webpack

habeldabelduuhh:
remove output Path from CleanWebpackPlugin. Default Webpack output path is sufficient. Because after creating a new Aurelia Project with "au new" the application isn't able to run, because CleanWebpackPlugin doesn't accept the config output path parameter which is set on Line 24


BlischkeFlorian:
works

3cp:
Thx very much! Pls close this one and sent PR to http://github.com/aurelia/v1 instead.

We are working on decouple skeleton from this cli repo! See #1179.

PR #1181, opened 2020-05-09
feat: replace bin name "au" with "au1"

3cp:
"au xxx" now prints out a deprecation warning. It will be totally removed before releasing Aurelia 2 CLI.

BREAKING CHANGE: bin name "au" is deprecated and reserved for Aurelia 2 CLI.


3cp:
@jwx pls review.

3cp:
To be further discussed.

PR #1190, opened 2021-04-12
fix(cli-bundler): fix error on latest JS syntax, replaced outdated esprima with meriyah

3cp:


3cp:
Pending verification with skeletons.

Issue #1205, opened 2024-04-17
`au run` fails with latest node.js update for Windows


namitoyokota:
**I'm submitting a bug report**

* **Library Version:**
3.0.2

**Environment:**
* **Operating System:**
Windows 11

* **Node Version:**
21.7.3

* **NPM Version:**
10.2.5

* **Browser:**
all

* **Language:**
all

* **Loader/bundler:**
all

**Current behavior:**
When running the `au run` command to build and create bundles, execution fails with the following error message in the console:
```
Local aurelia-cli v3.0.2
Starting 'run'...
`au run` is an alias of the `npm start`, you may use either of those; see README for more details.
{
  uid: 0,
  name: 'run',
  branch: false,
  error: Error: spawn EINVAL
      at ChildProcess.spawn (node:internal/child_process:421:11)
      at spawn (node:child_process:761:9)
      at C:\{PROJECT_PATH}\node_modules\aurelia-cli\lib\package-managers\base-package-manager.js:16:19
      at new Promise (<anonymous>)
      at exports.NPM.run (C:\{PROJECT_PATH}\node_modules\aurelia-cli\lib\package-managers\base-package-manager.js:15:12)
      at run (C:\{PROJECT_PATH}\aurelia_project\tasks\run.ts:20:16)
      at bound (node:domain:432:15)
      at runBound (node:domain:443:12)
      at asyncRunner (C:\{PROJECT_PATH}\node_modules\async-done\index.js:55:18)
      at process.processTicksAndRejections (node:internal/process/task_queues:77:11) {
    errno: -4071,
    code: 'EINVAL',
    syscall: 'spawn'
  },
  duration: [ 0, 29357600 ],
  time: 1713369231980
}
Error: spawn EINVAL
    at ChildProcess.spawn (node:internal/child_process:421:11)
    at spawn (node:child_process:761:9)
    at C:\{PROJECT_PATH}\node_modules\aurelia-cli\lib\package-managers\base-package-manager.js:16:19
    at new Promise (<anonymous>)
    at exports.NPM.run (C:\{PROJECT_PATH}\node_modules\aurelia-cli\lib\package-managers\base-package-manager.js:15:12)
    at run (C:\{PROJECT_PATH}\aurelia_project\tasks\run.ts:20:16)
    at bound (node:domain:432:15)
    at runBound (node:domain:443:12)
    at asyncRunner (C:\{PROJECT_PATH}\node_modules\async-done\index.js:55:18)
[...]

PR #1206, opened 2024-04-22
Fixed broken `au run` command for Windows

namitoyokota:
In the latest node.js update, Windows machine will fail to run `au run` command. More details in the following issue: #1205. This PR adds a `shell: true` option to the parameters of the `spawn()` method (only for Windows machine) as suggested by node.js. 

Additionally, I updated the machine set up guide in the readme document. It appears that a building step no longer exists within the `package.json`.

NOTE: This code change has been tested in macOS Sonoma and Windows 11.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,3 @@ dist
 /package-lock.json
 /yarn.lock
 .npmrc
-/my-app
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ dist
 /package-lock.json
 /yarn.lock
 .npmrc
+/my-app
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [3.0.3](https://github.com/aurelia/cli/compare/v3.0.2...v3.0.3) (2024-04-17)
+
+
+
 ## [3.0.2](https://github.com/aurelia/cli/compare/v3.0.1...v3.0.2) (2023-10-07)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,3 @@
-## [3.0.3](https://github.com/aurelia/cli/compare/v3.0.2...v3.0.3) (2024-04-17)
-
-
-
 ## [3.0.2](https://github.com/aurelia/cli/compare/v3.0.1...v3.0.2) (2023-10-07)
 
 
--- a/README.md
+++ b/README.md
@@ -27,9 +27,9 @@ The `au new` command now simplify wraps `npx makes aurelia/v1`. Users can direct
 1. Clone the aurelia-cli: `git clone https://github.com/aurelia/cli.git`
 2. Go into the cli directory: `cd cli`
 3. Run `npm install`
-6. Link the cli with: `npm link`
-7. Create a new project with `au new` or use an existing project. The linked CLI will be used to create the project.
-8. In the project directory, run `npm link aurelia-cli`. The linked CLI will then be used for `au` commands such as `au run`
+4. Link the cli with: `npm link`
+5. Create a new project with `au new` or use an existing project. The linked CLI will be used to create the project.
+6. In the project directory, run `npm link aurelia-cli`. The linked CLI will then be used for `au` commands such as `au run`
 
 ## Running the Tests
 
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -27,10 +27,9 @@ The `au new` command now simplify wraps `npx makes aurelia/v1`. Users can direct
 1. Clone the aurelia-cli: `git clone https://github.com/aurelia/cli.git`
 2. Go into the cli directory: `cd cli`
 3. Run `npm install`
-4. Run build `npm run build`
-5. Link the cli with: `npm link`
-6. Create a new project with `au new` or use an existing project. The linked CLI will be used to create the project.
-7. In the project directory, run `npm link aurelia-cli`. The linked CLI will then be used for `au` commands such as `au run`
+6. Link the cli with: `npm link`
```
3cp, line 30: Note the numbers need to be adjusted for removed ones.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## [3.0.3](https://github.com/aurelia/cli/compare/v3.0.2...v3.0.3) (2024-04-17)
```
3cp, line 1: Please remove the change. When we release new version, the script will automatically update change log file.

lib/package-managers/base-package-manager.js
```diff
@@ -11,11 +11,19 @@ exports.BasePackageManager = class {
   }
 
   run(command, args = [], workingDirectory = process.cwd()) {
+    const isWindows = process.platform === "win32";
```
3cp, line 14: Please move this line to the beginning of the file, so it only needs to calculate once.

lib/package-managers/base-package-manager.js
```diff
@@ -11,11 +11,19 @@ exports.BasePackageManager = class {
...
+    let executable = this.getExecutablePath(workingDirectory);
+    let options = { stdio: "inherit", cwd: workingDirectory };
+    if (isWindows) {
+      executable = `"${executable}"`
```
3cp, line 18: The safer way to quote a string is to use `JSON.stringify(executable)`.

lib/package-managers/base-package-manager.js
```diff
@@ -11,11 +11,19 @@ exports.BasePackageManager = class {
...
+      options = { ...options, shell: true }
```
3cp, line 19: `option.shell = true;` is much cheaper. Or even move it outside of the if.

`const options = { stdio: ..., cwd: ..., shell: isWindows };`.

lib/package-managers/base-package-manager.js
3cp: Also please add a brief comment to explain what it fixes.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "aurelia-cli",
-  "version": "3.0.2",
+  "version": "3.0.3",
```
3cp, line 3: Please remove this change too.

.gitignore
```diff
@@ -12,3 +12,4 @@ dist
 /package-lock.json
 /yarn.lock
 .npmrc
+/my-app
```
3cp, line 15: Please remove this change which is only related to your local setup.

3cp (changes requested):

3cp:
The doc "Create a new project with au new" doesn't mean to create new project within your cli repo folder. I think that where it confused you.

README.md
namitoyokota: Done

CHANGELOG.md
namitoyokota: Done

lib/package-managers/base-package-manager.js
namitoyokota: Good idea. Since there is no longer a need for a variable this way, I did as suggested within the function call parameters.

lib/package-managers/base-package-manager.js
namitoyokota: Done

.gitignore
namitoyokota: Done

package.json
namitoyokota: Done

3cp (approved):

3cp:
I will fix the CI setup after.

3cp:
Seems like nodejs 20 is not happy with aurelia-cli. I will fix those regressions before next release.

3cp:
It was mock-fs (only for unit tests) doesn't work with nodejs 20. I need to remove it.

PR #1209, opened 2024-09-04
chore(Deps): bumped Gulp to 5.0.0 and Nodemon to 3.1.4

raiseandfall:
Hi there, 

I've noticed 17 vulnerabilities in sub-deps. The two concerned dependencies are Nodemon and Gulp.
This PR bumps them to the following:
- Gulp to 5.0.0
- Nodemon to 3.1.4

I've tested post update and saw no issues.

This clears any vulnerabilities.
Could we bump a fix version for this to allow consumers apps to take advantage of it?
Thanks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "enquirer": "^2.3.6",
     "events": "^3.3.0",
     "fs-browser-stub": "^1.0.1",
-    "gulp": "^4.0.2",
+    "gulp": "^5.0.0",
     "htmlparser2": "^8.0.1",
     "https-browserify": "^1.0.0",
     "lodash": "^4.17.21",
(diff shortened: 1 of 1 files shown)


3cp:
When we tried gulp v5 with au2, we found out v5 has lots of issue with existing plugins. We cannot upgrade it yet.

3cp:
Did you try it with gulp-typescript? It didn't work last time we tried.

raiseandfall:
I didn't have a chance to test with gulp-typescript. Makes sense to wait for gulp plugins to be stable with v5. Hopefully we won't have to wait much longer.

raiseandfall:
> Did you try it with gulp-typescript? It didn't work last time we tried.

@3cp 
Coming back here (almost a year later 😄) I was able to test with gulp-typescript and it's running smoothly.
I'm aware stability with gulp 5 and plugins is still not great, so not expecting any movement forward if it's still too early.

PR #1215, opened 2025-08-04
fix: source maps for ES modules have wrong offsets after conversion to AMD (#1214)

nenadvicentic:
Fixes issue with ES module source maps, after conversion to AMD. Described here #1214.

PR #1216, opened 2025-08-06
fix: `bundled-source.js` - transformed source map is saved to cache.

nenadvicentic:
Previous pull request #1215 did not take in consideration `BundledSource` cache, so we would get correct source map only on first `au build`. Second `au build`, without source file change, when cache is used, CLI would fallback to original `file.sourceMap`, because transformed source map was not cached.

Issue is now fixed:

1. Added `_transformedSourceMap` property, to differentiate from `file.sourceMap`.
2. `_transformedSourceMap` saved/read from cache.
3. Two new unit tests added at the bottom `bundled-source.spec.js`. They are variation on 2 existing cache tests.
4. Tested in my own projects, with `build.options.cache` on and off.

### DIFF
--- a/lib/build/bundled-source.js
+++ b/lib/build/bundled-source.js
@@ -14,11 +14,14 @@ exports.BundledSource = class {
     this.includedIn = null;
     this.includedBy = null;
     this._contents = null;
+    this._transformedSourceMap = null;
     this.requiresTransform = true;
   }
 
   get sourceMap() {
-    return this.file.sourceMap;
+    return this._transformedSourceMap === null
+      ? this.file.sourceMap
+      : this._transformedSourceMap;
   }
 
   get path() {
@@ -215,6 +218,7 @@ exports.BundledSource = class {
 
       if (cache) {
         this.contents = cache.contents;
+        this._transformedSourceMap = cache.transformedSourceMap || null;
         deps = cache.deps;
       } else {
         let contents;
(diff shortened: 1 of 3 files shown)


lib/build/bundled-source.js
```diff
@@ -14,11 +14,14 @@ exports.BundledSource = class {
...
     this._contents = null;
+    this._transformedSourceMap = null;
     this.requiresTransform = true;
   }
 
   get sourceMap() {
-    return this.file.sourceMap;
+    return this._transformedSourceMap === null
+      ? this.file.sourceMap
+      : this._transformedSourceMap;
```
3cp, line 24: I think you can just do `return this._transformedSourceMap || this.file.sourceMap;`.

3cp (changes requested):
Looks good. Just one small suggestion.

lib/build/bundled-source.js
nenadvicentic: I adopted the suggested change and I also added check when caching - if property `_transformedSourceMap` is `null`, cache value is set to `undefined`, so it's skipped in `JSON.stringify`. No need to cache `null` values.

3cp (approved):

bigopon:
3.0.7 has been released, thanks @nenadvicentic @3cp 🙏

nenadvicentic:
@bigopon @3cp I've upgraded to 3.0.7 in one of the projects. Now everything works as expected. Thank you both!

PR #1218, opened 2026-05-20
Move yargs from devDependencies to dependencies

strepon:
The `yargs` package should be in dependencies, as it is used in the `lib/build/webpack-reporter.js` file.

I encounter this in my project when I upgraded `yargs` packages (which `aurelia-cli` relied on) brought by other packages to version 18, then the build fails as `yargs.argv` used in the file is not valid anymore in v18. Thus, `aurelia-cli` should depend on `yargs` v17 instead.


3cp (approved):
Thx!

3cp:
@bigopon please release a patch version.

bigopon:
3.0.8 is out, took a bit time there was some hiccups with npm.

